PR #84, opened 2023-11-28
Upgrade mountpoint version to 1.3.0

jjkr:
Pulling in new features from mountpoint 1.2 and 1.3 releases.


dlakhaws (changes requested):
lgtm! can we add an integration test with S3 Express?

PR #90, opened 2023-12-01
Upgrade MP and adding tests for S3 Express

dlakhaws:
*Description of changes:*
* upgrade to MP v1.3.1
* add automated testing for [S3 Express One Zone](https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-express-one-zone.html)


tests/e2e-kubernetes/testdriver.go
```diff
@@ -80,8 +87,44 @@ func (d *s3Driver) CreateVolume(ctx context.Context, config *framework.PerTestCo
...
+		// CreateBucketConfiguration: &types.CreateBucketConfiguration{
+		// 	LocationConstraint: types.BucketLocationConstraint(BucketRegion),
+		// },
+	}
+
+	if config.Prefix == custom_testsuites.S3ExpressTestIdentifier {
+		// assume us-east-1 since that's where our integration tests currently do their work
+		// https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-express-networking.html
+		regionAz := "use1-az4"
+		if BucketRegion == "us-west-2" {
```
jjkr, line 100: I'm guessing this won't work for regions that aren't `us-east-1` or `us-west-2`? We should at least track a task that will fix this for other regions.

.github/workflows/e2e-tests.yaml
```diff
@@ -61,7 +61,7 @@ jobs:
     needs: build
     strategy:
       matrix:
-        cluster-type: ["kops", "eksctl"]
+        cluster-type: ["eksctl", "kops"]
```
jjkr, line 64: Why the swap here?

jjkr (approved):

tests/e2e-kubernetes/testdriver.go
dlakhaws: yea I actually took a brief look into this but there's no programmatic way to see where S3 Express is supported, and right now the only NA regions it supports is us-east-1 and us-west-2, but agreed something to look into in the future

.github/workflows/e2e-tests.yaml
dlakhaws: unintentional, I had removed it in a previous commit to test kops only and then added it back in but in reverse order

jjkr (approved):

PR #102, opened 2023-12-08
Systemd refactor

jjkr:
Refactoring of the systemd code to pull out a new system package. The go systemd library was replaced by using the dbus library directly to address some liveness risks.


pkg/system/pts.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// Real os implementation of the Pts interface
+type OsPts struct{}
+
+func NewOsPts() Pts {
+	return &OsPts{}
+}
+
+// Create a new pseduo terminal (pts). Returns a ReaderCloser for the master device and a pts number
```
mikeran0, line 30: nit: pseudo

dlakhaws (dismissed):

pkg/system/pts.go
jjkr: 🤦 thanks. Fixed

Issue #106, opened 2023-12-12
Make hostTokenPath configurable (/var/lib/kubelet/plugins/s3.csi.aws.com/token)


mikkeloscar:
/kind bug

**What happened?**

The csi-driver is making assumptions about the kubelet directory (default `/var/lib/kubelet`) so it doesn't work in environments not running with the default. Specifically this constant: https://github.com/awslabs/mountpoint-s3-csi-driver/blob/d32aaa314e537900f2c5e66bac420e11a8b97e07/pkg/driver/mount.go#L46 means that in our environment where the kubelet directory is in a different location the `AWS_WEB_IDENTITY_TOKEN_FILE` doesn't work.

**What you expected to happen?**

It should be possible to define a custom path for the kubelet directory e.g. via a flag, such that `AWS_WEB_IDENTITY_TOKEN_FILE` can work in environments with custom settings. It's fine if the current constant is the default, it should just be possible to change.

**How to reproduce it (as minimally and precisely as possible)?**

Run kubelet with a different setting for `--root-dir` and configure the monthpath accordingly in the `s3-csi-driver` daemonset.

We run with a non-standard root because we do some tricks on the AMI to put the root on the root disk for instances without instance-store and for instance-storage instance-types, we put the root directory on the SSD (I don't think the actual use case is relevant, just to give context why we have it different from the default).

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`):
- Driver version:

dlakhaws:
Thank you for the report. We are indeed hardcoding that path and need to make a fix for setups with non a standard `root-dir`. This will be planned for our next release.

mikkeloscar:
Great! looking forward to that.

jFYI: even with this fixed I found another issue with mountpoint-s3: https://github.com/awslabs/mountpoint-s3/issues/675

mikkeloscar:
> jFYI: even with this fixed I found another issue with mountpoint-s3: https://github.com/awslabs/mountpoint-s3/issues/675

This issue is fixed by setting `AWS_DEFAULT_REGION` on the daemonset suhc that mountpoint-s3 uses the right credentials chain, see description of the issue: https://github.com/awslabs/mountpoint-s3/issues/675#issuecomment-1869637085

dlakhaws:
[...]

Issue #104, opened 2023-12-12
Make container resources configurable


marcheyer:
/feature

**Is your feature request related to a problem? Please describe.**
We require every container in every pod running in our cluster having set resource requests and limits.
This is not possible yet

**Describe the solution you'd like in detail**
I will submit a PR

marcheyer:
Additionally i want to make the resources configurable throught CloudFormation.
Can you give me a hint how to achieve this? This validation scheme should work:
`{
  "$ref": "#/definitions/MountPointCSIDriver",
  "$schema": "http://json-schema.org/draft-06/schema#",
  "definitions": {
    "MountPointCSIDriver": {
      "additionalProperties": false,
      "properties": {
        "node": {
          "resources": {
            "$ref": "#/definitions/Resources"
          }
        },
        "sidecars": {
          "nodeDriverRegistrar": {
            "resources": {
              "$ref": "#/definitions/Resources"
            }
          },
          "livenessProbe": {
            "resources": {
              "$ref": "#/definitions/Resources"
            }
          }
        },
        "initContainer": {
          "resources": {
            "$ref": "#/definitions/Resources"
          }
        }
      },
      "title": "MountPointCSIDriver",
      "type": "object"
    },
    "Limits": {
      "additionalProperties": false,
      "properties": {
        "cpu": {
          "type": "string"
        },
        "memory": {
          "type": "string"
        }
      },
      "title": "Limits",
      "type": "object"
    },
    "Resources": {
      "additionalProperties": false,
      "properties": {
        "limits": {
          "$ref": "#/definitions/Limits"
        },
        "requests": {
          "$ref": "#/definitions/Limits"
        }
      },
      "title": "Resources",
      "type": "object"
    }
  }
}`

PR #116, opened 2023-12-28
Making helm charts configurable

dlakhaws:
*Issue #, if available:*
* [104](https://github.com/awslabs/mountpoint-s3-csi-driver/issues/104)
* [106](https://github.com/awslabs/mountpoint-s3-csi-driver/issues/106)

*Description of changes:*
* Make helm chart configurable

### DIFF
--- a/pkg/driver/mount.go
+++ b/pkg/driver/mount.go
@@ -39,11 +39,11 @@ const (
 	defaultRegionEnv   = "AWS_DEFAULT_REGION"
 	stsEndpointsEnv    = "AWS_STS_REGIONAL_ENDPOINTS"
 	MountS3PathEnv     = "MOUNT_S3_PATH"
+	hostTokenPath      = "HOST_TOKEN_PATH"
 	defaultMountS3Path = "/usr/bin/mount-s3"
 	procMounts         = "/host/proc/mounts"
 	userAgentPrefix    = "--user-agent-prefix"
 	csiDriverPrefix    = "s3-csi-driver/"
-	hostTokenPath      = "/var/lib/kubelet/plugins/s3.csi.aws.com/token"
 )
 
 // Mounter is an interface for mount operations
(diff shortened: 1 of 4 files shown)


pkg/driver/mount.go
```diff
@@ -39,11 +39,11 @@ const (
 	defaultRegionEnv   = "AWS_DEFAULT_REGION"
 	stsEndpointsEnv    = "AWS_STS_REGIONAL_ENDPOINTS"
 	MountS3PathEnv     = "MOUNT_S3_PATH"
+	hostTokenPath      = "HOST_TOKEN_PATH"
```
jjkr, line 42: This doesn't look right. I would expect another code change that reads the environment variable. Unless mountpoint is reading the env var for us. Either way make sure this works on a clean cluster with service account auth.

pkg/driver/mount.go
```diff
@@ -195,6 +195,11 @@ func passthroughEnv() []string {
 	}
 	webIdentityFile := os.Getenv(webIdentityTokenEnv)
 	awsRoleArn := os.Getenv(roleArnEnv)
+	hostTokenPath := os.Getenv(hostTokenPathEnv)
+	if hostTokenPath == "" {
+		// set the default in case the env variable isn't found
+		hostTokenPath = "/var/lib/kubelet/plugins/s3.csi.aws.com/token"
+	}
 	if webIdentityFile != "" {
```
iostreamdoth, line 203: should this be  `==`  in place of `!=` ? if webIdentityFile is set should we not use it?

pkg/driver/mount.go
dlakhaws: latest commit wasn't pushed, should be there now!

pkg/driver/mount.go
dlakhaws: hmm not quite, we're trying to set the env variable here based on the driver env (if you look at `webIdentityFileEnv` it comes from driver.go, this isn't something set by the customer) so if the driver is running with the web token, that should be passed here as well (but with different paths)

jjkr (approved):

pkg/driver/mount.go
iostreamdoth: I see, so this line.
https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/pkg/driver/driver.go#L78C4-L82
I was little confused as it was pointing to '/csi/token' as path in driver. I assumed this might be some mechanism to override it. Thanks for clarification.

iostreamdoth (approved):

Issue #122, opened 2024-01-16
ImagePullSecret is missing


rainerleber:
/kind bug
The imagePullSecret ref is missing in the DaemonSet

**What happened?**
The Chart not pulling containers from private ECR

**What you expected to happen?**
Provide the PullSecret

**How to reproduce it (as minimally and precisely as possible)?**
Try with private ecr connection and authentication

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`):
- Driver version:

jjkr:
Thank you for the request. We will prioritize this. It likely won't make the release going out this week, but will be targeted for the next one.

PR #123, opened 2024-01-17
Add imagePullSecrets support

truestory1:
*Issue #122 

*Description of changes:*
add imagePullSecrets support for private registries (optional)

### DIFF
--- a/charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
@@ -37,6 +37,9 @@ spec:
       initContainers:
         - name: install-mountpoint
           image: {{ printf "%s%s:%s" (default "" .Values.image.containerRegistry) .Values.image.repository (default (printf "v%s" .Chart.AppVersion) (toString .Values.image.tag)) }}
+          {{- if .Values.imagePullSecrets }}
+          imagePullSecrets: {{- toYaml .Values.imagePullSecrets | nindent 8 }}
+          {{- end }}
           imagePullPolicy: IfNotPresent
           command:
             - "/bin/install-mp"
@@ -49,6 +52,9 @@ spec:
       containers:
         - name: s3-plugin
           image: {{ printf "%s%s:%s" (default "" .Values.image.containerRegistry) .Values.image.repository (default (printf "v%s" .Chart.AppVersion) (toString .Values.image.tag)) }}
+          {{- if .Values.imagePullSecrets }}
+          imagePullSecrets: {{- toYaml .Values.imagePullSecrets | nindent 8 }}
+          {{- end }}
           imagePullPolicy: IfNotPresent
           args:
             - --endpoint=$(CSI_ENDPOINT)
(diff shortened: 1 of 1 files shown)


charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
```diff
@@ -37,6 +37,9 @@ spec:
       initContainers:
         - name: install-mountpoint
           image: {{ printf "%s%s:%s" (default "" .Values.image.containerRegistry) .Values.image.repository (default (printf "v%s" .Chart.AppVersion) (toString .Values.image.tag)) }}
+          {{- if .Values.imagePullSecrets }}
+          imagePullSecrets: {{- toYaml .Values.imagePullSecrets | nindent 8 }}
```
dlakhaws, line 41: since we know it's a list, and should contain a `name` field, thoughts on doing it like this?
```
      {{- if .Values.imagePullSecrets }}
      imagePullSecrets:
      {{- range .Values.imagePullSecrets }}
        - name: {{ . }}
      {{- end }}
      {{- end }}
```

truestory1:
With current changes the template is rendered as
```
$ helm template -s templates/node.yaml . | yq '.spec.template.spec' | head -15
nodeSelector:
  kubernetes.io/os: linux
serviceAccountName: s3-csi-driver-sa
priorityClassName: system-node-critical
tolerations:
  - key: CriticalAddonsOnly
    operator: Exists
  - operator: Exists
    effect: NoExecute
    tolerationSeconds: 300
imagePullSecrets:
  - name: test-secret-name
initContainers:
  - name: install-mountpoint
    image: public.ecr.aws/mountpoint-s3-csi-driver/aws-mountpoint-s3-csi-driver:v1.1.0
...
...

```

where
```
$ cat values.yaml | grep -A1 imagePullSecrets  
imagePullSecrets:
  - name: test-secret-name
```

I did not add test scenario for this specific case.

charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
dlakhaws: @truestory1 take a look at this comment, still holds with the recent changes! this way users don't have to append `name:` at the beginning of each value

truestory1: With recent change we have:

```
$ helm template -s templates/node.yaml . | yq '.spec.template.spec' | head -15                                                        

nodeSelector:
  kubernetes.io/os: linux
serviceAccountName: s3-csi-driver-sa
priorityClassName: system-node-critical
tolerations:
  - key: CriticalAddonsOnly
    operator: Exists
  - operator: Exists
    effect: NoExecute
    tolerationSeconds: 300
imagePullSecrets:
  - name: test-secret-name
initContainers:
  - name: install-mountpoint
    image: public.ecr.aws/mountpoint-s3-csi-driver/aws-mountpoint-s3-csi-driver:v1.1.0

```
where
```
$ cat values.yaml | grep -A1 imagePullSecrets                                                                                         
imagePullSecrets:
  - test-secret-name
```

PS: I was confused by using or non-using the name field. Thanks again for pointing me to right destination

PR #131, opened 2024-01-22
Allow actions to be run for forked PRs

truestory1:
*Description of changes:*
according to https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#adding-permissions-settings
this change should fix https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/7616759109/job/20745254205?pr=123

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -39,7 +39,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       id-token: write
-      contents: read
+      contents: write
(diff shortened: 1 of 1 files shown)


jjkr (approved):

truestory1:
according to https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token "Maximum access for pull requests from public forked repositories" is read. I would close this PR. Due to security model seems it's not possible to implement this change.

PR #149, opened 2024-02-13
Refactor mounting

jjkr:
Refactoring the node and mount code for better separation of concerns and testability. Some of the mount logic was moved from the node server to the mount module. The Mounter class is no longer based on the k8s library code.

*Issue #, if available:* N/A

*Description of changes:*
Refactoring


pkg/driver/mount.go
```diff
@@ -39,27 +39,95 @@ const (
 	defaultRegionEnv   = "AWS_DEFAULT_REGION"
 	stsEndpointsEnv    = "AWS_STS_REGIONAL_ENDPOINTS"
 	MountS3PathEnv     = "MOUNT_S3_PATH"
-	hostTokenPathEnv   = "HOST_TOKEN_PATH"
+	hostPluginDirEnv   = "HOST_TOKEN_DIR"
```
dlakhaws, line 42: this is a bit confusing, plugin dir but variable is token dir?

pkg/driver/node.go
```diff
@@ -99,19 +108,27 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
...
+		hostPluginDirEnv = "/var/lib/kubelet/plugins/s3.csi.aws.com/"
 	}
-	if !mounted {
-		klog.V(4).Infof("NodePublishVolume: mounting %s at %s with options %v", bucket, target, mountpointArgs)
-		if err := d.Mounter.Mount(bucket, target, fstype, mountpointArgs); err != nil {
-			os.Remove(target)
-			return nil, status.Errorf(codes.Internal, "Could not mount %q at %q: %v", bucket, target, err)
-		}
-		klog.V(4).Infof("NodePublishVolume: %s was mounted", target)
+	hostTokenPath := path.Join(hostPluginDirEnv, "token")
```
dlakhaws, line 117: do we need to do any validation that `hostPluginDirEnv` ends in `/` or is that safe to assume?

jjkr: `path.Join` takes care of that for you. It handles either one.

dlakhaws (approved):

pkg/driver/mount.go
jjkr: Changed variable name to match

dlakhaws (approved):

Issue #86, opened 2023-11-29
Support Bottlerocket OS


yubingjiaocn:
/feature

**Is your feature request related to a problem? Please describe.**

When I'm trying to mount S3 bucket to a pod running on Bottlerocket OS worker node, mount failed and the error on pod events said:  

`
MountVolume.SetUp failed for volume "<Redacted>" : rpc error: code = Internal desc = Could not mount "<Redacted>" at "/var/lib/kubelet/pods/<Redacted>/volumes/kubernetes.io~csi/<Redacted>/mount": Mount failed: Failed to start systemd unit on host: SELinux policy denies access: Permission denied output:
`

System Info from `kubectl describe node`:
```
  Kernel Version:             5.15.136
  OS Image:                   Bottlerocket OS 1.16.1 (aws-k8s-1.27-nvidia)
  Operating System:           linux
  Architecture:               amd64
  Container Runtime Version:  containerd://1.6.24+bottlerocket
  Kubelet Version:            v1.27.7-eks-1670f88
  Kube-Proxy Version:         v1.27.7-eks-1670f88
```

**Describe the solution you'd like in detail**

I know Bottlerocket is not at the first batch of support. Is there any workaround or ETA to support Bottlerocket?

jjkr:
Thank you for the feature request. I do not have a workaround or ETA to share at this time unfortunately, but Bottlerocket is a great option for a k8s OS and I agree it would be awesome if the driver added compatibility.

As you're seeing it's the SELinux configuration that prevents this from functioning currently. Bottlerocket is security focused, so has a very locked down SELinux configuration that the CSI driver will need to work with specifically.

yubingjiaocn:
Hi @nalshamaajc . Tweaking SELinux on Bottlerocket seems is a good approach, but I'm not familiar with SELinux and don't know what to adjust. Per [Bottlerocket doc](https://github.com/bottlerocket-os/bottlerocket/blob/develop/SECURITY_GUIDANCE.md#limit-use-of-privileged-selinux-labels), define `super_t` label to CSI node container may be a way to grant permission. Or do you have any specific label to attach? Thx.

saikumarch7548:
@yubingjiaocn , @nalshamaajc  can you please help letting us know if you tired enabling SELinux on bottlerocket

evgeni-roif-gong:
[...]

PR #155, opened 2024-02-26
Add SELinux options for Bottlerocket

jjkr:
*Issue #, if available:* #86

*Description of changes:* Add support for Bottlerocket OS. New versions of Bottlerocket are adding SELinux policies and directories to support installing the CSI driver and this PR adds SELinux labels to the containers to work with these changes. The install and node driver containers must run with the `super_t` SELinux user. The Bottlerocket policies are configured such that the installed files will automatically get the correct labels applied, so it turns out there is no relabeling required.

### DIFF
--- a/charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
@@ -46,6 +46,14 @@ spec:
       initContainers:
         - name: install-mountpoint
           image: {{ printf "%s%s:%s" (default "" .Values.image.containerRegistry) .Values.image.repository (default (printf "v%s" .Chart.AppVersion) (toString .Values.image.tag)) }}
+          {{- with .Values.node.seLinuxOptions }}
+          securityContext:
+            seLinuxOptions:
+              user: {{ .user }}
+              type: {{ .type }}
+              role: {{ .role }}
+              level: {{ .level }}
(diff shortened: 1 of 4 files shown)


charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
```diff
@@ -46,6 +46,14 @@ spec:
       initContainers:
         - name: install-mountpoint
           image: {{ printf "%s%s:%s" (default "" .Values.image.containerRegistry) .Values.image.repository (default (printf "v%s" .Chart.AppVersion) (toString .Values.image.tag)) }}
+          {{- with .Values.node.seLinuxOptions }}
+          securityContext:
+            seLinuxOptions:
+              user: {{ .user }}
```
dlakhaws, line 52: suggestion to add default values to these

jjkr:
> Is there any specific example we can provide to customers in our examples directory?

As implemented, this will just work on newer Bottlerocket with no modifications. Once Bottlerocket cuts their release I will have to add them to the supported os matrix.

charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
jjkr: There are default values set in `values.yaml`

dlakhaws (approved):

PR #170, opened 2024-03-26
Replace HOST_TOKEN_DIR environment variable by HOST_PLUGIN_DIR

laghoule:
*Description of changes:*

* Remove duplicated empty node.resources values.
* The Helm chart and Kubernetes manifest are set up to use HOST_PLUGIN_DIR, but the mountpoint-s3-csi-driver utilizes HOST_TOKEN_DIR. Since HOST_PLUGIN_DIR and HOST_TOKEN_DIR serve the same purpose, we are replacing HOST_TOKEN_DIR with HOST_PLUGIN_DIR to standardize the configuration.

Tested in onPrem K8S lab (with IRSA), deployed with [K0S](https://github.com/k0sproject/k0s).

### DIFF
--- a/charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
@@ -91,8 +91,6 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
-            - name: HOST_TOKEN_DIR
-              value: {{ trimSuffix "/" .Values.node.kubeletPath }}/plugins/s3.csi.aws.com/
             - name: HOST_PLUGIN_DIR
               value: {{ trimSuffix "/" .Values.node.kubeletPath }}/plugins/s3.csi.aws.com/
             {{- with .Values.awsAccessSecret }}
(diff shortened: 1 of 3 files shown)


charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
```diff
@@ -91,6 +91,8 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
+            - name: HOST_TOKEN_DIR
```
jjkr, line 94: Thanks for catching this. My preference would actually be to eliminate the use of `HOST_TOKEN_DIR` in the code in favor of just reusing `HOST_PLUGIN_DIR` given they're set to the same thing. If you're not set up to make changes to the go code, we can fix this before the next release. Of course we're happy to take the contribution with an updated pr.

laghoule: Yeah, it's effectively a better approach.  I will update the PR with the modified code.
Thanks

jjkr (commented):
Code looks good. We have an issue with the CI on external PRs right now, so I'll check it on my end Monday morning and get this merged. Thank you!

jjkr (approved):

Issue #153, opened 2024-02-26
Support encryption using KMS key


justinas-b:
/feature

Hey! It would be very useful to have support for encryption using KMS key. Currently on S3 bucket policies we have below controls implemented:
- https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingServerSideEncryption.html
- https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingKMSEncryption.html

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "DenyIncorrectEncryptionHeader",
            "Effect": "Deny",
            "Principal": "*",
            "Action": "s3:PutObject",
            "Resource": "arn:aws:s3:::mybucket/*",
            "Condition": {
                "StringNotEquals": {
                    "s3:x-amz-server-side-encryption": "AES256"
                }
            }
        },
        {
            "Sid": "DenyUnEncryptedObjectUploads",
            "Effect": "Deny",
            "Principal": "*",
            "Action": "s3:PutObject",
            "Resource": "arn:aws:s3:::mybucket/*",
            "Condition": {
                "Null": { 
                    "s3:x-amz-server-side-encryption": "true"
                }
            }
        }
    ]
}
```

Even though we could read from S3/volumes, due to this limitation we cannot write to volumes/S3 buckets as there is no way to explicitly say that object should be encrypted when writing.

```
/app # echo "test" > /data/test.txt
sh: write error: I/O error
```

On `s3-csi-node` daemonset` i cannot find any related errors, even with increased verbosity.

justinas-b:
I see there is https://github.com/awslabs/mountpoint-s3/issues/534 in progress already to add support under `mountpoint-s3`, though once it is released i guess https://github.com/awslabs/mountpoint-s3-csi-driver will have to be updated

justinas-b:
Also, i see https://github.com/awslabs/mountpoint-s3/pull/715 was already merged, but i cannot see `--sse` nor `--sse-kms-key-id` flags in v1.4.1:

```
bash-4.2# mountpoint-s3/bin/mount-s3  --version
mount-s3 1.4.1
```

I am assuming those features were not released yet?

jjkr:
Thank you for the request. That's correct this feature is still in development in mountpoint itself and the issue you linked is the correct place to get updates on the current development. Once mountpoint is released, the CSI driver will make any necessary changes and release support shortly after.

[...]

PR #189, opened 2024-04-25
Update with Mountpoint 1.6.0 features

jjkr:
*Issue #, if available:* #153

*Description of changes:* Add examples and support for sse and max attempts options. Support for sse does not require any code changes because it is just an argument passthrough. The max attempts option is passed as an environment variable to mountpoint, so some conversion is needed from the mount option.


pkg/driver/mount.go
```diff
@@ -33,16 +33,18 @@ import (
...
+	accessKeyEnv         = "AWS_SECRET_ACCESS_KEY"
+	regionEnv            = "AWS_REGION"
+	defaultRegionEnv     = "AWS_DEFAULT_REGION"
+	stsEndpointsEnv      = "AWS_STS_REGIONAL_ENDPOINTS"
+	MountS3PathEnv       = "MOUNT_S3_PATH"
+	awsMaxAttemptsEnv    = "AWS_MAX_ATTEMPTS"
+	defaultMountS3Path   = "/usr/bin/mount-s3"
+	procMounts           = "/host/proc/mounts"
+	userAgentPrefix      = "--user-agent-prefix"
+	awsMaxAttemptsOption = "--aws-max-attempts"
```
dannycjones, line 46: Why are we adding the max attempts configuration as an option as well as an environment variable? What makes it different from something like `AWS_REGION` or `AWS_ACCESS_KEY_ID`?

jjkr: We need max attempts to be configurable by the user on a per PersistentVolume basis, so we have to accept it similar to other mount options. Unlike the other mount options, it's not just passed through to mountpoint and we have to do a conversion to remove the option from the CLI flags and pass it as an environment variable instead.

`AWS_REGION` we get from the driver's environment and that's passed through unchanged to mountpoint. `AWS_ACCESS_KEY_ID` is also unique where we get the key from a k8s secret and pass that through to the driver. Overall there's just different ways that the CSI driver gets options with different ways mountpoint accepts them and the driver has to glue everything together.

dannycjones (changes requested):
Docs change, then lgtm

dannycjones (approved):
LGTM!

Issue #127, opened 2024-01-18
Unmounting is not successful and the deletion of the deployment gets stuck


ruiarodrigues:
/kind bug

**What happened?**
I'm using Minikube locally on my laptop and deployed a pod that uses a mounted bucket using the driver. When I delete the pod (and the PV and PVC), the Kubernetes command get stuck.

```
% kubectl delete -f dfv.yaml      
persistentvolume "s3-pv" deleted
persistentvolumeclaim "s3-claim" deleted
pod "dfv-app" deleted
service "dfv-service" deleted
(stuck here forever)

```

In the logs of the driver, I can see that the bucket is mounted correctly but then the unmount with the same name fails stating that it's not mounted. But actually it is because I could access the files inside the bucket. If I remove the volumes mount from the yaml file, everything works fine.

```
I0118 09:34:11.527549       1 node.go:49] NodePublishVolume: called with args volume_id:"s3-csi-driver-volume" target_path:"/var/lib/kubelet/pods/e8ee9200-c877-4b39-b3d8-a3aa7de7c772/volumes/kubernetes.io~csi/s3-pv/mount" volume_capability:<mount:<mount_flags:"allow-delete" mount_flags:"region Lisbon" mount_flags:"force-path-style" mount_flags:"endpoint-url http://192.168.5.2:9000" > access_mode:<mode:MULTI_NODE_MULTI_WRITER > > volume_context:<key:"bucketName" value:"s3.select" > 
I0118 09:34:11.527601       1 node.go:81] NodePublishVolume: creating dir /var/lib/kubelet/pods/e8ee9200-c877-4b39-b3d8-a3aa7de7c772/volumes/kubernetes.io~csi/s3-pv/mount
I0118 09:34:11.527650       1 node.go:108] NodePublishVolume: mounting s3.select at /var/lib/kubelet/pods/e8ee9200-c877-4b39-b3d8-a3aa7de7c772/volumes/kubernetes.io~csi/s3-pv/mount with options [--allow-delete --endpoint-url=http://192.168.5.2:9000 --force-path-style --region=Lisbon]
[...]

PR #191, opened 2024-04-29
allows to configure proc-mounts hostPath

mmoscher:
Presents possible solution to Issue #127 

**Description of changes:**
Adds additional helm value to configure /proc/mounts. More precisely it allows changing the hostPath of the volumeMount "proc-mounts".

**Reason:**
On some hosts, eg. Ubuntu-22.04+k3s/Debian+k3s, /proc/mounts is a symlink and resolves to /proc/self/mounts.
This results in unexpected behavior, since the s3-plugin container as part of the daemon-set would not see all mounts. Therefore, mountpoint-s3 is unable to unmount its volumes.

In our case, ie. running mountpoint-s3-csi-driver on Ubuntu-22.04+k3s, it only works when configuring /proc/1/mounts.


mmoscher:
@jjkr, @muddyfish and @laghoule any chance to get this PR merged? You were mentioned because you three have the most recent posts. Thanks for your response!

laghoule:
@mmoscher, only @jjkr can merge.

dannycjones:
Thank you for the contribution, @mmoscher! This will be pulled into the next release alongside [Mountpoint v1.7.0](https://github.com/awslabs/mountpoint-s3/releases/tag/mountpoint-s3-1.7.0).

PR #195, opened 2024-05-07
Run unit tests as root in Github actions

muddyfish:
*Issue #, if available:* N/A

*Description of changes:* Run unit tests as root as part of the CI. Previously, some tests weren't running because they had an ['is running as root' check](https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/8938019603/job/24551349632?pr=193#step:8:78).

### DIFF
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -24,4 +24,4 @@ jobs:
       run: make bin
 
     - name: Test
-      run: make test
+      run: sudo make test
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -24,4 +24,5 @@ jobs:
       run: make bin
 
     - name: Test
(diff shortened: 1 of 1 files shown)


.github/workflows/unit-tests.yaml
```diff
@@ -24,4 +24,4 @@ jobs:
       run: make bin
 
     - name: Test
-      run: make test
+      run: sudo make test
```
dannycjones, line 27: just to prevent us removing it in future, can you include a small comment about why we're using `sudo`?

specifically, we want to run as root for systemd tests https://github.com/awslabs/mountpoint-s3-csi-driver/blob/5cd426dc6c884020d445e7c2f9c1f94de8a92528/pkg/system/systemd_test.go#L180-L182

dannycjones (approved):
LGTM!

jjkr (approved):

PR #201, opened 2024-05-17
Fix unit test workflow when running from forks

muddyfish:
*Issue #, if available:* NA

*Description of changes:* https://github.com/awslabs/mountpoint-s3-csi-driver/pull/199 has it's unit test workflow fail due to being created from a forked repository. This is due to `pull_request` having reduced permissions given when running from forked repositories. Using `pull_request_target` runs from the context of the fork, and still allows us to post comments to the PR. The unit test workflow doesn't make use of any AWS resources, so this is a trivial change.

### DIFF
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -7,8 +7,7 @@ on:
     branches: [ "main", "feature/*" ]
 
 permissions:
-  id-token: write
-  contents: read
+  pull-requests: write
 
 jobs:
   build:
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -3,8 +3,12 @@ name: Unit tests
 on:
   push:
     branches: [ "main" ]
-  pull_request:
-    branches: [ "main" ]
+  pull_request_target:
+    branches: [ "main", "feature/*" ]
+
+permissions:
+  id-token: write
+  contents: read
 
 jobs:
   build:
@@ -44,7 +48,7 @@ jobs:
         path: cover.html
 
     - name: Comment test coverage
-      if: ${{ github.event_name == 'pull_request' }}
+      if: ${{ github.event_name == 'pull_request_target' }}
       uses: actions/github-script@v7
       env:
         TOTAL_COVERAGE: ${{ steps.go-test-coverage.outputs.total-coverage }}
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [ "main" ]
   pull_request_target:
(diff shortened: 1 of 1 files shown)


dannycjones (changes requested):
I believe we need to have another think about how to post the coverage outcome. (Let's discuss at standup.)

Other ideas:
- Use a notice message: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-notice-message
- Use a job summary: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary

.github/workflows/unit-tests.yaml
```diff
@@ -3,9 +3,12 @@ name: Unit tests
 on:
   push:
     branches: [ "main" ]
-  pull_request:
+  pull_request_target:
     branches: [ "main" ]
```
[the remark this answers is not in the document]
dannycjones: Can we switch back to `pull_request` event since we no longer need any additional permissions?

.github/workflows/unit-tests.yaml
```diff
@@ -3,14 +3,16 @@ name: Unit tests
...
 
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
```
dannycjones, line 15: We can drop this too if switching back to `pull_request` event (and flip the `if` below back`).

dannycjones (changes requested):
I think we can flip back to `pull_request` event, and we should be able to merge (and see the new output)

.github/workflows/unit-tests.yaml
```diff
@@ -49,11 +49,6 @@ jobs:
...
-        script: |
-          github.rest.issues.createComment({
-            issue_number: context.issue.number,
-            owner: context.repo.owner,
-            repo: context.repo.repo,
-            body: 'Total test coverage: ${{ env.TOTAL_COVERAGE }}%\n Download report: ${{ env.ARTIFACT_URL }}'
-          })
+      run: |
+        echo "### Total test coverage: ${{ env.TOTAL_COVERAGE }}%" >> $GITHUB_STEP_SUMMARY
+        echo "Download report: ${{ env.ARTIFACT_URL }}" >> $GITHUB_STEP_SUMMARY
```
dannycjones, line 54: We need to drop the `uses: actions/github-script@v7` above.

dannycjones (changes requested):
Just need to fix the step def, then we should see job summary

dannycjones (approved):
LGTM!

You can view summary here: https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/9172012983/attempts/1#summary-25217617970

PR #214, opened 2024-06-26
Move `maxAttempts` to more generic handling code for adding environment variables

muddyfish:
*Description of changes:*

Refactors code that handles `maxAttempts` such that it's more reusable

### DIFF
--- a/pkg/driver/mount.go
+++ b/pkg/driver/mount.go
@@ -239,20 +239,7 @@ func (m *S3Mounter) Mount(bucketName string, target string,
 	if credentials != nil {
 		env = credentials.Env()
 	}
-
-	// Max attempts is passed to the driver as a mount option, but is passed to mp via env variable
-	// so we need to remove it from the options and add it to the env.
-	maxAttemptsIdx := -1
-	for i, o := range options {
-		if strings.HasPrefix(o, awsMaxAttemptsOption) {
-			maxAttemptsIdx = i
-			break
-		}
-	}
-	if maxAttemptsIdx != -1 {
-		env = append(env, strings.Replace(options[maxAttemptsIdx], awsMaxAttemptsOption, awsMaxAttemptsEnv, 1))
-		options = append(options[:maxAttemptsIdx], options[maxAttemptsIdx+1:]...)
-	}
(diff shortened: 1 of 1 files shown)


pkg/driver/mount.go
```diff
@@ -272,6 +259,23 @@ func (m *S3Mounter) Mount(bucketName string, target string,
 	return nil
 }
 
+func addOptionToEnvironmentVariables(optionName string, envName string, options []string, env []string) ([]string, []string) {
```
passaro, line 262: We should add some doc for this function: what's in the comment below is a start, plus what it is returning.

passaro: nit: `addOptionToEnvironmentVariables` or rather something with `move...` or `convert..`?

`moveOptionToEnvironment`?

passaro: nit: return order, should it be consistent with the parameters? so `options, env`?

pkg/driver/mount.go
```diff
@@ -272,6 +259,23 @@ func (m *S3Mounter) Mount(bucketName string, target string,
...
+	// so we need to remove it from the options and add it to the env.
+	optionIdx := -1
+	for i, o := range options {
+		if strings.HasPrefix(o, optionName) {
+			optionIdx = i
+			break
+		}
+	}
+	if optionIdx != -1 {
+		env = append(env, strings.Replace(options[optionIdx], optionName, envName, 1))
```
passaro, line 273: `Replace` is safe here because we just checked `HasPrefix`, right? May be worth a short comment.

passaro (commented):
Nice change, just a couple of minor suggestions.

pkg/driver/mount.go
```diff
@@ -272,6 +259,25 @@ func (m *S3Mounter) Mount(bucketName string, target string,
 	return nil
 }
 
+func moveOptionToEnvironmentVariables(optionName string, envName string, options []string, env []string) ([]string, []string) {
+	// Moves a parameter optionName from the options list to MP's environment variable list. We need this as options are
```
passaro, line 263: Looks good. But shouldn't it be moved above `func`? Or am I showing my poor Go knowledge?

muddyfish: No, I'm just too used to languages where you put the comment after the definition 😅

passaro (approved):
LGTM!

PR #216, opened 2024-07-04
Fix a typo in the e2e test suite

muddyfish:
*Description of changes:*

The e2e test "should concurrently access the single volume from pods on different node" did not include an argument for running the pods on different nodes - there was no difference between it and the test for "should concurrently access the single volume from pods on the same node".


passaro (approved):
LGTM

PR #223, opened 2024-07-18
Initial implementation of multicredential support

muddyfish:
*Issue #, if available:*

*Description of changes:*

Implements a proof of concept of pod level identity.

Todo:

- Implement volume level switch
- Tests
- Get the AWS region from either IMDS or from helm
- Update `deploy` k8s resources
- Error messages


pkg/driver/node.go
```diff
@@ -110,21 +114,71 @@ func (ns *S3NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePubl
...
+			var tokens map[string]*Token
+			if err := json.Unmarshal([]byte(tokensJson), &tokens); err != nil {
+				return nil, status.Error(codes.InvalidArgument, "Tokens bad format")
+			}
+			stsToken := tokens["sts.amazonaws.com"]
+			if stsToken != nil {
+				klog.V(4).Infof("NodePublishVolume: stsToken exp: %v", stsToken.ExpirationTimestamp)
+				hostTokenPath = path.Join(hostPluginDir, volumeID+".token")
+				// TODO cleanup these files on unmount and startup
+				os.WriteFile(path.Join("/csi/", volumeID+".token"), []byte(stsToken.Token), 0400)
```
unexge, line 141: Shouldn't we write to `hostTokenPath` here?

unexge (approved):

pkg/driver/node.go
muddyfish: No - we're writing from inside the CSI Driver container here, which mounts `/csi/` as `{{ trimSuffix "/" .Values.node.kubeletPath }}/plugins/s3.csi.aws.com/` in the underlying host.

Issue #174, opened 2024-03-31
Pod "Sometimes" cannot mount PVC in CSI version 1.4.0


zhushendhh:
Hello Team,

I am trying to test CSI driver 1.4.0 in K8s 1.27. But I found "Some Times" the Pod cannot mount the PVC and the CSI driver Pod reports below error:

```
I0331 13:53:37.687569       1 node.go:65] NodePublishVolume: req: volume_id:"s3-csi-driver-volume" target_path:"/var/lib/kubelet/pods/adaac085-e689-49f8-b9f5-d0467907d875/volumes/kubernetes.io~csi/comfyui-outputs-pv/mount" volume_capability:<mount:<mount_flags:"allow-delete" mount_flags:"region us-east-2" > access_mode:<mode:MULTI_NODE_MULTI_WRITER > > volume_context:<key:"bucketName" value:"comfyui-outputs-835894076989-us-east-2" > 
I0331 13:53:37.687634       1 node.go:112] NodePublishVolume: mounting comfyui-outputs-835894076989-us-east-2 at /var/lib/kubelet/pods/adaac085-e689-49f8-b9f5-d0467907d875/volumes/kubernetes.io~csi/comfyui-outputs-pv/mount with options [--allow-delete --region=us-east-2]
E0331 13:53:37.687730       1 driver.go:96] GRPC error: rpc error: code = Internal desc = Could not mount "comfyui-outputs-835894076989-us-east-2" at "/var/lib/kubelet/pods/adaac085-e689-49f8-b9f5-d0467907d875/volumes/kubernetes.io~csi/comfyui-outputs-pv/mount": Could not check if "/var/lib/kubelet/pods/adaac085-e689-49f8-b9f5-d0467907d875/volumes/kubernetes.io~csi/comfyui-outputs-pv/mount" is a mount point: stat /var/lib/kubelet/pods/adaac085-e689-49f8-b9f5-d0467907d875/volumes/kubernetes.io~csi/comfyui-outputs-pv/mount: no such file or directory, Failed to read /host/proc/mounts: open /host/proc/mounts: invalid argument
```

And even in the case of normal mounting, the CSI driver Pod still posting such of the above logs. Seems like the CSI driver keep tring to mount the same PVC to the sam Pod and failed. Not sure whether there is anything is misconfig.

Those problems only happend with "Karpenter" scale up worker nodes with new deployment/pod in EKS v1.27.
All mount operations are normal for static k8s worker nodes.

[...]

PR #229, opened 2024-07-23
Fix cannot mount PVC in driver (#174)

Shellmode:
*Issue #, if available:*
[174](https://github.com/awslabs/mountpoint-s3-csi-driver/issues/174)

*Description of changes:*
Since it's unclear why the driver cannot read `/host/proc/mounts` (with Karpenter) at the first time, retry to read `/host/proc/mounts` will success.

It is necessary to keep investigating the root cause, but this issue currently affects many Karpenter (and possibly other) users. To prevent users from having to manually restart `s3-csi-node-xxx` every time, it is better to solve the issue by retrying the read in the code.

### DIFF
--- a/pkg/driver/mount.go
+++ b/pkg/driver/mount.go
@@ -157,9 +157,14 @@ func NewS3Mounter(mpVersion string, kubernetesVersion string) (*S3Mounter, error
 }
 
 func (pml *ProcMountLister) ListMounts() ([]mount.MountPoint, error) {
+	retryTimes := 3
 	mounts, err := os.ReadFile(pml.ProcMountPath)
-	if err != nil {
-		return nil, fmt.Errorf("Failed to read %s: %w", procMounts, err)
+	if err != nil && retryTimes == 0 {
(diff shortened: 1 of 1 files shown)


Shellmode:
Passed locally test
image

pkg/driver/mount.go
```diff
@@ -157,9 +157,14 @@ func NewS3Mounter(mpVersion string, kubernetesVersion string) (*S3Mounter, error
...
 
 func (pml *ProcMountLister) ListMounts() ([]mount.MountPoint, error) {
+	retryTimes := 3
 	mounts, err := os.ReadFile(pml.ProcMountPath)
-	if err != nil {
-		return nil, fmt.Errorf("Failed to read %s: %w", procMounts, err)
+	if err != nil && retryTimes == 0 {
+		return nil, fmt.Errorf("Failed to read %s after %d tries: %w", procMounts, retryTimes, err)
+	} else {
+		retryTimes -= 1
```
unexge, line 165: Seems like we only retry once as there is no loops or recursion

PR #237, opened 2024-08-07
Pod-level identity improvements

unexge:
This PR adds the following improvements:

- Ensure we don't log sensitive information from `csi.NodePublishVolumeRequest`
- Disable caching with pod-level credentials
- Ensure to disable all other credential providers except STS provider
- Add a test case to make sure we're not using secrets if pod-level identity is enabled

---

### DIFF
--- a/tests/e2e-kubernetes/testsuites/credentials.go
+++ b/tests/e2e-kubernetes/testsuites/credentials.go
@@ -201,8 +201,9 @@ func (t *s3CSICredentialsTestSuite) DefineTests(driver storageframework.TestDriv
 		expectListToSucceed(pod)
 	}
 
-	expectFailToMount := func(ctx context.Context, withServiceAccountName string) {
-		vol := createVolume(ctx)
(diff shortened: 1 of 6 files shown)


tests/e2e-kubernetes/testsuites/credentials.go
```diff
@@ -243,50 +244,87 @@ func (t *s3CSICredentialsTestSuite) DefineTests(driver storageframework.TestDriv
...
+			sa, restoreServiceAccountRole := overrideServiceAccountRole(ctx, f, sa, *role.Arn)
+			deferCleanup(restoreServiceAccountRole)
+
+			waitUntilRoleIsAssumableWithWebIdentity(ctx, f, sa)
+
+			// Trigger recreation of our pods to use the new IAM role
+			killCSIDriverPods(ctx, f)
+		}
+
+		updateDriverLevelCredentials := func(ctx context.Context, policyARN string) {
```
muddyfish, line 311: Can we call this something like `updateDriverLevelK8sSecret`? The current name doesn't make it obvious

PR #246, opened 2024-08-30
Add retry to reading `/proc/mounts` out of ListMounts() function

Shellmode:
*Issue #, if available:*
https://github.com/awslabs/mountpoint-s3-csi-driver/issues/174

*Description of changes:*
Since it's unclear why the driver cannot read /host/proc/mounts (with Karpenter) at the first time, retry to read /host/proc/mounts will success.
We've discussed in https://github.com/awslabs/mountpoint-s3-csi-driver/pull/229

Recently released v1.8 added retry in the `ListMounts() function`, however I tried the new release and got the same error message still cannot mount S3. I found that if `ListMounts() function` ever return a `nil, error` it won't work.

Just leave error handling in `parseProcMounts() function` and retry reading `/proc/mounts` by calling `ListMounts() function` from other function, which will work.

It may be somehow confusing about **"retry"**, it may because other function/module refresh/restart which fix the issue (just like restart the pod).

Passed locally test
image

### DIFF
--- a/pkg/driver/mount.go
+++ b/pkg/driver/mount.go
@@ -47,15 +47,12 @@ const (
 	awsMaxAttemptsOption = "--aws-max-attempts"
 )
 
-const (
-	// Due to some reason that we haven't been able to identify, reading `/host/proc/mounts`
-	// fails on newly spawned Karpenter/GPU nodes with "invalid argument".
-	// It's reported that reading `/host/proc/mounts` works after some retries,
-	// and we decided to add retry mechanism until we find and fix the root cause of this problem.
-	// See https://github.com/awslabs/mountpoint-s3-csi-driver/issues/174.
-	procMountsReadMaxRetry     = 3
-	procMountsReadRetryBackoff = 100 * time.Millisecond
-)
+// Due to some reason that we haven't been able to identify, reading `/host/proc/mounts`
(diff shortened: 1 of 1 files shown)


pkg/driver/mount.go
```diff
@@ -177,21 +174,11 @@ func (pml *ProcMountLister) ListMounts() ([]mount.MountPoint, error) {
...
-
-		klog.Errorf("Failed to read %s on try %d: %v", pml.ProcMountPath, i, err)
-		time.Sleep(procMountsReadRetryBackoff)
-	}
+	mounts, err = os.ReadFile(pml.ProcMountPath)
 
 	if err != nil {
-		return nil, fmt.Errorf("Failed to read %s after %d tries: %w", pml.ProcMountPath, procMountsReadMaxRetry, err)
+		klog.Errorf("Failed to read %s on try %d: %v", pml.ProcMountPath, ProcMountsReadRetry, err)
+		ProcMountsReadRetry += 1
```
unexge, line 181: This block basically ignores the error and returns an empty response. Although it won't fail, some operations depending on this result e.g., unmounting and skipping already mounted paths won't work.

dannycjones:
We'll close this for now due to inactivity.

PR #251, opened 2024-09-09
Clean up tokens by parsing target path in `NodeUnpublishVolume`

unexge:
We need Pod ID and Volume ID to clean up any previously created
tokens. `NodeUnpublishVolume` call receives Volume ID but not Pod ID.
We were keeping the mapping from target path to Pod ID in memory
before but this is not ideal as it's not persistent and we were losing
this data when CSI driver restarts. With this change, we're parsing
target path to get Pod ID and use that to clean up tokens without
storing any state in the CSI driver.

---


pkg/driver/target_path.go
```diff
@@ -0,0 +1,54 @@
...
+	PodID    string
+	VolumeID string
+}
+
+// ParseTargetPath parses given target path from Kubernetes.
+// Target paths are generated by Kubernetes for CSI drivers to mount their filesystem.
+// They are in the form of "/var/lib/kubelet/pods/<pod-uuid>/volumes/kubernetes.io~csi/<volume-id>/mount".
+func ParseTargetPath(path string) (TargetPath, error) {
+	path, mount := filepath.Split(path)
+	if path == "" || mount != "mount" {
```
muddyfish, line 23: Is it possible to put this all in a for loop popping off a list? It's very repetitive right now

unexge: I feel like using a loop here would make it less readable. I can use a RegExp here, I think that would be more readable.

unexge: Updated to use a RegExp with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/251/commits/422d7322788e9ed3d0215bb6382c98c903eb6896. Seems more readable now

pkg/driver/target_path.go
```diff
@@ -2,13 +2,19 @@ package driver
...
-	"path/filepath"
+	"regexp"
 )
 
 var (
 	ErrInvalidTargetPath = errors.New("ParseTargetPath: Invalid target path")
 )
 
+var (
+	targetPathRegexp              = regexp.MustCompile("pods/(?P<podid>.+)/volumes/kubernetes.io~csi/(?P<volumeid>.+)/mount$")
```
muddyfish, line 13: Regex shouldn't allow `podid` or `volumeid` to contain `/`. Also, should start with a `/`:
```
/pods/(?P<podid>[^/]+)/volumes/kubernetes.io~csi/(?P<volumeid>[^/]+)/mount$
```

muddyfish (approved):

PR #253, opened 2024-09-11
Add authentication source to user-agent

unexge:


pkg/driver/version.go
```diff
@@ -78,5 +79,12 @@ func UserAgent(kubernetesVersion string) string {
 		b.WriteString(kubernetesVersion)
 	}
 
+	if authenticationSource != "" {
```
muddyfish, line 82: Authentication source should never be empty, so it's probably fine to remove this

muddyfish (approved):

PR #255, opened 2024-09-12
Feature/pod level identity docs

muddyfish:
*Issue #, if available:*

*Description of changes:*

Documentation for pod level identity feature

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -19,11 +19,16 @@ See [install.md](./install.md) for more details on this approach.
 
 **TODO** - Migrate driver level setup here.
 
+### Driver Level Credentials with IRSA
+
+Configuring [IAM Roles for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) (IRSA)
+is the recommended way to set up the CSI Driver if you want to use Driver Level credentials.
+
 ```mermaid
 graph LR;
     CSI[CSI Driver]
 
-    P["`Pod
+    P["`Application Pod
     *Driver IAM Credentials*`"]
 
     SA_D[Service Account - Driver]
@@ -48,7 +53,80 @@ graph LR;
     SA_P --> IAM_P
 ```
 
-### Pod Level Credentials
+### Driver Level Credentials with K8s Secrets
+
+The Mountpoint CSI Driver also supports sourcing long term AWS credentials from k8s secrets, though this is not 
+recommended.
+
+```mermaid
+graph LR;
+    CSI[CSI Driver]
+
+    P["`Application Pod
+    *K8s Secret Credentials*`"]
+
+    K8sS[K8s Secret]
+
+    IAM_LT["Long Term IAM Credentials"]
+
+    PV[Persistent Volume]
+    PVC[PVC]
+
+    P --> PVC
+
+    PVC --> PV
+
+    PV --> CSI
+
+    CSI --> K8sS
+
+    K8sS --> IAM_LT
+```
+
+By default, the CSI Driver checks the existence of a secret `aws-secret` in the installation namespace 
+(default `kube-system`). 
+The secret name configurable if installing with helm: `awsAccessSecret.name`, and the installation namespace is 
+configurable with the `--namespace` helm parameter.
+
+```bash
+kubectl create secret generic aws-secret \
+    --namespace kube-system \
+    --from-literal "key_id=${AWS_ACCESS_KEY_ID}" \
+    --from-literal "access_key=${AWS_SECRET_ACCESS_KEY}"
+```
+
+To use K8s secrets for authentication, the secret must exist before installation, or the CSI Driver pods must be 
+restarted to use the secret.
+
+### Driver Level Credentials with Node IAM Profiles
+
+To use an IAM [instance profile](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2_instance-profiles.html),
+attach the policy to the instance profile IAM role and turn on access to [instance metadata](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html)
+for the instance(s) on which the driver will run.
+
+```mermaid
+graph LR;
+    CSI[CSI Driver]
+
+    P["`Application Pod
+    *Node Instance Profile*`"]
+
+    IAM_Instance["Instance IAM Credentials"]
+
+    PV[Persistent Volume]
+    PVC[PVC]
+
+    P --> PVC
+
+    PVC --> PV
+
+    PV --> CSI
+
+    CSI --> IAM_Instance
+```
+
+
+## Pod Level Credentials
 
 Alternatively, you can configure the CSI Driver to use a pod's Service Account on your volumes. 
 With this approach, a multi-tenant architecture is possible using [IAM Roles for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) (IRSA).
@@ -59,7 +137,7 @@ each pod.
 graph LR;
     CSI[CSI Driver]
 
-    P["`Pod
+    P["`Application Pod
     *Pod IAM Credentials*`"]
 
     SA_D[Service Account - Driver]
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -8,21 +8,30 @@ Mountpoint specific configuration.
 The Mountpoint CSI Driver can be configured to ingest credentials via two approaches: globally for the entire 
 Kubernetes cluster, or using credentials assigned to pods.
 
-### Driver Level Credentials
+### Driver-Level Credentials
 
 By setting driver-level credentials, the whole cluster uses the same set of credentials. 
 
 Using this configuration, the credentials that are used are set at installation time, either using Service
 Accounts, or using K8s secrets. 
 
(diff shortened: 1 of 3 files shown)


docs/CONFIGURATION.md
```diff
@@ -0,0 +1,188 @@
...
+See [install.md](./install.md) for more details on this approach.
+
+**TODO** - Migrate driver level setup here.
+
+### Driver Level Credentials with IRSA
+
+Configuring [IAM Roles for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) (IRSA)
+is the recommended way to set up the CSI Driver if you want to use Driver Level credentials.
+
+```mermaid
```
unexge, line 27: It would be nice to set a different color on credential source in the diagram if possible. i.e., "IAM Credentials - Driver" would be in a different color here to make it easy to understand

docs/CONFIGURATION.md
```diff
@@ -0,0 +1,188 @@
...
+    *Driver IAM Credentials*`"]
+
+    SA_D[Service Account - Driver]
+    SA_P[Service Account - Pod]
+
+    IAM_D[IAM Credentials - Driver]
+    IAM_P[IAM Credentials - Pod]
+
+    PV[Persistent Volume]
+    PVC[PVC]
```
unexge, line 41:
```suggestion
    PVC[Persistent Volume Claim]
```
to be consistent with the others

docs/CONFIGURATION.md
```diff
@@ -0,0 +1,188 @@
...
+Using this configuration, the credentials that are used are set at installation time, either using Service
+Accounts, or using K8s secrets. 
+
+See [install.md](./install.md) for more details on this approach.
+
+**TODO** - Migrate driver level setup here.
+
+### Driver Level Credentials with IRSA
+
+Configuring [IAM Roles for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) (IRSA)
```
unexge, line 24: Maybe we can briefly explain how this option can be used (i.e., setting IAM role on CSI Driver's service account)

docs/CONFIGURATION.md
```diff
@@ -0,0 +1,188 @@
...
+
+```bash
+kubectl create secret generic aws-secret \
+    --namespace kube-system \
+    --from-literal "key_id=${AWS_ACCESS_KEY_ID}" \
+    --from-literal "access_key=${AWS_SECRET_ACCESS_KEY}"
+```
+
+To use K8s secrets for authentication, the secret must exist before installation, or the CSI Driver pods must be 
+restarted to use the secret.
```
unexge, line 99: It's already implied with the current sentence – but it would be nice to mention that the credentials won't be refreshed once read.

docs/CONFIGURATION.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Driver Level Credentials with Node IAM Profiles
```
unexge, line 101: We don't order things by recommendation – but I prefer this option being before then long-term credentials

docs/CONFIGURATION.md
```diff
@@ -0,0 +1,188 @@
...
+
+The Mountpoint CSI Driver can be configured to ingest credentials via two approaches: globally for the entire 
+Kubernetes cluster, or using credentials assigned to pods.
+
+### Driver Level Credentials
+
+By setting driver-level credentials, the whole cluster uses the same set of credentials. 
+
+Using this configuration, the credentials that are used are set at installation time, either using Service
+Accounts, or using K8s secrets.
```
unexge, line 16: It would be nice to mention the order CRT uses to detect credentials. So customers would know e.g., if they set long-term credentials and IRSA, long-term credentials would always take precedence and IRSA won't be used.

docs/CONFIGURATION.md
```diff
@@ -0,0 +1,188 @@
...
+
+    CSI --> IAM_Instance
+```
+
+
+## Pod Level Credentials
+
+Alternatively, you can configure the CSI Driver to use a pod's Service Account on your volumes. 
+With this approach, a multi-tenant architecture is possible using [IAM Roles for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) (IRSA).
+Using Pod Level Credentials with IRSA authentication allows the Mountpoint CSI Driver to use multiple credentials for
```
unexge, line 133: Same with driver-level IRSA, it would be nice to explain how they can configure this option

unexge (approved):
Left some nit comments, feel free to ignore or address later

docs/CONFIGURATION.md
dannycjones: Agree, so long as they are configurations we're committed to support.

Probably I'd focus on the two mechanisms we intend to support, which is IRSA and Kubernetes secrets (and we recommend against K8s secrets).

docs/CONFIGURATION.md
```diff
@@ -0,0 +1,188 @@
...
+    P --> SA_P
+
+    SA_D --> IAM_D
+    SA_P --> IAM_P
+```
+
+### Driver Level Credentials with K8s Secrets
+
+The Mountpoint CSI Driver also supports sourcing long term AWS credentials from k8s secrets, though this is not 
+recommended.
```
dannycjones, line 59: I'd recommend rephrasing this more like:

```suggestion
Where IAM Roles for Service Accounts (IRSA) isn't a viable option, Mountpoint CSI Driver also supports sourcing static AWS credentials from K8s secrets.

> [!WARNING]
> We do not recommend using long-term AWS credentials and K8s secrets. Instead, we recommend using short-term credentials with IRSA.
```

docs/CONFIGURATION.md
dannycjones, line 101: Do we document this anywhere else? I'd be inclined to prefer IRSA over any other option, and not include this here.

docs/CONFIGURATION.md
```diff
@@ -0,0 +1,188 @@
...
+
+    PVC --> PV
+
+    PV --> CSI
+
+    CSI --> IAM_Instance
+```
+
+
+## Pod Level Credentials
```
dannycjones, line 129: nit:

We're using pod-level as an adjective, so I think we want it hyphenated. Same with driver-level.

docs/CONFIGURATION.md
```diff
@@ -0,0 +1,188 @@
...
+
+Alternatively, you can configure the CSI Driver to use a pod's Service Account on your volumes.
```
dannycjones, line 131: rephrase and newline?

```suggestion
You can configure Mountpoint CSI Driver to use the credentials associated with the pod's Service Account rather than the driver's own credentials.

```

dannycjones (commented):
Just some rough feedback, I've not reviewed the full content.

docs/CONFIGURATION.md
muddyfish: I'm changing this to
> We do not recommend using long-term AWS credentials. Instead, we recommend using short-term credentials with IRSA.

and removing the bit about k8s secrets, as I don't think we should be against recommending their usage in general

docs/CONFIGURATION.md
muddyfish: We currently document it in the installation document: https://github.com/awslabs/mountpoint-s3-csi-driver/blame/main/docs/install.md#L48

docs/CONFIGURATION.md
muddyfish: This is described below the diagram? Or am I misunderstanding you?

docs/CONFIGURATION.md
dannycjones: Ah, OK. I see also that with IRSA being EKS-only, then it makes sense to clarify this as an alternative.

docs/CONFIGURATION.md
unexge: I meant like mentioning they need to annotate their service account with IAM role ARN and configure OIDC provider for their cluster. Just adding some basic information and linking to EKS guide would be sufficient in my opinion.

docs/CONFIGURATION.md
muddyfish: IRSA isn't technically EKS-only - though it's a lot more involved setting it up outside EKS. You'd need to set up an OIDC server: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-issuer-discovery

If I understand it, the EKS team don't 'support' it, but it works.

docs/CONFIGURATION.md
```diff
@@ -0,0 +1,291 @@
...
+#### Service Account configuration for EKS Clusters
+
+EKS allows [using Kubernetes service accounts to authenticate requests to S3](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html). 
+To set this up follow these steps:
+
+##### Create a Kubernetes service account for the driver and attach the policy to the service account
+
+> [!NOTE]
+> The same service account name (`s3-csi-driver-sa`) must be specified both in this command and when creating a driver 
+> pod (in the pod spec `deploy/kubernetes/base/node-daemonset.yaml`, Helm value `node.serviceAccount.name`).
```
unexge, line 83: and `--service-account-role-arn` for `eksctl`/`aws eks`

unexge (approved):

docs/CONFIGURATION.md
muddyfish: This note is just about configuring the name of the service account rather than the ARN role

PR #258, opened 2024-09-13
Pod-level identity improvements

unexge:

### DIFF
--- a/pkg/driver/mount.go
+++ b/pkg/driver/mount.go
@@ -53,6 +53,9 @@ const (
 	awsMaxAttemptsOption        = "--aws-max-attempts"
 )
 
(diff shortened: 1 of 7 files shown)


pkg/driver/mount.go
```diff
@@ -230,6 +234,11 @@ func (m *S3Mounter) IsMountPoint(target string) (bool, error) {
 	}
 	for _, mp := range mountPoints {
 		if mp.Path == target {
+			if mp.Device != mountpointDeviceName {
```
muddyfish, line 237: Can we add a test for this?

unexge: Added unit-tests with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/258/commits/57b70d260c9b03ef0984e5f672cd74d9c5fee9bc. Wanted to add integration-tests as well, but we'd need to spawn a privileged Pod to make some assertions on the underlying host like checking `/proc/mounts` or something. Not sure if that's worth it

muddyfish (approved):

PR #267, opened 2024-09-27
Clarify only IAM role is created by eksctl in CONFIGURATION.md

dannycjones:
*Issue #, if available:* N/A

*Description of changes:*

I was following the instructions in order to create a cluster for testing. These changes will hopefully make it a bit clearer.

I also removed some validation steps as I think they're ultimately distracting and not appropriate at the step they were mentioned. They could be re-added elsewhere if there's a good home for them.


docs/CONFIGURATION.md
```diff
@@ -282,15 +288,7 @@ eksctl create iamserviceaccount \
     --role-only
 ```
 
-See [Pod-Level identity](https://github.com/awslabs/mountpoint-s3-csi-driver/tree/main/examples/kubernetes/static_provisioning/pod_level_identity.yaml)
-examples for how to set up Pod-Level identity with IRSA.
-
-##### [Optional] Validate the account was successfully created
```
unexge, line 288: I think this section is useful, it'd be better if we say “check if you have the right IAM Role ARN with `eks.amazonaws.com/role-arn` annotation.”

dannycjones: I'm not really sure what you mean. The problem I have with this section is that we aren't dealing with service accounts at this point, but I'm open to suggestions on where this can be useful.

dannycjones: I'm wondering if it makes more sense to include this when working on a troubleshooting document, which I think @muddyfish mentioned adding recently?

muddyfish: I included the troubleshooting stuff I was talking about in this document in the STS section at the bottom.

muddyfish: But this bit is where we're creating the service account, or at least the line above this one is.

dannycjones: >this bit is where we're creating the service account, or at least the line above this one is

The service account isn't created in this documentation, right? (Otherwise my changes really don't make sense). 

What I understood is that we create IAM roles that will later be used for service accounts, when created either when installing the CSI driver or when setting up the namespace for the pod.

dannycjones: @muddyfish can you check the thread above? I'd like to try and get this PR merged this week.

docs/CONFIGURATION.md
```diff
@@ -103,12 +103,18 @@ graph LR;
...
+This is supported for both driver-level and pod-level identity.
 
-##### Create a Kubernetes service account for the driver and attach the policy to the service account
+The following sections describe how to create the supporting resources for IRSA with driver-level identity.
 
-> [!NOTE]
+##### Create an IAM role for use by the CSI driver's service account
+
+The following command will use `eksctl` to create the IAM role that will be used by the CSI driver's service account.
+The service account is not created by this command, only the IAM role.
```
unexge, line 114:
```suggestion
The service account is not created by this command (due to `--role-only` option), only the IAM role.
```

docs/CONFIGURATION.md
```diff
@@ -103,12 +103,18 @@ graph LR;
 
 #### Service Account configuration for EKS Clusters
 
-EKS allows [using Kubernetes service accounts to authenticate requests to S3](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html).
-To set this up follow these steps:
+EKS allows using Kubernetes service accounts to authenticate requests to S3 using [IAM Roles for Service Accounts (IRSA)](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html).
+This is supported for both driver-level and pod-level identity.
```
unexge, line 107: Should driver-level and pod-level here link to corresponding sections?

unexge (dismissed):

docs/CONFIGURATION.md
dannycjones: Updated with links (in both occurrences).

unexge (approved):

PR #271, opened 2024-10-04
Mention one-to-one mapping of PVCs and PVs in the docs and examples

unexge:
The Driver only supports Static Provisioning as of today, and in order to configure it correctly, one needs to ensure there is no `storageClassName` set (it should be `""`) and also there is a one-to-one mapping between PVC and PV using `claimRef`. This PR clarifies this requirement and updates example manifests to specify one-to-one mapping between PVC and PV. See a recent issue regarding this: https://github.com/awslabs/mountpoint-s3-csi-driver/issues/249

--- 

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -3,6 +3,40 @@
 See [the Mountpoint documentation](https://github.com/awslabs/mountpoint-s3/blob/main/doc/CONFIGURATION.md) for
 Mountpoint specific configuration.
 
+## Static Provisioning
+
+The driver only supports Static Provisioning as of today, and you need an existing S3 Bucket to use.
+
+To use Static Provisioning, you should set `storageClassName` field of your PersistentVolume (PV) and PersistentVolumeClaim (PVC) to `""` (empty string).
+Also, in order to make sure no other PVCs can claim your PV, you should define a one-to-one mapping using `claimRef`:
+
+```yaml
+apiVersion: v1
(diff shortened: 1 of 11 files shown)


docs/CONFIGURATION.md
```diff
@@ -14,15 +48,15 @@ Mountpoint CSI driver. For more details on creating a policy and an IAM role, re
 ["Creating an IAM role"](https://docs.aws.amazon.com/eks/latest/userguide/s3-csi.html#s3-create-iam-role) from the
 EKS User Guide.
 
-The Mountpoint CSI Driver can be configured to ingest credentials via two approaches: globally for the entire 
+The Mountpoint CSI Driver can be configured to ingest credentials via two approaches: globally for the entire
```
muddyfish, line 51: Nit: Can these single character changes go in their own PR so it doesn't confuse the commit history?

muddyfish (approved):

docs/CONFIGURATION.md
unexge: I think that should be fine as this is a doc-only change. For long-term, we should probably look for enforcing a standard format for Go (`gofmt`), YAML and Markdown (`Prettier`).

dannycjones: I agree with Simon, if we do find any changes like this then let's just make two commits and merge them in separately.

Agree enforcing format checkers would be better, but I'd prefer we remove the noise from PRs and commit history regardless.

Issue #279, opened 2024-10-21
Run Mountpoint inside container


unexge:
Currently, the CSI Driver runs Mountpoint on the underlying host using `systemd` via it's D-Bus API. This approach has some downsides:
- Kubernetes/kubelet is not aware of the `systemd` process 
   - https://github.com/awslabs/mountpoint-s3-csi-driver/issues/82
   - It requires ssh'ing into the node and using `journalctl` to get Mountpoint logs
- The CSI Driver has implicit `systemd` dependency on the host it runs
- Due to writing to host filesystem and opening pseudo tty, the CSI Driver does not support SELinux
  - https://github.com/awslabs/mountpoint-s3-csi-driver/issues/169
  - https://github.com/awslabs/mountpoint-s3-csi-driver/issues/108
- It requires you to create cache directory on the underlying host using some workarounds
  - https://github.com/awslabs/mountpoint-s3-csi-driver/issues/141

Running Mountpoint inside a container would help to address these downsides.

everpeace:
Thanks for opening this issue.

As I described [here](https://github.com/awslabs/mountpoint-s3-csi-driver/issues/82#issuecomment-1830915362), "Running Mountpoint inside a container ___without any privileges___" would be better.

How about updating the issue title??

unexge:
Hey @everpeace, agree with you. We haven't finalized the design yet, but we aim to make Mountpoint containers unprivileged.

anurag4DSB:
Hello,

Could we expect any downtime or migration tasks if we are already using systemd mounter ?
If yes, how will migration work from systemd mounter to pod mounter ?

unexge:
Hey @anurag4DSB, we're still working on upgrade path for this feature and will be sharing more details once its finalized and ready to release.

dgard1981:
@unexge, are you able to share any kind of tentative timeline for this fix? It's been well over a year since a feature request was made in #169, and unfortunately the limitation means that it's impossible for many users to actually use the driver.

If it's still going to be a while, can a note be added to the README warning users of this limitation? It's not at all obvious from the logs available through Kubernetes what the issue was, and I feel a note would save people a lot of time.

unexge:
[...]

PR #291, opened 2024-11-11
Add `aws-s3-csi-mounter` component

unexge:
This is part of https://github.com/awslabs/mountpoint-s3-csi-driver/issues/279.

This new component, `aws-s3-csi-mounter`, will be the entry point for the container running Mountpoint. It will be responsible for receiving mount options and FUSE file descriptor and spawning Mountpoint process until completion.

---

### DIFF
--- a/cmd/aws-s3-csi-mounter/csimounter/csimounter.go
+++ b/cmd/aws-s3-csi-mounter/csimounter/csimounter.go
@@ -6,6 +6,8 @@ import (
 	"os/exec"
 	"slices"
 
+	"k8s.io/klog/v2"
+
 	"github.com/awslabs/aws-s3-csi-driver/pkg/podmounter/mountoptions"
 )
 
@@ -64,6 +66,8 @@ func Run(options Options) (int, error) {
 	// so Mountpoint logs can be viewable with `kubectl logs`.
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
+
+	klog.Info("Starting Mountpoint process")
 
 	return options.CmdRunner(cmd)
 }
--- a/cmd/aws-s3-csi-mounter/csimounter/csimounter.go
+++ b/cmd/aws-s3-csi-mounter/csimounter/csimounter.go
@@ -1,0 +1,69 @@
+package csimounter
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"slices"
+
+	"github.com/awslabs/aws-s3-csi-driver/pkg/podmounter/mountoptions"
+)
+
+// A CmdRunner is responsible for running given `cmd` until completion and returning its exit code and its error (if any).
+// This is mainly exposed for mocking in tests, `DefaultCmdRunner` is always used in non-test environments.
+type CmdRunner func(cmd *exec.Cmd) (int, error)
+
+// DefaultCmdRunner is a real CmdRunner implementation that runs given `cmd`.
+func DefaultCmdRunner(cmd *exec.Cmd) (int, error) {
+	err := cmd.Run()
+	if err != nil {
+		return 0, err
+	}
+	return cmd.ProcessState.ExitCode(), nil
+}
+
+// An Options represents options to use while mounting Mountpoint.
+type Options struct {
+	MountpointPath string
+	MountOptions   mountoptions.Options
+	CmdRunner      CmdRunner
+}
+
+// Run runs Mountpoint with given options until completion and returns its exit code and its error (if any).
+func Run(options Options) (int, error) {
+	if options.CmdRunner == nil {
+		options.CmdRunner = DefaultCmdRunner
+	}
+
+	mountOptions := options.MountOptions
+
+	fuseDev := os.NewFile(uintptr(mountOptions.Fd), "/dev/fuse")
+	if fuseDev == nil {
+		return 0, fmt.Errorf("passed file descriptor %d is invalid", mountOptions.Fd)
+	}
+
+	args := mountOptions.Args
+
+	// By default Mountpoint runs in a detached mode. Here we want to monitor it by relaying its output,
+	// and also we want to wait until it terminates. We're passing `--foreground` to achieve this.
+	const foreground = "--foreground"
+	if !slices.Contains(args, foreground) {
+		args = append(args, foreground)
+	}
+
+	args = append([]string{
+		mountOptions.BucketName,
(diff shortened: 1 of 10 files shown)


pkg/util/testutil/assert/assert.go
```diff
@@ -0,0 +1,25 @@
+// Package assert provides utilities for making assertions during tests.
```
muddyfish, line 1: Is there any widely accepted testing repository that has assert equal/not equals we can use instead?

cmd/aws-s3-csi-mounter/main.go
```diff
@@ -0,0 +1,53 @@
...
+	"time"
+
+	"k8s.io/klog/v2"
+
+	"github.com/awslabs/aws-s3-csi-driver/cmd/aws-s3-csi-mounter/csimounter"
+	"github.com/awslabs/aws-s3-csi-driver/pkg/podmounter/mountoptions"
+)
+
+var mountSockPath = flag.String("mount-sock-path", "/sock/mount.sock", "Path of the Unix socket to receive mount options from.")
+var mountSockRecvTimeout = flag.Duration("mount-sock-recv-timeout", 2*time.Minute, "Timeout for receiving mount options from passed Unix socket.")
```
muddyfish, line 23: 2 minutes is a long time. Why this number?

pkg/podmounter/mountoptions/mount_options.go
```diff
@@ -0,0 +1,130 @@
...
+		return Options{}, fmt.Errorf("expected to got one file descriptor from unix socket %s, but got %d", sockPath, len(fds))
+	}
+
+	options.Fd = fds[0]
+	return options, nil
+}
+
+// parseUnixRights parses given socket control message to extract passed file descriptors.
+func parseUnixRights(buf []byte) ([]int, error) {
+	socketControlMessages, err := syscall.ParseSocketControlMessage(buf)
```
muddyfish, line 115: I'm not sure what this is meant to do

pkg/podmounter/mountoptions/mount_options.go
```diff
@@ -0,0 +1,130 @@
...
+
+var (
+	messageRecvSize = 1024
+	// We only pass one file descriptor and its 32 bits
+	unixRightsRecvSize = syscall.CmsgSpace(4)
+)
+
+// Recv receives passed mount options via `Send` function through given `sockPath`.
+func Recv(ctx context.Context, sockPath string) (Options, error) {
+	l, err := net.Listen("unix", sockPath)
```
muddyfish, line 60: I've not looked through the protocol for this - is it documented anywhere?

pkg/util/socket.go
```diff
@@ -0,0 +1,32 @@
...
+
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+// ErrUnixSocketNotExists is returned if the socket won't appear within the timeout duration.
+var ErrUnixSocketNotExists = errors.New("util/socket: unix socket does not exists")
+
+// WaitForUnixSocket waits for the duration of `timeout` until `path` exists by checking it every `interval`.
+// It returns `ErrUnixSocketNotExists` if the socket won't appear within the timeout, otherwise it returns nil.
+func WaitForUnixSocket(timeout time.Duration, interval time.Duration, path string) error {
```
muddyfish, line 18: Nit: Would it be more efficient to use something like https://github.com/fsnotify/fsnotify to receive a notification from the OS that the file exists?

pkg/util/testutil/assert/assert.go
unexge: https://github.com/stretchr/testify is the most common one as far as I know. I think some part of Go community discourages [assert libraries](https://go.dev/wiki/TestComments#assert-libraries). I'm not against it – I just didn't wanted to pull a big package for simple assertions. We also depend on https://github.com/onsi/gomega in our integration tests as it's used in Kubernetes testing framework. 

I think I can create an issue to standardizing all assertions to one package in the codebase.

cmd/aws-s3-csi-mounter/main.go
unexge: Kinda arbitrary. This is the time this component will wait until the CSI Driver Node Pod sends mount options to it. I didn't wanted to make it short and cause retry loop. I don't expect it to wait until the whole timeout duration unless we have a bug in one of our components.

pkg/podmounter/mountoptions/mount_options.go
unexge: as the comment suggests "... parses given socket control message to extract passed file descriptors.". Maybe I can add references to "Ancillary messages" and "SCM_RIGHTS" from https://man7.org/linux/man-pages/man7/unix.7.html.

pkg/podmounter/mountoptions/mount_options.go
unexge: Do you mean the first parameter to `net.Listen`? If so, see https://pkg.go.dev/net#Listen

pkg/util/socket.go
unexge: I think it would, but since this is not in the data path, and we don't need low-latency, I think it would complicate this function as well. Do you think we need it?

cmd/aws-s3-csi-mounter/csimounter/csimounter.go
```diff
@@ -0,0 +1,73 @@
...
+	if fuseDev == nil {
+		return 0, fmt.Errorf("passed file descriptor %d is invalid", mountOptions.Fd)
+	}
+
+	args := mountOptions.Args
+
+	// By default Mountpoint runs in a detached mode. Here we want to monitor it by relaying its output,
+	// and also we want to wait until it terminates. We're passing `--foreground` to achieve this.
+	const foreground = "--foreground"
+	if !slices.Contains(args, foreground) {
```
muddyfish, line 52: We should probably also check for the presence of `-f` here as well

pkg/podmounter/mountoptions/mount_options.go
muddyfish: I was more meaning the protocol for send/receive, but I can now see it's just json serialising some structure, which seems fine

pkg/util/socket.go
muddyfish: That's fair, no need for it now and we can think about it later if we discover it's something we want

muddyfish (approved):
Looks good to me, only thoughts are that we might want to check for the `-f` flag as well as the `--foreground` one

PR #292, opened 2024-11-13
Add end-to-end tests for local cache

unexge:

### DIFF
--- a/tests/e2e-kubernetes/testsuites/cache.go
+++ b/tests/e2e-kubernetes/testsuites/cache.go
@@ -1,0 +1,188 @@
+package custom_testsuites
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"slices"
+
+	"github.com/google/uuid"
+	. "github.com/onsi/ginkgo/v2"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/kubernetes/test/e2e/framework"
+	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
+	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
+	e2evolume "k8s.io/kubernetes/test/e2e/framework/volume"
+	storageframework "k8s.io/kubernetes/test/e2e/storage/framework"
+	admissionapi "k8s.io/pod-security-admission/api"
+	"k8s.io/utils/ptr"
+)
+
+const volumeName1 = "volume1"
+const root = int64(0)
+const defaultNonRootGroup = int64(2000)
+
+type s3CSICacheTestSuite struct {
+	tsInfo storageframework.TestSuiteInfo
+}
+
+func InitS3CSICacheTestSuite() storageframework.TestSuite {
+	return &s3CSICacheTestSuite{
+		tsInfo: storageframework.TestSuiteInfo{
+			Name: "cache",
+			TestPatterns: []storageframework.TestPattern{
+				storageframework.DefaultFsPreprovisionedPV,
+			},
+		},
+	}
+}
+
+func (t *s3CSICacheTestSuite) GetTestSuiteInfo() storageframework.TestSuiteInfo {
+	return t.tsInfo
+}
+
+func (t *s3CSICacheTestSuite) SkipUnsupportedTests(_ storageframework.TestDriver, pattern storageframework.TestPattern) {
+	if pattern.VolType != storageframework.PreprovisionedPV {
+		e2eskipper.Skipf("Suite %q does not support %v", t.tsInfo.Name, pattern.VolType)
+	}
+}
+
+func (t *s3CSICacheTestSuite) DefineTests(driver storageframework.TestDriver, pattern storageframework.TestPattern) {
+	f := framework.NewFrameworkWithCustomTimeouts(NamespacePrefix+"cache", storageframework.GetDriverTimeouts(driver))
+	f.NamespacePodSecurityLevel = admissionapi.LevelPrivileged
+
+	type local struct {
+		config *storageframework.PerTestConfig
+
+		// A list of cleanup functions to be called after each test to clean resources created during the test.
+		cleanup []func(context.Context) error
+	}
+
+	var l local
+
+	deferCleanup := func(f func(context.Context) error) {
+		l.cleanup = append(l.cleanup, f)
+	}
+
+	cleanup := func(ctx context.Context) {
+		var errs []error
+		slices.Reverse(l.cleanup) // clean items in reverse order similar to how `defer` works
+		for _, f := range l.cleanup {
+			errs = append(errs, f(ctx))
+		}
+		framework.ExpectNoError(errors.NewAggregate(errs), "while cleanup resource")
+	}
+	BeforeEach(func(ctx context.Context) {
+		l = local{}
+		l.config = driver.PrepareTest(ctx, f)
+		DeferCleanup(cleanup)
+	})
+
+	createPod := func(ctx context.Context, additionalMountOptions []string, podModifiers ...func(*v1.Pod)) *v1.Pod {
+		cacheDir := randomCacheDir()
+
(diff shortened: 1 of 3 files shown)


lattaai9:
Hello, I tried to solve the issue.

This is what I did:

Updated the CONTRIBUTING.md file to include a note about confirming the licensing of contributions when submitting a pull request. This addresses the criteria mentioned in the issue.

You can review changes in this commit: https://github.com/lattaai9/awslabs-mountpoint-s3-csi-driver-292/commit/4b2149f234d9db99a26c0e87a638925ddd1cf499.


> [!CAUTION]
> Disclaimer: The commit was created by Latta AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---
This issue was tried to solve for free by Latta AI - https://latta.ai/ourmission

If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

tests/e2e-kubernetes/testsuites/util.go
```diff
@@ -76,6 +77,27 @@ func checkListingPath(f *framework.Framework, pod *v1.Pod, path string) {
...
+	first := filepath.Join(basePath, "first")
+	second := filepath.Join(dir, "second")
+
+	seed := time.Now().UTC().UnixNano()
+	testWriteSize := 1024 // 1KB
+
+	checkWriteToPath(f, pod, first, testWriteSize, seed)
+	checkReadFromPath(f, pod, first, testWriteSize, seed)
+	e2evolume.VerifyExecInPodSucceed(f, pod, fmt.Sprintf("mkdir %s && cd %s && touch %s", dir, dir, second))
+	checkListingPath(f, pod, dir)
```
muddyfish, line 94: Is it worth verifying what the actual response is here?

tests/e2e-kubernetes/testsuites/cache.go
```diff
@@ -0,0 +1,188 @@
...
+
+func (t *s3CSICacheTestSuite) SkipUnsupportedTests(_ storageframework.TestDriver, pattern storageframework.TestPattern) {
+	if pattern.VolType != storageframework.PreprovisionedPV {
+		e2eskipper.Skipf("Suite %q does not support %v", t.tsInfo.Name, pattern.VolType)
+	}
+}
+
+func (t *s3CSICacheTestSuite) DefineTests(driver storageframework.TestDriver, pattern storageframework.TestPattern) {
+	f := framework.NewFrameworkWithCustomTimeouts(NamespacePrefix+"cache", storageframework.GetDriverTimeouts(driver))
+	f.NamespacePodSecurityLevel = admissionapi.LevelPrivileged
```
muddyfish, line 53: Why do we need this? Just for the cache creation on underlying host thing?

tests/e2e-kubernetes/testsuites/cache.go
```diff
@@ -0,0 +1,188 @@
...
+		cleanup []func(context.Context) error
+	}
+
+	var l local
+
+	deferCleanup := func(f func(context.Context) error) {
+		l.cleanup = append(l.cleanup, f)
+	}
+
+	cleanup := func(ctx context.Context) {
```
muddyfish, line 68: Not blocking: Don't we have this setup in a load of other tests? At some point, maybe we should try and de-duplicate

tests/e2e-kubernetes/testsuites/cache.go
```diff
@@ -0,0 +1,188 @@
...
+						VolumeMounts: []v1.VolumeMount{
+							{
+								Name:      volumeName1,
+								MountPath: e2epod.VolumeMountPath1,
+							},
+						},
+					})
+				})
+
+				e2evolume.VerifyExecInPodSucceed(f, pod, fmt.Sprintf("cat %s | grep -q 'hello world!'", testFile))
```
muddyfish, line 146: I don't understand what feature this is trying to test

tests/e2e-kubernetes/testsuites/util.go
unexge: Addressed with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/292/commits/fb140b96afaa2b20a09785fd55e3ef1d51bea5e3

tests/e2e-kubernetes/testsuites/cache.go
unexge: Yes, namely it fails with:
```
pod Create API error: pods "pvc-tester-rcb9t" is forbidden: violates PodSecurity "baseline:latest": hostPath volumes (volume "make-cache-dir")
```

I realized we don't need the Pods to be privileged though (in L89) and changed it. Also, added a comment on this section to mention why this is required. https://github.com/awslabs/mountpoint-s3-csi-driver/pull/292/commits/991728e06c74d59e1d178a050909dd31618a4af2

tests/e2e-kubernetes/testsuites/cache.go
unexge: Yeah, that's true. It's a bit tricky because we create and use a local variable in the function scope and reuse it with each test run via `BeforeEach` lifecycle events, see `local` struct in each test suite. Might need to change that pattern a bit to have something reusable.

tests/e2e-kubernetes/testsuites/cache.go
unexge: This test-case verifies multiple containers can use the same cache/Mountpoint without any problem. The init container writes to `testFile` and this container verifies it has the expected content.

unexge:
> Are the thoughts here that MP should work as a filesystem, or that we want to test it actually uses a cache?

Former - I wanted to have a smoke test mainly, I expect Mountpoint to have more extensive tests to ensure cache works as expected. Here just wanted to ensure the filesystem still works when cache is enabled with multi-container, non-root etc setups.

tests/e2e-kubernetes/testsuites/cache.go
unexge: Good point! Should be pre-populated with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/292/commits/70310991b4f191f10eb0672ee2278d26a2d67811

tests/e2e-kubernetes/testsuites/util.go
```diff
@@ -76,6 +78,36 @@ func checkListingPath(f *framework.Framework, pod *v1.Pod, path string) {
...
+
+	dir := filepath.Join(basePath, "test-dir")
+	first := filepath.Join(basePath, "first")
+	second := filepath.Join(dir, "second")
+
+	seed := time.Now().UTC().UnixNano()
+	testWriteSize := 1024 // 1KB
+
+	checkWriteToPath(f, pod, first, testWriteSize, seed)
+	checkReadFromPath(f, pod, first, testWriteSize, seed)
```
muddyfish, line 103: Can we do multiple read checks for verifying it works a second time (cache works)?

unexge: Addressed with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/292/commits/7b15d44b50b9600ef4767fbea776378bdedff1bc

tests/e2e-kubernetes/testsuites/cache.go
```diff
@@ -0,0 +1,237 @@
...
+				})
+				checkBasicFileOperations(ctx, pod, bucketName, e2epod.VolumeMountPath1)
+			})
+
+			It("basic file operations as non-root", func(ctx context.Context) {
+				mountOptions := []string{
+					"allow-delete",
+					"allow-other",
+					fmt.Sprintf("uid=%d", *e2epod.GetDefaultNonRootUser()),
+					fmt.Sprintf("gid=%d", defaultNonRootGroup),
```
muddyfish, line 154: Nit: Would these constants be useful as functions/moved to `e2epod`?

muddyfish (approved):

tests/e2e-kubernetes/testsuites/cache.go
unexge: `e2epod` is a Kubernetes package, but I can move these constants to `util.go` in a follow-up PR.

PR #296, opened 2024-11-14
Add sample Dockerfile for building Mountpoint from source

unexge:
Sometimes it's useful to be able to build Mountpoint from source to test work-in-progress features. This PR adds a sample Dockerfile to achieve that. One can build the CSI Driver image by compiling Mountpoint from source by:
```
$ DOCKERFILE="Dockerfile.source" REGISTRY=ACCOUNT_ID.dkr.ecr.eu-north-1.amazonaws.com TAG=latest IMAGE_NAME=mountpoint-csi-driver-testing OUTPUT_TYPE=docker make image
```

---


muddyfish:
This doesn't build MP from source exactly - instead from a published commit. Could we rename this to something else? Say `Dockerfile.github`?

unexge:
For me, I'd still consider this as building from source as opposed to using prebuilt artifacts. Mountpoint has also a [Dockerfile.source](https://github.com/awslabs/mountpoint-s3/blob/main/docker/Dockerfile.source), and it also builds Mountpoint by cloning the repository.

muddyfish:
I'm fine approving this, but I'm unsure what we'd call a build from local sources. Perhaps `Dockerfile.local`

muddyfish (dismissed):

unexge:
That sounds good to me, changed to `Dockerfile.local`.

PR #306, opened 2024-12-02
[#276] Add amazonaws.com to auds

devkoriel:
*Issue #, if available:*

#276 

*Description of changes:*

Add `amazonaws.com` to tokenRequests audience list just like other tools.

### DIFF
--- a/charts/aws-mountpoint-s3-csi-driver/templates/csidriver.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/csidriver.yaml
@@ -12,4 +12,6 @@ spec:
   tokenRequests:
     - audience: "sts.amazonaws.com"
       expirationSeconds: 3600
+    - audience: "amazonaws.com"
+      expirationSeconds: 3600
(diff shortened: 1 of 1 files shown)


unexge:
Hey @devkoriel, thanks for the pull request. Could you please provide some context on your use-case to help us understand why this would be needed?

devkoriel:
I think other many users might be using `amazonaws.com` as the audience. This is why I made this PR.  BTW, if I use pod-level authentication, I don't need to create SA for csi driver right?

unexge:
Is there a product/tool/documentation that asks/suggests using `amazonaws.com`? Because by default `sts.amazonaws.com` is used in IRSA and that's the suggested audience in EKS documentations.

unexge:
I'll close this for now, please let us know if there is a use-case that needs this change. Also, one option could be using [Post Rendering](https://helm.sh/docs/topics/advanced/#post-rendering) hook to tweak this audience field if you're using Helm.

PR #308, opened 2024-12-03
[helm] Support change priorityClassName

DingGGu:
*Issue #, if available:*

*Description of changes:*

Support customize priorityClassName of node DaemonSet

### DIFF
--- a/charts/aws-mountpoint-s3-csi-driver/Chart.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: aws-mountpoint-s3-csi-driver
 description: A Helm chart for installing the Mountpoint for Amazon S3 CSI Driver. This CSI driver allows your Kubernetes applications to access S3 objects through a file system interface.
-version: 1.10.0
+version: 1.10.1
 kubeVersion: ">=1.23.0-0"
 home: https://github.com/awslabs/mountpoint-s3-csi-driver
 sources:
--- a/charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
@@ -23,7 +23,7 @@ spec:
         {{- toYaml . | nindent 8 }}
         {{- end }}
       serviceAccountName: {{ .Values.node.serviceAccount.name }}
-      priorityClassName: system-node-critical
+      priorityClassName: {{ .Values.node.priorityClassName }}
       {{- with .Values.node.affinity }}
       affinity: {{- toYaml . | nindent 8 }}
       {{- end }}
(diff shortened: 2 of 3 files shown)


unexge:
Hey @DingGGu, thanks for the pull request. Could you please share some context on your use-case to customize priority class? Changing it to something with a lower priority than `system-node-critical` might cause Kubernetes to evict CSI Driver Pods and S3 CSI Driver would stop functioning in that node.

DingGGu:
> Hey @DingGGu, thanks for the pull request. Could you please share some context on your use-case to customize priority class? Changing it to something with a lower priority than `system-node-critical` might cause Kubernetes to evict CSI Driver Pods and S3 CSI Driver would stop functioning in that node.

Hello.

If you try to deploying a new DaemonSet on an running node, there is a risk that critical Pods may be evicted by the DaemonSet.

Therefore, when deploying a new DaemonSet, we set priorityClassName to the `default` or lower, check if there are any affected existing workloads on each nodes, and then adjust PriorityClass back to system-node-critical.

If you don't understand, please let me know again.

unexge:
Hey @DingGGu thanks for explaining your use-case. I've a question about updating `priorityClassName`. I think updating `priorityClassName` of a DaemonSet PodSpec would cause Pods to be re-created, and then the new Pods would be scheduled with the updated `priorityClassName` which would cause the same behavior as you originally wanted to avoid.

I also checked if it is possible to update `priorityClassName` of a scheduled Pod rather than the DaemonSet, but seems like `priorityClassName` only used during admission time to populate `priority` field, so updating `priorityClassName` afterward might be no-op.

> The priority admission controller uses the priorityClassName field and populates the integer value of the priority. 

[Pod priority](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#pod-priority) and https://github.com/kubernetes/kubernetes/blob/1504f10e7946f95a8b1da35e28e4c7453ff62775/plugin/pkg/admission/priority/admission.go#L144-L202

unexge:
Hey @DingGGu, thanks for the details. My main concern is adding something that can be easily misused, though your use-case seems like a valid case. If you're using Helm, would using [Post Rendering](https://helm.sh/docs/topics/advanced/#post-rendering) hook to modify `priorityClassName` work for you? 

This [linked example](https://github.com/thomastaylor312/advanced-helm-demos/tree/master/post-render) from Helm docs uses Kustomize to patch Helm manifest before applying the changes and might solve your use-case.

DingGGu:
@unexge 
Using postRender seems great (thanks for letting me know that exists).
However, we often want to find a simple solution.
I think adding priorityClassName doesn't add much to the complexity or unusability of the code.

`ebs-csi-driver` already provides change priorityClassName.
You can see: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/templates/_node.tpl#L48

Helm is just a templating tool, and I think it should provide these options "easy to use", since developers can't take into consideration all the user's requirements.

I'm closing this case, as it seems like there's no further progress.

unexge:
Please let us know if the Helm post rendering work-around wouldn't work for you @DingGGu. Sorry for the friction.

PR #317, opened 2024-12-10
Update E2E Tests to Support Explicit STS Region Configuration 

jiaeenie:
*Issue:*

This PR addresses Credentials test failures in clusters without IMDS access. 

*Description of Changes:*

- Add a new flag `imds-available` to indicate if IMDS is available.
- Configure the STS region through the `bucket-region` flag and pass the region explicitly to the tests.
- Add a test case to verify automatic STS region detection when the `imds-available` is set to true.

### DIFF
--- a/tests/e2e-kubernetes/testsuites/credentials.go
+++ b/tests/e2e-kubernetes/testsuites/credentials.go
@@ -53,7 +53,7 @@ const serviceAccountTokenAudienceSTS = "sts.amazonaws.com"
 const roleARNAnnotation = "eks.amazonaws.com/role-arn"
 const credentialSecretName = "aws-secret"
 
-// DefaultRegion specifies the STS region explicitl.
+// DefaultRegion specifies the STS region explicitly.
 var DefaultRegion string
 
 // IMDSAvailable indicates whether the Instance Metadata Service is accessible.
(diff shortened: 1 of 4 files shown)


unexge (commented):
Thanks for the PR! Could you please pass `--imds-available` flag in our CI as we know it's available in our test clusters?

tests/e2e-kubernetes/testsuites/credentials.go
```diff
@@ -465,7 +472,8 @@ func (t *s3CSICredentialsTestSuite) DefineTests(driver storageframework.TestDriv
 				})
 
 				It("should use up to date role associated with pod's service account", func(ctx context.Context) {
-					vol := createVolumeResourceWithMountOptions(enablePodLevelIdentity(ctx), l.config, pattern, []string{"allow-delete"})
+					var mountOptions = append([]string{"allow-delete"}, fmt.Sprintf("region %s", DefaultRegion))
```
unexge, line 475: nit: You can write `mountOptions := []string{"allow-delete", fmt.Sprintf("region %s", DefaultRegion)}` instead of using `append` here

unexge (approved):

Issue #284, opened 2024-11-04
Driver crashes unexpectedly with `Failed to read /host/proc/mounts` requiring pod restart


dienhartd:
/kind bug
> **_NOTE:_**  If this is a filesystem related bug, please take a look at the [Mountpoint repo](https://github.com/awslabs/mountpoint-s3) to submit a [bug report](https://github.com/awslabs/mountpoint-s3/issues/new?assignees=&labels=bug&projects=&template=bug-report.yml)

**What happened?**
Periodically without warning one of my s3 mountpoint driver pods will crash with GRPC errors until I delete it. It will usually cause a dependent pod to fail to start. The replacement immediately after this pod's deletion works fine, but requires manual intervention after noticing dependent pod crashes due to missing pv.

**What you expected to happen?**
Error not to occur.

**How to reproduce it (as minimally and precisely as possible)?**
Unclear. 

**Anything else we need to know?**:
Logs
```
I1104 11:59:40.249998       1 credential.go:95] NodePublishVolume: Using driver identity
I1104 11:59:40.250015       1 node.go:146] NodePublishVolume: mounting d-cluster at /var/lib/kubelet/pods/97e71fea-b356-4d87-a086-5f06fe651ea7/volumes/kubernetes.io~csi/s3-pv/mount with options [--allow-delete --allow-other --gid=100 --uid=1000]
E1104 11:59:40.250106       1 mount.go:214] Failed to read /host/proc/mounts on try 1: open /host/proc/mounts: invalid argument
E1104 11:59:40.250106       1 mount.go:214] Failed to read /host/proc/mounts on try 1: open /host/proc/mounts: invalid argument
E1104 11:59:40.250106       1 mount.go:214] Failed to read /host/proc/mounts on try 1: open /host/proc/mounts: invalid argument
E1104 11:59:40.250106       1 mount.go:214] Failed to read /host/proc/mounts on try 1: open /host/proc/mounts: invalid argument
E1104 11:59:40.350345       1 mount.go:214] Failed to read /host/proc/mounts on try 2: open /host/proc/mounts: invalid argument
E1104 11:59:40.350345       1 mount.go:214] Failed to read /host/proc/mounts on try 2: open /host/proc/mounts: invalid argument
E1104 11:59:40.350345       1 mount.go:214] Failed to read /host/proc/mounts on
try 2: open /host/proc/mounts: invalid argument
[...]

PR #321, opened 2024-12-16
Use `HostToContainer` mount propagation instead of  `/host/proc/mounts`

jiaeenie:
*Description of changes:*
Currently, we spawn Mountpoint processes on the host using systemd. As a result, the mounts created by Mountpoint are not visible inside the CSI Driver Pod. To work around this, we were mounting `/proc/mounts` from host and parsing this file to check existing mounts on the host. Mounting `/proc/mounts` causes problems with Karpenter sometimes and its also its blocked by some SELinux policies. Such as this [issue](https://github.com/awslabs/mountpoint-s3-csi-driver/issues/284).

This commit instead uses `HostToContainer` mount propagation on `hostPath` mount for `/var/lib/kubelet`. Thanks to `HostToContainer`, any new mounts created inside `/var/lib/kubelet` gets automatically propagated to our Pod from the host.


pkg/driver/node/mounter/systemd_mounter_test.go
```diff
@@ -387,7 +409,7 @@ sysfs /sys sysfs rw,seclabel,nosuid,nodev,noexec,relatime 0 0`),
 			err = os.WriteFile(procMountsPath, test.procMountsContent, 0755)
 			assertNoError(t, err)
 
-			mounter := &mounter.SystemdMounter{MountLister: &mounter.ProcMountLister{ProcMountPath: procMountsPath}}
+			mounter := &mounter.SystemdMounter{Mounter: mount.NewFakeMounter(parseProcMounts(test.procMountsContent))}
```
unexge, line 412: I think we can change `test.procMountsContent` to be a `[]mount.MountPoint`, so we don't have to parse and convert them from a string. That why we wouldn't need `parseProcMounts` function.

jiaeenie: Thanks for reviewing, I was thinking about it at the first place, but I realised that for each test, the` procMountsContent` might be different, e.g: "non existing mount on /proc/mounts", so it might be cleaner to keep the parsing function in this case?

jiaeenie: Reviews have been addressed in the following commits.

pkg/driver/node/mounter/systemd_mounter.go
```diff
@@ -49,7 +49,7 @@ func (m *SystemdMounter) IsMountPoint(target string) (bool, error) {
 		return false, err
 	}
 
-	mountPoints, err := m.MountLister.ListMounts()
+	mountPoints, err := m.Mounter.List()
```
muddyfish, line 52: How does this work? I did a quick check here https://github.com/kubernetes/mount-utils/blob/887f9a11be38008c0ba049117324b7f645ab5ea2/mount_linux.go#L426-L427 and it looks like it's just parsing `/proc/mounts`?

unexge:
The change looks good to me. One additional thing we should do is to ensure `targetPath` passed in `NodePublishVolume` is inside kubelet path. Because we rely on propagation of mounts under kubelet path, any other mount wouldn't be propagated to our Pod. This shouldn't be the case, but it would be good to emit an error just in case. Something like that in `S3NodeServer.NodePublishVolume`:

```go

const defaultKubeletPath = "/var/lib/kubelet"

var kubeletPath = getKubeletPath()

func getKubeletPath() string {
	path := os.Getenv("KUBELET_PATH")
	if path == "" {
		return defaultKubeletPath
	}
	return path
}

// ...

func (ns *S3NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolumeRequest) (*csi.NodePublishVolumeResponse, error) {
        // ...

	if !filepath.HasPrefix(target, kubeletPath) {
		klog.Errorf("NodePublishVolume: target path %q is not in kubelet path %q. This might cause mounting issues, please ensure you have correct kubelet path configured.", req.GetTargetPath(), kubeletPath)
	}
}
```

pkg/driver/node/mounter/systemd_mounter.go
jiaeenie: Thanks for reviewing. So our current code was basically copy-paste of `k8s.io/mount-utils` because we were mounting `/proc/mounts` for our container from host at `/host/proc/mounts` and that library has [hardcoded path ](https://github.com/kubernetes/mount-utils/blob/887f9a11be38008c0ba049117324b7f645ab5ea2/mount_linux.go#L46) for `/proc/mounts`. With this change, we no longer mount `/host/proc/mounts` but just enable propagation from host to our `/proc/mounts`, so we can just use the library as is without any modifications. Which is achieved by [Mount Propagation](https://kubernetes.io/docs/concepts/storage/volumes/#mount-propagation) --updates our containers' `/proc/mounts` from host whenever there is a new mount is the on `/var/lib/kubelet`.

muddyfish, line 52: Can we add a comment here then that we're using `Mounter.List` instead of `Mounter.IsMountPoint` because we also want to verify that the mount point is an `mount-s3` mount point as well as being a mount point

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# Unreleased
+[Documentation](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/README.md)
+
```
unexge, line 3: Could you also add this change to here?

unexge:
The CI failure is unrelated, "Delete cluster" step [doesn't handle "NotFound" errors](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/tests/e2e-kubernetes/scripts/eksctl.sh#L105-L134) currently and tries to remove some resources for maximum number of retries configured with takes around ~20min and that causes temporary credentials to expire and operations to fail.

I've restarted the failed tasks and will submit a separate PR to fix the CI.

charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
unexge, line 1: We need to add `KUBELET_PATH` env from Helm values to environment of `s3-plugin` container, so if one customizes path of kubelet via `node.kubeletPath` Helm value, we can get the correct kubelet path in our container.

```yaml
      containers:
        - name: s3-plugin
          # ...
          env:
            # ...
            - name: KUBELET_PATH
              value: {{ .Values.node.kubeletPath }}
```

unexge (approved):
LGTM. Thanks @jiaeenie!

muddyfish (approved):

UnyieldingOrca:
Can you make a patch release for this fix? Or is there a way to deploy a prerelease version of the csi driver myself?

unexge:
Hey @UnyieldingOrca, we plan to do a release soon, but we don't have any date we can share at the moment. We'll be sharing updates in this ticket: https://github.com/awslabs/mountpoint-s3-csi-driver/issues/284. 

Regarding deploying a pre-release version, you'd first need to build our Docker image by using our [Makefile](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/5689655303715385df504fb8a1c30c32ab14d483/Makefile), and push it to your registry, and then install our Helm chart from source by specifying your registry.
```bash 
helm upgrade --install aws-mountpoint-s3-csi-driver \
    --namespace kube-system \
    ... \
    --set image.repository="1234567890.dkr.ecr.eu-north-1.amazonaws.com/your-registry" \
    --set image.tag=your-tag \
   ./charts/aws-mountpoint-s3-csi-driver
```

 Unfortunately, it's not trivial at the moment.

UnyieldingOrca:
Thanks for the guidance. I ended up following this and it fixed my issue.

PR #327, opened 2024-12-27
Add non-root end-to-end test cases to credentials test suite

unexge:

### DIFF
--- a/tests/e2e-kubernetes/testsuites/cache.go
+++ b/tests/e2e-kubernetes/testsuites/cache.go
@@ -26,7 +26,6 @@ import (
 
 const volumeName1 = "volume1"
 const root = int64(0)
-const defaultNonRootGroup = int64(2000)
 
 type s3CSICacheTestSuite struct {
 	tsInfo storageframework.TestSuiteInfo
@@ -183,13 +182,9 @@ func (t *s3CSICacheTestSuite) DefineTests(driver storageframework.TestDriver, pa
 			mountOptions := append(baseMountOptions,
 				"allow-delete",
 				"allow-other",
-				fmt.Sprintf("uid=%d", *e2epod.GetDefaultNonRootUser()),
+				fmt.Sprintf("uid=%d", defaultNonRootUser),
 				fmt.Sprintf("gid=%d", defaultNonRootGroup))
-			podModifiers := append(basePodModifiers, func(pod *v1.Pod) {
-				pod.Spec.Containers[0].SecurityContext.RunAsUser = e2epod.GetDefaultNonRootUser()
-				pod.Spec.Containers[0].SecurityContext.RunAsGroup = ptr.To(defaultNonRootGroup)
-				pod.Spec.Containers[0].SecurityContext.RunAsNonRoot = ptr.To(true)
-			})
+			podModifiers := append(basePodModifiers, podModifierNonRoot)
 
 			pod, bucketName := createPod(ctx, mountOptions, podModifiers...)
 			checkBasicFileOperations(ctx, pod, bucketName, e2epod.VolumeMountPath1)
--- a/tests/e2e-kubernetes/testsuites/cache.go
+++ b/tests/e2e-kubernetes/testsuites/cache.go
@@ -184,7 +184,7 @@ func (t *s3CSICacheTestSuite) DefineTests(driver storageframework.TestDriver, pa
 				"allow-other",
 				fmt.Sprintf("uid=%d", defaultNonRootUser),
 				fmt.Sprintf("gid=%d", defaultNonRootGroup))
-			podModifiers := append(basePodModifiers, podModifierNonRoot)
+			podModifiers := append([]func(*v1.Pod){podModifierNonRoot}, basePodModifiers...)
 
 			pod, bucketName := createPod(ctx, mountOptions, podModifiers...)
 			checkBasicFileOperations(ctx, pod, bucketName, e2epod.VolumeMountPath1)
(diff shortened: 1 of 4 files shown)


jiaeenie (commented):
Shall we change this [line](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/ee7f1ed293d2db90cc489666c4909759192bc68f/tests/e2e-kubernetes/testsuites/mountoptions.go#L88C3-L88C146) too? Such as replacing "uid=1000", "gid=2000" with
```
fmt.Sprintf("uid=%d", defaultNonRootUser),
fmt.Sprintf("gid=%d", defaultNonRootGroup))
```
And this [line](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/ee7f1ed293d2db90cc489666c4909759192bc68f/tests/e2e-kubernetes/testsuites/mountoptions.go#L127C48-L127C59) and this [line](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/ee7f1ed293d2db90cc489666c4909759192bc68f/tests/e2e-kubernetes/testsuites/mountoptions.go#L92) to use the variable `defaultNonRootGroup`.

unexge:
Good suggestion @jiaeenie! Addressed with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/327/commits/1c0014d4a9015dc5fd13a2d8f4d45821af6c334f

tests/e2e-kubernetes/testsuites/credentials.go
```diff
@@ -424,32 +457,49 @@ func (t *s3CSICredentialsTestSuite) DefineTests(driver storageframework.TestDriv
...
+					}
+
+					pod := createPod(ctx, vol, podModifiers...)
 
 					return pod, sa
 				}
 
 				It("should use pod's service account's read-only role", func(ctx context.Context) {
-					pod, _ := createPodWithServiceAccountAndPolicy(ctx, iamPolicyS3ReadOnlyAccess, false)
+					pod, _ := createPodWithServiceAccountAndPolicy(ctx, iamPolicyS3ReadOnlyAccess, false, false)
```
muddyfish, line 492: Is there a way to pass in keyword arguments here? Right now I can't understand what these boolean arguments are for.

tests/e2e-kubernetes/testsuites/cache.go
```diff
@@ -183,13 +182,9 @@ func (t *s3CSICacheTestSuite) DefineTests(driver storageframework.TestDriver, pa
 			mountOptions := append(baseMountOptions,
 				"allow-delete",
 				"allow-other",
-				fmt.Sprintf("uid=%d", *e2epod.GetDefaultNonRootUser()),
+				fmt.Sprintf("uid=%d", defaultNonRootUser),
```
muddyfish, line 185: Where does this variable come from? I think it would be better to do explicit imports rather than clobbering the variable namespace

tests/e2e-kubernetes/testsuites/cache.go
```diff
@@ -183,13 +182,9 @@ func (t *s3CSICacheTestSuite) DefineTests(driver storageframework.TestDriver, pa
...
 				fmt.Sprintf("gid=%d", defaultNonRootGroup))
-			podModifiers := append(basePodModifiers, func(pod *v1.Pod) {
-				pod.Spec.Containers[0].SecurityContext.RunAsUser = e2epod.GetDefaultNonRootUser()
-				pod.Spec.Containers[0].SecurityContext.RunAsGroup = ptr.To(defaultNonRootGroup)
-				pod.Spec.Containers[0].SecurityContext.RunAsNonRoot = ptr.To(true)
-			})
+			podModifiers := append([]func(*v1.Pod){podModifierNonRoot}, basePodModifiers...)
```
muddyfish, line 187: Same here, could we add the file's namespace here so it's easier to search? Also, why does it need the `[]func(*v1.Pod){...}` bit?

tests/e2e-kubernetes/testsuites/cache.go
```diff
@@ -332,6 +327,13 @@ func ensureCacheDirExistsInNode(pod *v1.Pod, cacheDir string) {
 		MountPath: "/cache",
 	}
 
+	if pod.Spec.SecurityContext == nil {
+		pod.Spec.SecurityContext = &v1.PodSecurityContext{}
+	}
+	// We need to set this false at Pod-level as `chmod-cache-dir` needs to run as `root` and this
+	// would prevent container creation if its true.
+	pod.Spec.SecurityContext.RunAsNonRoot = ptr.To(false)
```
muddyfish, line 335: Why are we always changing this even if the test explicitly set the security context?

tests/e2e-kubernetes/testsuites/credentials.go
unexge: Go doesn't support named arguments, usually structs are used if such a thing is desired, but I think it'd make this more verbose. Since this is a local helper function and not used outside this test-case, I think I'd prefer current one.

tests/e2e-kubernetes/testsuites/cache.go
unexge: It's coming from `util.go` in the same directory/package. Go declarations are package-scoped, if an identifier used without any package prefix it means it's coming from the same package (or imported via dot import, but that's discouraged [except in test packages](https://go.dev/wiki/CodeReviewComments#import-dot))

tests/e2e-kubernetes/testsuites/cache.go
unexge: For local/unexported identifiers (i.e., the identifiers starts with a lowercase, see https://go.dev/tour/basics/3) that means they're coming from the same package and there is no importing mechanism for them.

This `append` block basically infers the type of slice, and `[]func(*v1.Pod){...}` is to make sure it infers the correct type.

tests/e2e-kubernetes/testsuites/cache.go
unexge: If you instantiate Kubernetes testing framework with `LevelRestricted`, [it sets this to `true`](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/framework/pod/utils.go#L195-L230), meaning all containers in the Pod must run as non-root (i.e., not `0`). 

Since this is a generic helper function and gets called from different security levels, we're just making sure to set it to `false`, so we can spawn our privileged container needed to create cache directory. The other containers created for this Pod would probably have their `RunAsUser` set to non-root (due to security level), so we just need to remove this restriction at Pod-level for our container. 

This whole `ensureCacheDirExistsInNode` is very hacky due to fact Mountpoint running in the host and hopefully this will go away with running Mountpoint inside container.

muddyfish (approved):

PR #328, opened 2024-12-27
Split the functionality in `node/mounter` into smaller packages

unexge:
Currently, `node/mounter` package contains lots of logic on top of mounting, and it makes harder to change/add new mounting logic for https://github.com/awslabs/mountpoint-s3-csi-driver/issues/279. This PR tries to address that by splitting the functionality in `node/mounter` into smaller packages. 

Changes:
- [Create `mountpoint.Args` for parsing and accessing Mountpoint args](https://github.com/awslabs/mountpoint-s3-csi-driver/commit/73fbc70964f41a131e8a34915f56240f2fce821b) - Previously, accessing and manipulating of Mountpoint arguments was spread over to the codebase. This new `mountpoint` package contains all the logic related to Mountpoint arguments.
- [Add `envprovider` package for accessing environment variables](https://github.com/awslabs/mountpoint-s3-csi-driver/commit/96752b481d5d065f591aae9bd966b00a97343873) - This package contains all the logic related to accessing environment variables to pass Mountpoint.
- [Move STS region detection into `regionprovider` package](https://github.com/awslabs/mountpoint-s3-csi-driver/commit/7518590c9244a3f927fea6146904f46a2e0f2511) - This package contains all the logic related to detecting regions, which is currently just for STS to use in Pod-level identity.
- [Create `credentialprovider` package for getting AWS credentials](https://github.com/awslabs/mountpoint-s3-csi-driver/commit/bcc72d49d6de752b36649be3ac9334db5be7bf12) - This package contains all the logic related to providing AWS credentials for Mountpoint. One other change with this refactor is that, now this package does not assume any path to write credentials to, and instead accepts `writePath` and `envPath` in `Credentials.Dump` method. This is needed as the path to write credentials will be different with `SystemdMounter` and upcoming `PodMounter` for containerization. Another change is that, previously we were writing service account tokens for IRSA to `/var/lib/kubelet/plugins/s3.csi.aws.com/`, now we're writing them to `/var/lib/kubelet/pods/<pod-uuid>/volumes/kubernetes.io~csi/<volume-id>/credentials` – which is unique for each mount and makes cleaning credentials easier. This new logic will be also required in `PodMounter` as Mountpoint Pods won't have `hostPath` mount and instead each will have its [`emptyDir` to write credentials to](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/pkg/podmounter/mppod/creator.go#L101-L107).
- [Use newly created packages in `SystemdMounter`](https://github.com/awslabs/mountpoint-s3-csi-driver/commit/4c7f236427571c23a46ac50b49149e714c9cdcc7) - This change basically applies all previous changes to `SystemdMounter`.
- [Ensure to link documentations if some configuration is missing](https://github.com/awslabs/mountpoint-s3-csi-driver/pull/328/commits/545e24dd1b686fe678828bf74a15d1a9ed3478c1)
- [Ignore nil errors in SystemdMounter.ensureCredentialsDirExists](https://github.com/awslabs/mountpoint-s3-csi-driver/pull/328/commits/b7e02ac641b7f2349df738177a35d56bb8aebe94)

---


pkg/driver/driver.go
```diff
@@ -91,14 +91,6 @@ func NewDriver(endpoint string, mpVersion string, nodeID string) (*Driver, error
 }
 
 func (d *Driver) Run() error {
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
-	tokenFile := os.Getenv(webIdentityTokenEnv)
-	if tokenFile != "" {
-		klog.Infof("Found AWS_WEB_IDENTITY_TOKEN_FILE, syncing token")
-		go tokenFileTender(ctx, tokenFile, "/csi/token")
-	}
```
unexge, line 100: This logic has been moved to [credentials_sts_web_identity.go](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/4c7f236427571c23a46ac50b49149e714c9cdcc7/pkg/driver/node/credentialprovider/credentials_sts_web_identity.go). Since we enabled [`requiresRepublish`](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/charts/aws-mountpoint-s3-csi-driver/templates/csidriver.yaml#L15C3-L15C26) as part of Pod-level identity support, Kubernetes will call `NodePublishVolume` periodically to update existing service account tokens before they expire, and the credential provider will be called as part of this method. Another reason for this change is that, this assumes a single location for service account tokens for Driver-level identity, but with containerization this won't be the case. See the note regarding credential provider in the original PR description for more context.

pkg/driver/node/credentialprovider/awsprofile/aws_profile.go
```diff
@@ -75,14 +76,8 @@ func CleanupAWSProfile(basepath string) error {
...
 }
 
-func writeAWSProfileFile(path string, content string) error {
-	err := os.WriteFile(path, []byte(content), awsProfileFilePerm)
-	if err != nil {
-		return err
-	}
-	// If the given file exists, `os.WriteFile` just truncates it without changing it's permissions,
-	// so we need to ensure it always has the correct permissions.
-	return os.Chmod(path, awsProfileFilePerm)
```
unexge, line 85: [`renameio.WriteFile` does this](https://github.com/google/renameio/blob/v1.0.1/writefile.go#L30-L33) for us.

pkg/driver/node/mounter/systemd_mounter.go
```diff
@@ -168,48 +164,10 @@ func (m *SystemdMounter) Mount(bucketName string, target string, credentials *Mo
...
-func addUserAgentToOptions(options []string, userAgent string) []string {
-	// first remove it from the options in case it's in there
-	for i := len(options) - 1; i >= 0; i-- {
-		if strings.Contains(options[i], userAgentPrefix) {
-			options = append(options[:i], options[i+1:]...)
-		}
-	}
-	// add the hard coded S3 CSI driver user agent string
-	return append(options, userAgentPrefix+"="+userAgent)
-}
```
unexge, line 201: The logic for moving `--aws-max-attempts` to env and adding user-agent to arguments has been moved to `node.go` as it's unrelated to mount mechanism as needs to be there for both `SystemdMounter` and `PodMounter`.

pkg/driver/node/node.go
```diff
@@ -56,28 +59,18 @@ var (
...
-	if volumeContext[mounter.VolumeCtxAuthenticationSource] == mounter.AuthenticationSourcePod {
-		podID := volumeContext[mounter.VolumeCtxPodUID]
-		volumeID := req.GetVolumeId()
-		if podID != "" && volumeID != "" {
-			err := ns.credentialProvider.CleanupToken(volumeID, podID)
-			if err != nil {
-				klog.V(4).Infof("NodeStageVolume: Failed to cleanup token for pod/volume %s/%s: %v", podID, volumeID, err)
-			}
-		}
-	}
```
unexge, line 79: This `NodeStageVolume` wasn't enabled at all – we need to add `STAGE_UNSTAGE_VOLUME` node capability to enable – and also this cleaning of tokens has been moved to `Unmount` method of mounter.

pkg/driver/node/node.go
```diff
@@ -217,24 +182,12 @@ func (ns *S3NodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUn
...
-			klog.V(4).Infof("NodeUnpublishVolume: Volume ID from parsed target path differs from Volume ID passed: %s (parsed) != %s (passed)", targetPath.VolumeID, volumeID)
-		} else {
-			err := ns.credentialProvider.CleanupToken(targetPath.VolumeID, targetPath.PodID)
-			if err != nil {
-				klog.V(4).Infof("NodeUnpublishVolume: Failed to cleanup token for pod/volume %s/%s: %v", targetPath.PodID, volumeID, err)
-			}
-		}
-	} else {
-		klog.V(4).Infof("NodeUnpublishVolume: Failed to parse target path %s: %v", target, err)
-	}
```
unexge, line 237: This clean up logic has been moved to `Unmount` method of mounter.

pkg/driver/node/credentialprovider/awsprofile/awsprofiletest/aws_profile.go
jiaeenie, line 1: Shall we consider renaming this folder and package to something more indicative, like `awsprofiletestutils` (or simply `utils`), and use `awsprofile_test_utils` as the package name? This change may help differentiate it from the existing `awsprofile_test` package and "aws_profile_test.go" test file.

unexge: By convention, usually test util packages named like `<pkg>test`. For example:
- [httptest](https://pkg.go.dev/net/http/httptest) for `net/http`
- [iotest](https://pkg.go.dev/testing/iotest) for `io`
- [fstest](https://pkg.go.dev/testing/fstest) for `fs`

muddyfish:
Is it possible to split this refactor our into many smaller PRs? It's quite hard to understand the context here with so many changes

unexge:
Since each commit builds on top of each other, it would be hard to create lots of PRs and that's why I split them into commits and added description for each. Would it be possible to review it commit-by-commit?

pkg/mountpoint/args.go
```diff
@@ -0,0 +1,119 @@
...
+	ArgAllowRoot       = "--allow-root"
+	ArgRegion          = "--region"
+	ArgCache           = "--cache"
+	ArgUserAgentPrefix = "--user-agent-prefix"
+	ArgAWSMaxAttempts  = "--aws-max-attempts"
+)
+
+// An Args represents arguments to be passed to Mountpoint during mount.
+type Args struct {
+	args sets.Set[string]
```
muddyfish, line 22: I feel a set is the wrong data structure here. Why aren't we using a dictionary/hashmap? Then we don't have to iterate over the entire data structure to see presence, and we don't have to do as much messing around with formatting

pkg/driver/node/envprovider/provider.go
```diff
@@ -0,0 +1,74 @@
...
+	EnvSharedCredentialsFile = "AWS_SHARED_CREDENTIALS_FILE"
+	EnvRoleARN               = "AWS_ROLE_ARN"
+	EnvWebIdentityTokenFile  = "AWS_WEB_IDENTITY_TOKEN_FILE"
+	EnvEC2MetadataDisabled   = "AWS_EC2_METADATA_DISABLED"
+
+	EnvMountpointCacheKey = "UNSTABLE_MOUNTPOINT_CACHE_KEY"
+)
+
+// An Environment represents a list of environment variables.
+type Environment = []string
```
muddyfish, line 27: This should probably be a dict or list of tuples internally

pkg/driver/node/regionprovider/imds.go
```diff
@@ -0,0 +1,34 @@
+package regionprovider
```
muddyfish, line 1: Has any code here changed? The removed code isn't in this commit

pkg/driver/node/regionprovider/provider.go
```diff
@@ -0,0 +1,69 @@
...
+// SecurityTokenService tries to detect AWS region to use for STS.
+//
+// It looks for the following (in-order):
+//  1. `stsRegion` passed via volume context
+//  2. Region set for S3 bucket via mount options
+//  3. `AWS_REGION` or `AWS_DEFAULT_REGION` env variables
+//  4. Calling IMDS to detect region
+//
+// It returns [ErrUnknownRegion] if all of them fails.
+func (p *Provider) SecurityTokenService(volumeContext map[string]string, args mountpoint.Args) (string, error) {
```
muddyfish, line 43: The name here doesn't describe the functionality. From what I understand, it either returns the region to use *for* STS or returns an error. Perhaps `getSTSRegion`?

pkg/mountpoint/args.go
unexge: Not sure if Mountpoint currently uses such an argument but - dict/map is not a correct data structure to represent command line arguments IMO as you could have multiple values for same option, and also order of arguments might matter.

Like this `find` usage:
```
$ find . -name  "*file*" -type f -o \
         -name "*folder*" -type d
```

But that's being said the current implementation does not preserve original ordering and not sure if we need this behavior.

I didn't want to change original parsing function as that might cause some breaking changes if we do unintentional changes, but I think I added enough test cases now, and we should have enough confidence to make more refactor.

I think ideal data structure for this is: `List[Tuple[Key, Optional[Value]]]`

pkg/driver/node/regionprovider/imds.go
unexge: There is no code change (other than function name) in this package (`regionprovider/{imds, provider, provider_test}.go`). It just moved out of now removed `pkg/driver/node/mounter/credential_provider.go`

pkg/driver/node/envprovider/provider.go
unexge: Yes I think this should be `map[string]string`. Making it a map would cause lots of diffs in the tests though, they were already `[]string` and due to way type aliases work in Go, we can just use `[]string` in place of `Environment`. But if we make it a map, then we would need to update all call sites to use a proper type, which then would make this PR even bigger. I think it's best to do that as a follow-up PR.

pkg/mountpoint/args.go
muddyfish: Either an ordered map or what you proposed sounds ideal to me as I don't think MP currently supports multiple arguments with the same name.

pkg/driver/node/regionprovider/provider.go
muddyfish: I think `get` adds value as otherwise it's not obvious what we're actually doing with STS. I'd be happy with `regionProvider.getSTS(...)`

pkg/mountpoint/args.go
unexge: Address this in a separate PR: https://github.com/awslabs/mountpoint-s3-csi-driver/pull/349

pkg/driver/node/envprovider/provider.go
unexge: Addressed in a separate PR: https://github.com/awslabs/mountpoint-s3-csi-driver/pull/362

PR #329, opened 2025-01-02
Use kubectl matching with cluster's version

unexge:
Currently, we use stable version of kubectl (v1.32 as of today), and that might cause some problems as kubectl [emits a warning](https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/12561703668/job/35021134764#step:11:4567): 
> WARNING: version difference between client (1.32) and server (1.29) exceeds the supported minor version skew of +/-1

---

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -134,6 +134,7 @@ jobs:
       - name: Install tools
         run: |
           export ACTION=install_tools
+          export K8S_VERSION=${{ matrix.kubernetes-version }}
           tests/e2e-kubernetes/scripts/run.sh
       - name: Run Controller Tests
         run: |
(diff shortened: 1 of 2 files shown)


unexge:
Will move assume role duration change to a different PR as it requires some additional changes.

> [Increase assume role duration to 75 minutes in end-to-end tests](https://github.com/awslabs/mountpoint-s3-csi-driver/commit/b5180f9c01bfe92bf5029d7790290004493ee296) - Our assume role duration in our CI is 1 hour by default, but with the recent changes, our end-to-end test job started taking more than 1 hour sometimes, and that causes our [CI to fail due to expired credentials](https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/12519460437/job/34924009290#step:18:180).

jiaeenie (approved):
Thanks for the fix @unexge, looks good to me!

PR #340, opened 2025-01-14
Add `volumecontext` package for accessing volume context from CSI

unexge:

This might not make sense since it's moved out of its original context, but the original motivation was that these constants was spread out to different packages and that was causing circular dependencies.

For example, if you need to use `VolumeCtxBucketName` in `pkg/driver/node/mounter` package, you'd need to import `pkg/driver/node` package which would cause a circular dependency as `pkg/driver/node` imports `pkg/driver/node/mounter`. In situations like that, it's best to extract common things into a leaf package that doesn't import anything, which is what this PR does with `pkg/driver/node/volumecontext` package.

It also makes finding these volume context keys easier by placing them into a singular place.

---

### DIFF
--- a/pkg/driver/node/mounter/credential_provider.go
+++ b/pkg/driver/node/mounter/credential_provider.go
@@ -84,15 +84,15 @@ func (c *CredentialProvider) CleanupToken(volumeID string, podID string) error {
 
 // Provide provides mount credentials for given volume and volume context.
 // Depending on the configuration, it either returns driver-level or pod-level credentials.
-func (c *CredentialProvider) Provide(ctx context.Context, volumeID string, volumeContext map[string]string, mountpointArgs []string) (*MountCredentials, error) {
-	if volumeContext == nil {
+func (c *CredentialProvider) Provide(ctx context.Context, volumeID string, volumeCtx map[string]string, mountpointArgs []string) (*MountCredentials, error) {
+	if volumeCtx == nil {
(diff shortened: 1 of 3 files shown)


pkg/driver/node/mounter/credential_provider.go
```diff
@@ -96,7 +89,7 @@ func (c *CredentialProvider) Provide(ctx context.Context, volumeID string, volum
 		return nil, status.Error(codes.InvalidArgument, "Missing volume context")
 	}
 
-	authenticationSource := volumeContext[VolumeCtxAuthenticationSource]
+	authenticationSource := volumeContext[volumecontext.AuthenticationSource]
```
muddyfish, line 92: Not blocking: The only difference here being capitalisation makes this more challenging to read at first glance.

muddyfish (approved):

pkg/driver/node/mounter/credential_provider.go
unexge: I can rename variables to `volumeCtx`, to have similar naming convention to built-in `context` package and `ctx` variable used widely.

unexge: Addressed with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/340/commits/5da88fc2fe558f98706c7f8092be45cd35b6e511

muddyfish (approved):

PR #344, opened 2025-01-16
Make path argument variadic in `PathOnHost` and `PathInsideMountpointPod`

unexge:
This is to make this API nicer to use by having a similar semantics to `filepath.Join`

---

### DIFF
--- a/pkg/podmounter/mppod/path.go
+++ b/pkg/podmounter/mppod/path.go
@@ -16,12 +16,19 @@ const CommunicationDirName = "comm"
 
 // PathOnHost returns the full path on the host that refers to `path` inside Mountpoint Pod.
 // This function should be used in the CSI Driver Node Pod which uses `hostPath` volume to mount kubelet.
-func PathOnHost(podPathOnHost string, path string) string {
-	return filepath.Join(podPathOnHost, "/volumes/kubernetes.io~empty-dir/", CommunicationDirName, path)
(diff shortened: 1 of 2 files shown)


pkg/podmounter/mppod/path.go
```diff
@@ -16,12 +16,19 @@ const CommunicationDirName = "comm"
 
 // PathOnHost returns the full path on the host that refers to `path` inside Mountpoint Pod.
 // This function should be used in the CSI Driver Node Pod which uses `hostPath` volume to mount kubelet.
-func PathOnHost(podPathOnHost string, path string) string {
-	return filepath.Join(podPathOnHost, "/volumes/kubernetes.io~empty-dir/", CommunicationDirName, path)
+func PathOnHost(podPathOnHost string, path ...string) string {
+	return filepath.Join(append([]string{
```
muddyfish, line 20: Nit: Can we split these two onto multiple lines? I feel it would be more obvious how this works (prepend to a list, and then splatting that list)

muddyfish (dismissed):

pkg/podmounter/mppod/path.go
unexge: Addressed with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/344/commits/567a4e67099bc5f7e735e06c905111de55df0ec1

muddyfish (approved):

PR #345, opened 2025-01-16
Make sure `mountoptions.{Send, Recv}` respects `ctx`'s deadline

unexge:
`net.{Listen, Dial}` functions does not accept context argument, but only `{ListenContext, DialContext}` counterparts do. But even those functions does not respect to context's deadline, so we need to call `SetDeadline` with passed context's deadline explicitly to ensure context's deadline is respected.

---

### DIFF
--- a/pkg/podmounter/mountoptions/mount_options.go
+++ b/pkg/podmounter/mountoptions/mount_options.go
@@ -32,13 +32,22 @@ func Send(ctx context.Context, sockPath string, options Options) error {
 		return fmt.Errorf("failed to marshal message to send %s: %w", sockPath, err)
 	}
 
-	conn, err := net.Dial("unix", sockPath)
+	var d net.Dialer
+	conn, err := d.DialContext(ctx, "unix", sockPath)
 	if err != nil {
 		return fmt.Errorf("failed to dial to unix socket %s: %w", sockPath, err)
 	}
 	defer conn.Close()
 
 	unixConn := conn.(*net.UnixConn)
+
+	// `unixConn.WriteMsgUnix` does not respect `ctx`'s deadline, we need to call `unixConn.SetDeadline` to ensure `unixConn.WriteMsgUnix` has a deadline.
+	if deadline, ok := ctx.Deadline(); ok {
+		err := unixConn.SetDeadline(deadline)
+		if err != nil {
+			return fmt.Errorf("failed to set deadline on unix socket %s: %w", sockPath, err)
+		}
+	}
 
 	unixRights := syscall.UnixRights(options.Fd)
 	messageN, unixRightsN, err := unixConn.WriteMsgUnix(message, unixRights, nil)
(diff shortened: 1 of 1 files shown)


pkg/podmounter/mountoptions/mount_options.go
```diff
@@ -32,14 +32,23 @@ func Send(ctx context.Context, sockPath string, options Options) error {
...
+	conn, err := d.DialContext(ctx, "unix", sockPath)
 	if err != nil {
 		return fmt.Errorf("failed to dial to unix socket %s: %w", sockPath, err)
 	}
 	defer conn.Close()
 
 	unixConn := conn.(*net.UnixConn)
 
+	// `unixConn.WriteMsgUnix` does not respect `ctx`'s deadline, we need to call `unixConn.SetDeadline` to ensure `unixConn.WriteMsgUnix` has a deadline.
+	if deadline, ok := ctx.Deadline(); ok {
```
muddyfish, line 45: Can `ctx`'s deadline change, and if so is this mirrored automatically? Does it matter?

unexge:
> For my own understanding, why do we care about deadlines, and what are they used for?

It's for timeouts, usually contexts are used to indicate timeouts on the operations. For example, before calling `mountoptions.Recv`, we create a context with a timeout for this operation here https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/cmd/aws-s3-csi-mounter/main.go#L49-L52.

But without this change, that timeout is ignored basically, and that function might block forever until it's cancelled on a higher level somehow

pkg/podmounter/mountoptions/mount_options.go
unexge: This `ctx` is immutable basically, you can create a new context from this one with a new timeout via `context.WithTimeout(ctx, newTimeout)` but this would be a different context. Here we just respect the deadline/timeout passed to this function on the original call.

pkg/podmounter/mountoptions/mount_options.go
```diff
@@ -55,20 +64,30 @@ func Send(ctx context.Context, sockPath string, options Options) error {
...
-	l, err := net.Listen("unix", sockPath)
+	var lc net.ListenConfig
+	l, err := lc.Listen(ctx, "unix", sockPath)
 	if err != nil {
 		return Options{}, fmt.Errorf("failed to listen unix socket %s: %w", sockPath, err)
 	}
 	defer l.Close()
 
+	// `l.Accept` does not respect `ctx`'s deadline, we need to call `ul.SetDeadline` to ensure `l.Accept` has a deadline.
+	if deadline, ok := ctx.Deadline(); ok {
```
muddyfish, line 83: Is it useful to make this a function? It's repeated twice here and I'm not sure if we'd want to use this more often

unexge: So seems like these are two different types in two functions, `net.UnixListener` and `net.UnixConn` and they don't share any common interface. So we need to have a generic function with a type bound on `SetDeadline` method. Not sure if this is better or create confusion:
```go
type deadline interface {
	SetDeadline(t time.Time) error
}

func populateDeadlineFromContext[Conn deadline](ctx context.Context, conn Conn) error {
	if deadline, ok := ctx.Deadline(); ok {
		return conn.SetDeadline(deadline)
	}

	return nil
}
```

muddyfish (approved):

PR #346, opened 2025-01-16
Create a shorter symlink for Unix socket in `mountoptions` package

unexge:
Go returns `invalid argument` if you try to `net.{Dial, Listen}` on a Unix socket if the path is longer than 108 characters.

This limitation is only to the argument passed to those functions, if we receive a Unix socket path longer than 108 characters, we create a symlink for that path in temporary dir for that platform and use that path in network operations to work around this limitation.

We were just omitting a warning before, but it'd be hard to work around this limitation on the caller side, so that warning wouldn't be much actionable.

---

### DIFF
--- a/pkg/podmounter/mountoptions/mount_options.go
+++ b/pkg/podmounter/mountoptions/mount_options.go
@@ -8,7 +8,10 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"net"
+	"os"
+	"path/filepath"
 	"syscall"
 
 	"k8s.io/klog/v2"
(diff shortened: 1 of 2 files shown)


unexge:
Originally there were tests for this functionality in `PodMounter`, but now since this is a separate change I think it's fair to add tests here as well.

unexge:
Added tests with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/346/commits/8f044d48f3519765735849a5af5cf12856c4d7e9

muddyfish (approved):

unexge:
Realized this is only working on Unix and not on Linux, used an alternative approach in https://github.com/awslabs/mountpoint-s3-csi-driver/pull/368/commits/1aa2530d9ddbb3be0808e7f9ee9e8cb2054a8df1 which works in both.

PR #347, opened 2025-01-16
Write error log to `mount.err` if Mountpoint fails to start

unexge:
This is to let `PodMounter` doing this mount operation know that mount failed immediately instead of waiting for timeout, and also to propagate mount errors to users.

---

### DIFF
--- a/cmd/aws-s3-csi-mounter/csimounter/csimounter.go
+++ b/cmd/aws-s3-csi-mounter/csimounter/csimounter.go
@@ -1,7 +1,10 @@
 package csimounter
 
 import (
+	"bytes"
 	"fmt"
+	"io"
+	"io/fs"
 	"os"
 	"os/exec"
 	"slices"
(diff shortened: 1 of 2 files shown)


cmd/aws-s3-csi-mounter/csimounter/csimounter.go
```diff
@@ -62,12 +70,22 @@ func Run(options Options) (int, error) {
...
 	cmd.Stdout = os.Stdout
-	cmd.Stderr = os.Stderr
+	cmd.Stderr = io.MultiWriter(os.Stderr, &stderrBuf)
 
-	klog.Info("Starting Mountpoint process")
+	exitCode, err := options.CmdRunner(cmd)
+	if err != nil {
+		// If Mountpoint fails, write it to `mountErrorPath` to let `PodMounter` running in the same node know.
+		if writeErr := os.WriteFile(mountErrorPath, stderrBuf.Bytes(), mountErrorFileperm); writeErr != nil {
+			klog.Infof("Failed to write error logs to %s: %v\n", mountErrorPath, err)
```
muddyfish, line 85: Nit: we should mention these are mount-time error logs

muddyfish (approved):
Approved, though it would be nice to have that log line improved. I'm assuming it's intended that the actual usage of this file isn't ready yet?

unexge:
> I'm assuming it's intended that the actual usage of this file isn't ready yet?

Yes, it's splitted from https://github.com/awslabs/mountpoint-s3-csi-driver/tree/unexge/add-podmounter which adds `PodMounter` via this commit and uses this functionality https://github.com/awslabs/mountpoint-s3-csi-driver/commit/1177d060c370159d99a5edca66d316962140a212. Wanted to make this change a seperate PR for ease of review

cmd/aws-s3-csi-mounter/csimounter/csimounter.go
unexge: Addressed with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/347/commits/c32596f5165adde8f3b335c25cadeb02d0e25222

muddyfish (approved):

PR #349, opened 2025-01-20
Create `mountpoint.Args` for parsing and accessing Mountpoint args 

unexge:
Previously, accessing and manipulating of Mountpoint arguments was spread over to the codebase. This new mountpoint package contains all the logic related to Mountpoint arguments.

---


pkg/mountpoint/args.go
```diff
@@ -0,0 +1,144 @@
...
+
+	for _, arg := range passedArgs {
+		var key, value string
+
+		parts := strings.SplitN(strings.Trim(arg, " "), "=", 2)
+		if len(parts) == 2 {
+			// Ex: --key=value, key=value
+			key, value = parts[0], parts[1]
+		} else {
+			// Ex: --key value, key value
```
muddyfish, line 62: This section doesn't make sense to me what it's meant to catch. Don't we only support one key/value pair per argument?

pkg/mountpoint/args.go
```diff
@@ -0,0 +1,144 @@
...
+			// Ex: --key, key
+			parts = strings.SplitN(strings.Trim(parts[0], " "), " ", 2)
+			if len(parts) == 1 {
+				// Ex: --key, key
+				key = parts[0]
```
muddyfish, line 67: Nit: An explicit assignment of `value` to `ArgNoValue` would be helpful here.

pkg/mountpoint/args.go
```diff
@@ -0,0 +1,144 @@
...
+func (a *Args) Remove(key ArgKey) (ArgValue, bool) {
+	arg, exists := a.find(key)
+	if exists {
+		a.args.Delete(arg)
+	}
+	return arg.value, exists
+}
+
+// SortedList returns ordered list of normalized arguments.
+func (a *Args) SortedList() []string {
```
muddyfish, line 118: Nit: Why do we need this? If it's just for equality, couldn't we return a set? (Presumably this isn't called enough to care about time complexity though)

pkg/mountpoint/args.go
```diff
@@ -0,0 +1,144 @@
...
+		args = append(args, arg.String())
+	}
+	slices.Sort(args)
+	return args
+}
+
+// find tries to find given key from [Args], and returns whole entry, and whether the key was found.
+func (a *Args) find(key ArgKey) (Arg, bool) {
+	key = normalizeKey(key)
+	for _, arg := range a.args.UnsortedList() {
```
muddyfish, line 130: Can you remind me again why we can't use a hashmap here and avoid doing this iteration?

pkg/mountpoint/args.go
unexge: Yes we only support one argument per line, the comments lists some example usages which would fall into the associated blocks. Here, this block is if the user provide arguments like: `--key value` or `key value` (without `--` prefix) without using `=` between key and value.

pkg/mountpoint/args.go
unexge: It's used while passing these arguments to Mountpoint via [systemd](https://github.com/awslabs/mountpoint-s3-csi-driver/pull/349/files#diff-7d42d437046be1e45cd3cd87a2685779f91fce6ddd67d6d097d1fe395f7fb8cdR158) or via [exec.Command](https://github.com/awslabs/mountpoint-s3-csi-driver/pull/349/files#diff-69831569493f68f24f15945836d12837d45e5a5099bbf0e397001c4b6c91861eR65) and also in tests as you mentioned

pkg/mountpoint/args.go
unexge: It's to support multi arguments with the same key. I think iteration wouldn't matter here much as we'd be dealing with a dozen arguments at max, but also not sure if we have a multi argument use-case with Mountpoint currently. It was just to create an ideal argument parser, might not be needed with Mountpoint args.

pkg/mountpoint/args.go
muddyfish: Ah, could you make the comments '`--key value`, or `key value`'? Or some other way to make it obvious the comma isn't part of the example

pkg/driver/node/mounter/systemd_mounter.go
```diff
@@ -168,36 +169,18 @@ func (m *SystemdMounter) Mount(bucketName string, target string, credentials *Mo
...
-func addUserAgentToOptions(options []string, userAgent string) []string {
-	// first remove it from the options in case it's in there
-	for i := len(options) - 1; i >= 0; i-- {
-		if strings.Contains(options[i], userAgentPrefix) {
-			options = append(options[:i], options[i+1:]...)
-		}
-	}
-	// add the hard coded S3 CSI driver user agent string
-	return append(options, userAgentPrefix+"="+userAgent)
+func addUserAgentToArguments(args mountpoint.Args, userAgent string) mountpoint.Args {
```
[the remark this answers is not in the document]
unexge: Addressed with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/349/commits/e35a62924c0b845623bfa2f1028c53f0afe607ec

pkg/mountpoint/args.go
unexge: Addressed with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/349/commits/21a0fe0d891ae509f46d60aef81a05c514e179a5

pkg/mountpoint/args.go
unexge: Addressed with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/349/commits/21a0fe0d891ae509f46d60aef81a05c514e179a5

pkg/mountpoint/args_test.go
```diff
@@ -0,0 +1,403 @@
...
+				"--allow-other",
+				"--gid=2000",
+				"--uid=1000",
+			},
+		},
+		{
+			name: "with multiple spaces",
+			input: []string{
+				"--allow-other",
+				"--uid    1000",
```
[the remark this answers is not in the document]
unexge: Addressed with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/349/commits/41c14188bfda5eacea6c914373d82f9e2e810715

unexge:
Also added a new commit, [Use mountpoint.Args in CredentialProvider](https://github.com/awslabs/mountpoint-s3-csi-driver/pull/349/commits/35491f7592133dd77f94aa307b201563f92beb2b), the whole `CredentialProvider` was split into smaller parts in the original PR, but since this is a standalone PR, I also updated usages in `CredentialProvider` as well.

PR #351, opened 2025-01-21
Add reusable matrix creation workflow

muddyfish:
*Issue #, if available:* N/A

*Description of changes:* Adds a new workflow which generates a reusable matrix of K8s versions that we can use in other workflows (e2e tests)


unexge (approved):
Change LGTM, but not sure if I understand the motivation behind this.

muddyfish:
> Change LGTM, but not sure if I understand the motivation behind this.

I'm planning on creating a new workflow for manually deleting a cluster, and possibly also splitting out the controller tests from the e2e tests

PR #354, opened 2025-01-21
Allow overriding of hardcoded DaemonSet tolerations

### LINKED ISSUE #348: Additional taints conflict with default taints and cause constant restarts
/kind bug

**What happened?**

In our cluster we want to run the s3-csi-driver only on certain nodes using a nodeSelector. We can't use `tolerateAllTaints` because we are also running Cilium, which should be run with a startup taint so that pods are started after Cilium is initialised. So we are currently adding the node taint via the `tolerations` value.

What we discovered is that this seems to clash with the [hardcoded tolerations](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/charts/aws-mountpoint-s3-csi-driver/templates/node.yaml#L34-L38) when `tolerateAllTaints` is disabled. We can see that after 300s all the pods in the DaemonSet are restarted. This happens all the time.

**What you expected to happen?**

It would be great if the hardcoded tolerations could be disabled or overridden. A simple solution would be to move them from the template to the `tolerations` value.

**How to reproduce it (as minimally and precisely as possible)?**

Use the s3-csi-driver with `tolerateAllTaints: false` and any other additional `tolerations`.

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`): EKS v1.31
- Driver version: v1.11.0

jon-rei:
Resolves #348 

*Description of changes:*

This PR will move the hardcoded tolerations from the DaemonSet template to the values, ~~I just used the [aws-fsx-csi-driver](https://github.com/kubernetes-sigs/aws-fsx-csi-driver/blob/master/charts/aws-fsx-csi-driver/templates/node-daemonset.yaml#L39-L45) helm chart as inspiration. As this is probably the simplest solution, it may lead to unexpected changes if someone already has `tolerations` set.~~
I have added a new value: `node.defaultTolerations` and put it in the `else` condition in `tolerateAllTaints`.

---

### DIFF
--- a/charts/aws-mountpoint-s3-csi-driver/values.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/values.yaml
@@ -34,13 +34,10 @@ node:
       memory: 40Mi
     limits:
       memory: 256Mi
+  # Tolerates all taints and overrides defaultTolerations
   tolerateAllTaints: false
-  tolerations:
-    - key: CriticalAddonsOnly
-      operator: Exists
-    - operator: Exists
-      effect: NoExecute
-      tolerationSeconds: 300
+  defaultTolerations: true
+  tolerations: []
   affinity:
     nodeAffinity:
       requiredDuringSchedulingIgnoredDuringExecution:
--- a/charts/aws-mountpoint-s3-csi-driver/values.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/values.yaml
@@ -35,6 +35,12 @@ node:
     limits:
       memory: 256Mi
   tolerateAllTaints: false
+  tolerations:
+    - key: CriticalAddonsOnly
+      operator: Exists
+    - operator: Exists
+      effect: NoExecute
+      tolerationSeconds: 300
   affinity:
     nodeAffinity:
       requiredDuringSchedulingIgnoredDuringExecution:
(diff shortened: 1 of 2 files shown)


charts/aws-mountpoint-s3-csi-driver/values.yaml
```diff
@@ -35,6 +35,12 @@ node:
     limits:
       memory: 256Mi
   tolerateAllTaints: false
+  tolerations:
```
unexge, line 38: So with this change, if someone provides `node.tolerations`, it will override existing tolerations, which could be a breaking change I think. Imagine someone relies on default tolerations provided by the CSI Driver and with this change they will go away. Not sure how big is a deal this, one alternative could be creating a new variable like `node.defaultTolerations` for these tolerations, but then one would need to override two Helm values to tweak this configuration.

jon-rei: Yes, that was also something I was thinking about.

I guess for now, creating a new value is probably the best solution without introducing a breaking change. I have now added `node.defaultTolerations` and put it in the `else` condition in `tolerateAllTaints`. For me this is the simplest solution after playing around with helm templating.

unexge (approved):
The change looks good to me. Just need to solve CI failure, it's probably due to some recent changes in our CI and not related to this change.

muddyfish (approved):

unexge:
Thanks @jon-rei!

PR #357, opened 2025-01-24
Add Kubernetes v1.32.1 to CI

yerzhan7:
*Description of changes:* EKS now supports v1.32. 

- Added support for Kubernetes v1.32.1 to CI
- Upgraded `eksctl` to v0.202.0 as 0.191.0 does not support v1.32

- https://docs.aws.amazon.com/eks/latest/userguide/platform-versions.html#platform-versions-1.32
- https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.32.md

### DIFF
--- a/.github/matrix.yaml
+++ b/.github/matrix.yaml
@@ -2,7 +2,7 @@ matrix:
   cluster-type: [ "eksctl", "kops" ]
   arch: [ "x86", "arm" ]
   family: [ "AmazonLinux2", "Bottlerocket" ]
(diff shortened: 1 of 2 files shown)


unexge (dismissed):
LGTM, thank you!. It looks like the PR still uses the configuration from `main` though, we won't know if everything works until we merge this I think (we might need `eksctl` bump if they added 1.32 support via new version for example).

unexge (approved):

PR #358, opened 2025-01-24
Add CSI Driver version label to Mountpoint pods

renanmagagnin:
*Description of changes:*
We currently add Mountpoint version, Pod UID and Volume Name as labels to Mountpoint Pod. These labels are useful to query Mountpoint Pods, for example one can run `kubectl get pods -n mount-s3 -ls3.csi.aws.com/mountpoint-version=1.11.0 `to get Mountpoint Pods using Mountpoint v1.11.0.

In this PR, we add the CSI Driver's version as a new label. The Pod creator gets CSI Driver version from a new environment variable which is propagated throughout the call chain. Unit tests and integrations tests were also updated accordingly. 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ all-push-skip-if-present:
 
 .PHONY: build_image
 build_image:
-	DOCKER_BUILDKIT=1 docker buildx build -f ${DOCKERFILE} -t=${IMAGE}:${TAG} --platform=${PLATFORM} --build-arg VERSION=$(VERSION) .
+	DOCKER_BUILDKIT=1 docker buildx build -f ${DOCKERFILE} -t=${IMAGE}:${TAG} --platform=${PLATFORM} .
 
 .PHONY: push-manifest
 push-manifest: create-manifest
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ all-push-skip-if-present:
 
 .PHONY: build_image
 build_image:
-	DOCKER_BUILDKIT=1 docker buildx build -f ${DOCKERFILE} -t=${IMAGE}:${TAG} --platform=${PLATFORM} .
+	DOCKER_BUILDKIT=1 docker buildx build -f ${DOCKERFILE} -t=${IMAGE}:${TAG} --platform=${PLATFORM} --build-arg VERSION=$(VERSION) .
 
 .PHONY: push-manifest
 push-manifest: create-manifest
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -76,7 +76,7 @@ all-push-skip-if-present:
 
 .PHONY: build_image
 build_image:
-	DOCKER_BUILDKIT=1 docker buildx build -f ${DOCKERFILE} -t=${IMAGE}:${TAG} --platform=${PLATFORM} .
+	DOCKER_BUILDKIT=1 docker buildx build -f ${DOCKERFILE} -t=${IMAGE}:${TAG} --platform=${PLATFORM} --build-arg VERSION=$(VERSION) .
```
unexge, line 79: Looks like we already embed CSI Driver's version in Go binary, see this line in `Makefile`: https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/Makefile#L23

cmd/aws-s3-csi-controller/main.go
```diff
@@ -48,6 +49,7 @@ func main() {
 			Image:           *mountpointImage,
 			ImagePullPolicy: corev1.PullPolicy(*mountpointImagePullPolicy),
 		},
+		CSIDriverVersion: *csiDriverVersion,
```
unexge, line 52: We can just get driver's version via: `version.GetVersion().DriverVersion`, using version package `github.com/awslabs/aws-s3-csi-driver/pkg/driver/version`

pkg/podmounter/mppod/creator.go
```diff
@@ -10,9 +10,10 @@ import (
 
 // Labels populated on spawned Mountpoint Pods.
 const (
-	LabelVersion    = "s3.csi.aws.com/mountpoint-version"
-	LabelPodUID     = "s3.csi.aws.com/pod-uid"
-	LabelVolumeName = "s3.csi.aws.com/volume-name"
+	LabelVersion          = "s3.csi.aws.com/mountpoint-version"
```
unexge, line 13: Could you rename this to `LabelMountpointVersion` to make it clear?

pkg/podmounter/mppod/creator.go
```diff
@@ -10,9 +10,10 @@ import (
...
+	LabelPodUID           = "s3.csi.aws.com/pod-uid"
+	LabelVolumeName       = "s3.csi.aws.com/volume-name"
+	LabelCSIDriverVersion = "s3.csi.aws.com/csi-driver-version"
```
dannycjones, line 16: Would it make sense to be more like `mounted-by-csi-driver-version`? (bad name, but my thinking is that its a bit unclear right now)

pkg/podmounter/mppod/creator.go
```diff
@@ -24,9 +25,10 @@ type ContainerConfig struct {
 
 // A Config represents configuration for spawned Mountpoint Pods.
 type Config struct {
-	Namespace string
-	Version   string
-	Container ContainerConfig
+	Namespace        string
+	Version          string
```
unexge, line 29: Same with the label, could you rename it to `MountpointVersion`?

unexge (commented):
Could you also update [controller tests](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/tests/controller/controller_test.go#L687-L707) to assert this please?

pkg/podmounter/mppod/creator.go
renanmagagnin: Done, thanks!

cmd/aws-s3-csi-controller/main.go
renanmagagnin: I removed the new build-arg and used this instead, thanks!

pkg/podmounter/mppod/creator.go
renanmagagnin: Done!

pkg/podmounter/mppod/creator.go
renanmagagnin: Done!

unexge (approved):
LGTM. Thanks @renanmagagnin!

PR #362, opened 2025-01-27
Add `envprovider` package for accessing environment variables

unexge:

---

### DIFF
--- a/pkg/driver/node/envprovider/provider.go
+++ b/pkg/driver/node/envprovider/provider.go
@@ -1,0 +1,90 @@
+// Package envprovider provides utilities for accessing environment variables to pass Mountpoint.
+package envprovider
+
+import (
+	"fmt"
+	"os"
+	"slices"
+)
+
+const (
+	EnvRegion                = "AWS_REGION"
+	EnvDefaultRegion         = "AWS_DEFAULT_REGION"
+	EnvSTSRegionalEndpoints  = "AWS_STS_REGIONAL_ENDPOINTS"
+	EnvMaxAttempts           = "AWS_MAX_ATTEMPTS"
+	EnvProfile               = "AWS_PROFILE"
+	EnvConfigFile            = "AWS_CONFIG_FILE"
+	EnvSharedCredentialsFile = "AWS_SHARED_CREDENTIALS_FILE"
+	EnvRoleARN               = "AWS_ROLE_ARN"
+	EnvWebIdentityTokenFile  = "AWS_WEB_IDENTITY_TOKEN_FILE"
+	EnvEC2MetadataDisabled   = "AWS_EC2_METADATA_DISABLED"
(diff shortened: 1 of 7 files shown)


pkg/driver/node/envprovider/provider.go
```diff
@@ -0,0 +1,90 @@
...
+	EnvConfigFile            = "AWS_CONFIG_FILE"
+	EnvSharedCredentialsFile = "AWS_SHARED_CREDENTIALS_FILE"
+	EnvRoleARN               = "AWS_ROLE_ARN"
+	EnvWebIdentityTokenFile  = "AWS_WEB_IDENTITY_TOKEN_FILE"
+	EnvEC2MetadataDisabled   = "AWS_EC2_METADATA_DISABLED"
+	EnvAccessKeyID           = "AWS_ACCESS_KEY_ID"
+	EnvSecretAccessKey       = "AWS_SECRET_ACCESS_KEY"
+	EnvSessionToken          = "AWS_SESSION_TOKEN"
+
+	EnvMountpointCacheKey = "UNSTABLE_MOUNTPOINT_CACHE_KEY"
```
muddyfish, line 25: Nit: Not aligned

pkg/driver/node/envprovider/provider_test.go
```diff
@@ -0,0 +1,211 @@
...
+		{
+			name: "some allowed env vars set",
+			env: map[string]string{
+				"AWS_REGION":                 "us-west-1",
+				"AWS_DEFAULT_REGION":         "us-east-1",
+				"AWS_STS_REGIONAL_ENDPOINTS": "regional",
+				"AWS_MAX_ATTEMPTS":           "10",
+			},
+			want: []string{
+				"AWS_DEFAULT_REGION=us-east-1",
```
muddyfish, line 66: Missing test cases where environment variables don't exist

muddyfish: It's also missing cases where random environment variables we don't care about aren't passed through

unexge: The previous test-case `no env vars set` (L52-56) tests when there is no env variable set, and this one includes additional `AWS_MAX_ATTEMPTS=10` which shouldn't be passed. But I can move this to a separate test-case to make it clear.

pkg/driver/node/envprovider/provider.go
unexge: Addressed with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/362/commits/83d852bc785f51635dd8072772fe72713cfaba78. gofmt formats that way if there is an empty line between blocks, I wanted to separate this one from others as it's unstable, but it indeed seemed odd. Now it should be fine.

muddyfish (approved):

PR #366, opened 2025-01-31
Get default Mountpoint namespace from env in CSI Controller

unexge:
To have consistency with other options and to ensure we don't hard-code Mountpoint namespace, get it from `MOUNTPOINT_NAMESPACE` env by default.

---

### DIFF
--- a/cmd/aws-s3-csi-controller/main.go
+++ b/cmd/aws-s3-csi-controller/main.go
@@ -22,7 +22,7 @@ import (
 	"github.com/awslabs/aws-s3-csi-driver/pkg/podmounter/mppod"
 )
 
-var mountpointNamespace = flag.String("mountpoint-namespace", "mount-s3", "Namespace to spawn Mountpoint Pods in.")
+var mountpointNamespace = flag.String("mountpoint-namespace", os.Getenv("MOUNTPOINT_NAMESPACE"), "Namespace to spawn Mountpoint Pods in.")
 var mountpointVersion = flag.String("mountpoint-version", os.Getenv("MOUNTPOINT_VERSION"), "Version of Mountpoint within the given Mountpoint image.")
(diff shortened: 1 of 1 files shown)


cmd/aws-s3-csi-controller/main.go
```diff
@@ -22,7 +22,7 @@ import (
 	"github.com/awslabs/aws-s3-csi-driver/pkg/podmounter/mppod"
 )
 
-var mountpointNamespace = flag.String("mountpoint-namespace", "mount-s3", "Namespace to spawn Mountpoint Pods in.")
+var mountpointNamespace = flag.String("mountpoint-namespace", os.Getenv("MOUNTPOINT_NAMESPACE"), "Namespace to spawn Mountpoint Pods in.")
```
muddyfish, line 25: Shouldn't we also include an update to set this variable in this PR as well?

unexge: It's in an upcoming commit here https://github.com/awslabs/mountpoint-s3-csi-driver/commit/57b5699c24475ed080d56e5841e053fb9b0ab063#diff-0ea7410c82938b74ea30e5de3a925389ed5ab54bcaa447f3e59248faa1ae2f93R69-R70. I just wanted to send small unrelated changes first to not have a similar problem as https://github.com/awslabs/mountpoint-s3-csi-driver/pull/328

muddyfish: So currently this code isn't used right?

unexge: Yes, this binary isn't even installed in our Docker images yet, which will be with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/364, and then the previously shared commit will add Kubernetes manifest to use this binary. We only run tests for this component yet

muddyfish (approved):

PR #378, opened 2025-02-18
Create `mppod/watcher` package for watching Mountpoint Pods

unexge:
This PR introduces `mppod/watcher` package for properly watching Mountpoint Pods using Kubernetes [informers](https://pkg.go.dev/k8s.io/client-go/informers). Informers does caching and handles transient errors from the control plane.

Also fixes the potential race conditions mentioned in https://github.com/awslabs/mountpoint-s3-csi-driver/pull/368/files#r1950851325. The `Wait` method set up a watcher first, and then checks if the Pod already seen by doing a `Get` and if not, it just listens events from the watcher.

---

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -21,11 +21,13 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"time"
 
 	"github.com/awslabs/aws-s3-csi-driver/pkg/driver/node"
 	"github.com/awslabs/aws-s3-csi-driver/pkg/driver/node/credentialprovider"
 	"github.com/awslabs/aws-s3-csi-driver/pkg/driver/node/mounter"
 	"github.com/awslabs/aws-s3-csi-driver/pkg/driver/version"
+	"github.com/awslabs/aws-s3-csi-driver/pkg/podmounter/mppod/watcher"
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"google.golang.org/grpc"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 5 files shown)


pkg/driver/driver.go
```diff
@@ -40,6 +42,8 @@ const (
 	grpcServerMaxReceiveMessageSize = 1024 * 1024 * 2 // 2MB
 
 	unixSocketPerm = os.FileMode(0700) // only owner can write and read.
+
+	podWatcherResyncPeriod = 15 * time.Second
```
muddyfish, line 46: Why this number? 
This [forum post](https://groups.google.com/g/kubernetes-sig-api-machinery/c/PbSCXdLDno0) suggests it should be between a few minutes and an hour.

unexge: I don't have a strong preference, I originally searched on GitHub to see what other projects were using and most were using less than a minute and that's why I choose 15s:
- https://github.com/goodrain/rainbond/blob/d2b9c165aced4763004212a187f6c2ac6263181a/builder/job/job.go#L126
- https://github.com/getsentry/sentry-kubernetes/blob/42f5d8f98c6e4b684e5d3c1b40190d97913a5d8c/informers.go#L29
- https://github.com/kstone-io/kstone/blob/0175be48f20f12619f4db37ce081685c8c96ad2c/pkg/k8s/client.go#L80
- https://github.com/tektoncd/cli/blob/d025e8b3f1ac13cb18766138e69c083963f80129/pkg/pods/pod.go#L105
- https://github.com/elastic-ai/elastic-gpu-agent/blob/260910781d49ce2a639d0c890c3eb11ad75935af/pkg/kube/sitter.go#L61

One other reasoning was that, the CSI RPC calls from Kubernetes uses [2m timeout by default](https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/csi/csi_plugin.go#L54), and I wanted to pick a number that's less than that so in case of a failure, the resync period would allow us to recover within the same RPC call.

I think I can increase it to a minute to begin with, and we can tweak it later if we see any problem.

muddyfish (approved):

PR #380, opened 2025-02-19
Retry on retryable errors on `mountoptions.Send`

unexge:
There was a race condition between `mountoptions.Recv` & `mountoptions.Send` and it was causing some test [failures](https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/13389126567/job/37392589399#step:7:342) [time](https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/13293255514/job/37118910394#step:7:343) to [time](https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/13292235088/job/37115630157#step:7:342).

In `mountoptions.Recv`, we're calling [`net.Listen`](https://pkg.go.dev/net#Listen) first (which causes Unix socket to be created), and then we're calling [`UnixListener.Accept`](https://pkg.go.dev/net#UnixListener.Accept) to accept a connection from that socket. On the other side, we were calling `util.WaitForUnixSocket` to wait until the Unix socket exists, and then we're calling [`net.Dial`](https://pkg.go.dev/net#Dial) to connect that Unix socket in `mountoptions.Send`. 

The race condition was happening between `net.Listen` and `UnixListener.Accept` calls, as `mountoptions.Send` was waiting for `net.Listen` and then calling `net.Dial`, which was resulting an `connect: connection refused` error if `net.Dial` called before `UnixListener.Accept`.

With this PR, `mountoptions.Send` automatically retries on both `connect: connection refused` and `file does not exist` and removes `util.WaitForUnixSocket` as that's already handled with `mountoptions.Send`.

## Testing

I was able to reproduce this race condition with:
```bash
$ go test -count 1000 -failfast -run TestPodMounter/Mounting/Correctly_passes_mount_options ./pkg/driver/node/mounter/pod_mounter_test.go
--- FAIL: TestPodMounter (10.00s)
    --- FAIL: TestPodMounter/Mounting (10.00s)
        --- FAIL: TestPodMounter/Mounting/Correctly_passes_mount_options (10.00s)
            pod_mounter_test.go:63: Chdir(`/tmp/TestPodMounterMountingCorrectly_passes_mount_options2188697748/001`)
            pod_mounter_test.go:158: Expected no error, but got: failed to accept connection from unix socket pods/979f0c82-e7f8-4cac-a509-a4b54e39d0bb/volumes/kubernetes.io~empty-dir/comm/mount.sock: accept unix pods/979f0c82-e7f8-4cac-a509-a4b54e39d0bb
/volumes/kubernetes.io~empty-dir/comm/mount.sock: i/o timeout
FAIL
FAIL    command-line-arguments  10.626s
FAIL
```

prior to this change, and now I verified running the same test many times and didn't get the error:
```bash
$ go test -count 10000 -failfast  -timeout 30m -run TestPodMounter/Mounting/Correctly_passes_mount_options ./pkg/driver/node/mounter/pod_mounter_test.go
ok      command-line-arguments  1271.068s
```
---


pkg/driver/node/mounter/pod_mounter.go
```diff
@@ -138,14 +138,6 @@ func (pm *PodMounter) Mount(ctx context.Context, bucketName string, target strin
 	podMountSockPath := mppod.PathOnHost(podPath, mppod.KnownPathMountSock)
 	podMountErrorPath := mppod.PathOnHost(podPath, mppod.KnownPathMountError)
 
-	klog.V(4).Infof("Waiting for Mountpoint Pod %s to be ready to accept connections on %s", pod.Name, podMountSockPath)
-
-	// Wait until Mountpoint Pod is ready to accept connections
-	err = util.WaitForUnixSocket(mountpointPodReadinessTimeout, mountpointPodReadinessCheckInterval, podMountSockPath)
```
muddyfish, line 144: Why don't we need this anymore?

pkg/podmounter/mountoptions/mount_options.go
```diff
@@ -64,6 +63,40 @@ func Send(ctx context.Context, sockPath string, options Options) error {
...
+	err := wait.PollUntilContextCancel(ctx, unixSocketDialRetryInterval, true, func(ctx context.Context) (bool, error) {
+		conn, err := d.DialContext(ctx, "unix", sockPath)
+		if err == nil {
+			// no error, stop retrying and return the Unix connection.
+			unixConn = conn.(*net.UnixConn)
+			return true, nil
+		}
+
+		if errors.Is(err, syscall.ENOENT) ||
+			errors.Is(err, syscall.ECONNREFUSED) {
```
muddyfish, line 88: Can we change the indentation here? It's confusing that it's the same level as the actual body of the if

pkg/driver/node/mounter/pod_mounter.go
unexge: > With this PR, mountoptions.Send automatically retries on both connect: connection refused and file does not exist and removes util.WaitForUnixSocket as that's already handled with mountoptions.Send.

Basically the retry on `errors.Is(err, syscall.ENOENT)` achieves the same thing as `util.WaitForUnixSocket` and it's no longer needed.

pkg/podmounter/mountoptions/mount_options.go
unexge: It's `go fmt`'s format, I can either make it one line or keep like this.

muddyfish: I feel one line would be better

unexge: Updated with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/380/commits/e64586d7520cf7de986728255abbfcc3173ce462

Issue #108, opened 2023-12-13
AWS ROSA support


jsanchezmartinez:
/feature

**Is your feature request related to a problem? Please describe.**
Current Helm Chart not working in AWS ROSA.

**Describe the solution you'd like in detail**
AWS ROSA support.

**Additional context**
ImagePullBackOff
> I1213 11:10:02.072905 1 driver.go:61] Driver version: 1.1.0, Git commit: c681ab1f19ccba5976e3263f0e3df65718750369, build date: 2023-12-05T19:47:03Z, nodeID: ip-10-252-14-48.eu-west-1.compute.internal, mount-s3 version: 1.3.1
> I1213 11:10:02.075541 1 mount_linux.go:282] Detected umount with safe 'not mounted' behavior
> I1213 11:10:02.087546 1 driver.go:113] Listening for connections on address: &net.UnixAddr{Name:"/csi/csi.sock", Net:"unix"}
> 
> exitCode: 2

dlakhaws:
Thank you for the feature request. This is something we are tracking, we are planning to publish a public roadmap shortly and will provide updates on the timeline for adding this feature.

zonybob:
I'd like to second this request. We run on our own AWS infra (not ROSA), but it is still OpenShift atop RHCOS.
For me, the blockers were:

1. The autogenerated `/opt/mountpoint-s3-csi/bin` directory did not have the necessary SELinux context. I had to `chcon -R -t bin_t /opt/mountpoint-s3-csi/bin`
2. It seems SELinux also blocks the ability to open `/dev/pts/N` device files. I had to use `audit2allow` and follow https://access.redhat.com/solutions/7015317 to add the below

```
module fix_devpts 1.0;

require {
        type init_t;
        type devpts_t;
        class chr_file open;
}

#============= init_t ==============
allow init_t devpts_t:chr_file open;
```

After doing all of the above, I was able to use the driver to mount PVCs as documented.

I don't believe I'll be permitted to take these workarounds to any production environments, so I am very interested in seeing the driver work "out of the box" on RHCOS

gisjedi:
@dlakhaws Are external PRs accepted? We have interest in assisting with technical contributions to resolve, if that is the case.

akhilesh-delphix:
@dlakhaws : do you have any tentative date for this feature availability. We are planning to use S3 bucket as storage with in openshift cluster and eagerly waiting for this feature.

[...]

PR #383, opened 2025-02-19
Add VOLUME_MOUNT_GROUP capability in PodMounter

yerzhan7:
*Issue:* https://github.com/awslabs/mountpoint-s3-csi-driver/issues/108

#### Description

- **Problem**: Currently, if `fsGroup` is set on customer's container and container processes are not run as root (`0`), then customer containers won't be able to access mountpoint.
- **Solution**: Adding `VOLUME_MOUNT_GROUP` capability in PodMounter in order to support Pods that have security context set with value for `fsGroup`.
- **Implementation**: 
  - Adding new `VOLUME_MOUNT_GROUP` capability to CSI Node driver
  - Setting `--gid` to the value of `fsGroup` parameter
  - Additionally setting `--allow-other`, `--dir-mode=770` and `--file-mode=660`

#### Documentation: 
- https://kubernetes-csi.github.io/docs/support-fsgroup.html
- https://kubernetes.io/docs/tasks/configure-pod-container/security-context/


pkg/driver/node/node.go
```diff
@@ -115,6 +118,18 @@ func (ns *S3NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePubl
...
 	args := mountpoint.ParseArgs(mountpointArgs)
+
+	if capMount := volCap.GetMount(); capMount != nil && util.UsePodMounter() {
+		if volumeMountGroup := capMount.GetVolumeMountGroup(); capMount.GetVolumeMountGroup() != "" {
+			// We need to add the following flags to support fsGroup
+			// If these flags were already set by customer in PV mountOptions then we won't override them
+			args.SetIfAbsent(mountpoint.ArgGid, volumeMountGroup)
+			args.SetIfAbsent(mountpoint.ArgAllowOther, "")
+			args.SetIfAbsent(mountpoint.ArgDirMode, "770")
+			args.SetIfAbsent(mountpoint.ArgFileMode, "660")
```
unexge, line 129: This block can update some arguments partially, should we prevent that? Maybe we should set these only if `gid` is not passed by the customer?

unexge: Also, you can use `mountpoint.ArgNoValue` instead of empty string in `args.SetIfAbsent(mountpoint.ArgAllowOther, "")`, so it'd become `args.SetIfAbsent(mountpoint.ArgAllowOther, mountpoint.ArgNoValue)`

pkg/mountpoint/args.go
```diff
@@ -94,6 +97,13 @@ func (a *Args) Set(key ArgKey, value ArgValue) {
 	a.args.Insert(arg{key, value})
 }
 
+// SetIfAbsent sets value of given key only if that key does not exist.
+func (a *Args) SetIfAbsent(key ArgKey, value ArgValue) {
+	if !a.Has(key) {
+		a.args.Insert(arg{key, value})
```
unexge, line 103:
```suggestion
		a.Set(key, value)
```

or

```suggestion
                key = normalizeKey(key)
		a.args.Insert(arg{key, value})
```

`Set` method already normalizes the key, if we don't use that we should normalize here as well

unexge (approved):
Left some small comments but overall LGTM! Thanks, @yerzhan7!

pkg/driver/node/node.go
```diff
@@ -115,6 +118,18 @@ func (ns *S3NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePubl
 	}
 
 	args := mountpoint.ParseArgs(mountpointArgs)
+
+	if capMount := volCap.GetMount(); capMount != nil && util.UsePodMounter() {
+		if volumeMountGroup := capMount.GetVolumeMountGroup(); capMount.GetVolumeMountGroup() != "" {
```
unexge, line 123: Regarding my first comment, I was thinking of adding `args.Has(mountpoint.ArgGid)` check here. 
```suggestion
		if volumeMountGroup := capMount.GetVolumeMountGroup(); volumeMountGroup != "" && !args.Has(mountpoint.ArgGid) {
```

So we don't end up updating arguments partially if the user already provided a `gid` (and probably other settings)

unexge (approved):

PR #384, opened 2025-02-20
Speed up end-to-end tests

unexge:
[Recent change](https://github.com/awslabs/mountpoint-s3-csi-driver/pull/381) to add `PodMounter` tests to our CI made our slow end-to-end tests even slower, and it started timing out. This PR implements some improvements to speed up end-to-end tests:
- [Run end-to-end tests in parallel using ginkgo](https://github.com/awslabs/mountpoint-s3-csi-driver/commit/a1291d20a33e01e4f69328fff238090e7800f63d) – Most end-to-end tests can be run in parallel as they create isolated namespaces, pods, and volumes. Once exception is credential tests – which need to change some cluster-level resources – and they already [declared to run as serial](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/tests/e2e-kubernetes/testsuites/credentials.go#L266-L270). So, making tests run in parallel using `ginkgo` CLI should be safe to do.
- [Statically bind PV and PVC in tests](https://github.com/awslabs/mountpoint-s3-csi-driver/commit/f68ff1cdca5fbb7b9ff9599d963c01a9851d7351) – Waiting for PV and PVC to bound takes 6-10 seconds as I can see from the past test runs, with this change we pre-bind them to speed up binding process.
- [Use `createVolumeResourceWithMountOptions` whenever possible instead of `storageframework.CreateVolumeResource`](https://github.com/awslabs/mountpoint-s3-csi-driver/commit/5eacf95b56ca5d4b1dc5db733024c63d89dc531a) – To make previous optimization more effective.
- [Increase end-to-end test timeout to 60m](https://github.com/awslabs/mountpoint-s3-csi-driver/pull/384/commits/2b793e29e1e3e4fd17a856d800302a0cf3765885)
- [Use m7g.large instance type in arm64 clusters](https://github.com/awslabs/mountpoint-s3-csi-driver/pull/384/commits/7086ddc69648841f551036a4c9843947cc092ce6) – We were using `m7g.medium` instance type on `arm64` clusters which only has 1 CPU and takes longer to run the tests, `m7g.large` has the same specs as `c5.large` - which what we use on `x86` clusters.
---

### DIFF
--- a/tests/e2e-kubernetes/scripts/run.sh
+++ b/tests/e2e-kubernetes/scripts/run.sh
@@ -203,7 +203,7 @@ elif [[ "${ACTION}" == "install_driver" ]]; then
 elif [[ "${ACTION}" == "run_tests" ]]; then
   set +e
   pushd tests/e2e-kubernetes
-  KUBECONFIG=${KUBECONFIG} ginkgo -p -vv -timeout 30m -- --bucket-region=${REGION} --commit-id=${TAG} --bucket-prefix=${CLUSTER_NAME} --imds-available=true
+  KUBECONFIG=${KUBECONFIG} ginkgo -p -vv -timeout 60m -- --bucket-region=${REGION} --commit-id=${TAG} --bucket-prefix=${CLUSTER_NAME} --imds-available=true
   EXIT_CODE=$?
   print_cluster_info
   exit $EXIT_CODE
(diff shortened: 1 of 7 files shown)


tests/e2e-kubernetes/scripts/run.sh
```diff
@@ -55,13 +55,11 @@ KOPS_VERSION=1.28.5
...
 NODE_COUNT=${NODE_COUNT:-3}
 if [[ "${ARCH}" == "x86" ]]; then
-  INSTANCE_TYPE_DEFAULT=c5.large
   AMI_ID_DEFAULT=$(aws ssm get-parameters --names /aws/service/ami-amazon-linux-latest/al2023-ami-kernel-default-x86_64 --region ${REGION} --query 'Parameters[0].Value' --output text)
 else
-  INSTANCE_TYPE_DEFAULT=m7g.medium
   AMI_ID_DEFAULT=$(aws ssm get-parameters --names /aws/service/ami-amazon-linux-latest/al2023-ami-kernel-default-arm64 --region ${REGION} --query 'Parameters[0].Value' --output text)
 fi
-INSTANCE_TYPE=${INSTANCE_TYPE:-$INSTANCE_TYPE_DEFAULT}
+INSTANCE_TYPE=${INSTANCE_TYPE:-c5.large}
```
yerzhan7, line 62: Wouldn't it be broken for arm architecture as `c5.large` is an x86 instance? I.e. we would select arm64 AMI for x86 instance

yerzhan7 (approved):

muddyfish (approved):

PR #386, opened 2025-02-21
Support alternative partitions in e2e tests

muddyfish:
*Issue #, if available:* N/A

*Description of changes:* Removes hard-coded `aws` partition, and uses sts getCallerIdentity to retrieve an ARN which can be used to extract the current partition. Additionally, replaces busybox from dockerhub with busybox from ecr public.

### DIFF
--- a/tests/e2e-kubernetes/testsuites/credentials.go
+++ b/tests/e2e-kubernetes/testsuites/credentials.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	awsarn "github.com/aws/aws-sdk-go-v2/aws/arn"
 	"github.com/aws/aws-sdk-go-v2/aws/retry"
 	"github.com/aws/aws-sdk-go-v2/service/iam"
 	iamtypes "github.com/aws/aws-sdk-go-v2/service/iam/types"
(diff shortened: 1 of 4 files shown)


tests/e2e-kubernetes/testsuites/credentials.go
```diff
@@ -33,9 +34,9 @@ import (
 )
 
 const (
-	iamPolicyS3FullAccess     = "arn:aws:iam::aws:policy/AmazonS3FullAccess"
-	iamPolicyS3ReadOnlyAccess = "arn:aws:iam::aws:policy/AmazonS3ReadOnlyAccess"
-	iamPolicyS3NoAccess       = "arn:aws:iam::aws:policy/AmazonEC2ReadOnlyAccess" // `AmazonEC2ReadOnlyAccess` gives no S3 access
+	iamPolicyS3FullAccess     = "AmazonS3FullAccess"
+	iamPolicyS3ReadOnlyAccess = "AmazonS3ReadOnlyAccess"
+	iamPolicyS3NoAccess       = "AmazonEC2ReadOnlyAccess" // `AmazonEC2ReadOnlyAccess` gives no S3 access
```
yerzhan7, line 39: There is also partition hardcoded in this file: https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/tests/e2e-kubernetes/scripts/eksctl-patch.json

Do we need to adjust it too?

muddyfish: Possibly - I can't remember how I created my test China cluster

PR #387, opened 2025-02-21
Update Mountpoint Pod spec to tolerate all taints

unexge:


pkg/podmounter/mppod/creator.go
```diff
@@ -100,6 +100,16 @@ func (c *Creator) Create(pod *corev1.Pod, pvc *corev1.PersistentVolumeClaim) *co
 					},
 				},
 			},
+			Tolerations: []corev1.Toleration{
```
muddyfish, line 103: Thanks for the comments - the justification makes sense

muddyfish (approved):

PR #388, opened 2025-02-23
PodMounter: Add group read access to credentials if Pod is not root

yerzhan7:
*Issue:* https://github.com/awslabs/mountpoint-s3-csi-driver/issues/108

#### Description

- **Problem**: If `fsGroup` security context is applied on Mountpoint Pod and it's not run as default root user (`0`), then currently Mountpoint Pod fails to read credentials that are provided by CSI Driver Node Pod.
- **Solution**: Adding group read permissions to credentials folder/files if parent folder has non-zero gid.
- **Implementation**: 
  - Added logic to check gid of parent folder in Mount() method
  - Added new `CredentialDirPerm` and `CredentialFilePerm` fields to credential's `ProvideContext`
  - Use these fields when creating credential files/directory
- **Manual Testing**: Successfully validated this fix in personal ROSA cluster.


pkg/driver/node/mounter/pod_mounter.go
```diff
@@ -36,24 +36,28 @@ const targetDirPerm = fs.FileMode(0755)
 // This is mainly exposed for testing, in production platform-native function (`mountSyscallDefault`) will be used.
 type mountSyscall func(target string, args mountpoint.Args) (fd int, err error)
 
+type fileGroupID func(path string) (gid uint32, err error)
```
unexge, line 39: Seems like this is only exposed for testing, and I feel like it would be cleaner if we just change the `gid` of the parent directory in our tests to test different cases via [os.Chown](https://pkg.go.dev/os#Chown).

pkg/util/file.go
```diff
@@ -37,3 +45,18 @@ func ReplaceFile(destPath string, sourcePath string, perm fs.FileMode) error {
 
 	return nil
 }
+
+// FileGroupID returns gid of the file or directory.
+func FileGroupID(path string) (gid uint32, err error) {
```
unexge, line 50: Since this is a public function, it would be nice to add some tests for this function in `file_test.go`.

pkg/util/file.go
```diff
@@ -5,6 +5,14 @@ import (
...
 	"io/fs"
 	"os"
+	"syscall"
+)
+
+const (
+	FileMode600 = fs.FileMode(0600) // User: read/write, Group: none, Others: none
+	FileMode640 = fs.FileMode(0640) // User: read/write, Group: read-only, Others: none
+	FileMode700 = fs.FileMode(0700) // User: full access, Group: none, Others: none
+	FileMode750 = fs.FileMode(0750) // User: full access, Group: read/execute only, Others: none
```
unexge, line 15: I feel like it'd be nicer if we name these constants after their permissions, something like:
```suggestion
	FileModeUserReadWrite      = fs.FileMode(0600) // User: read/write, Group: none, Others: none
	FileModeUserGroupReadWrite = fs.FileMode(0640) // User: read/write, Group: read-only, Others: none
	FileModeUserFull           = fs.FileMode(0700) // User: full access, Group: none, Others: none
	FileModeUserFullGroupRead  = fs.FileMode(0750) // User: full access, Group: read/execute only, Others: none
```

Though, feel free to change them if you can come up with better names :)

unexge (commented):
Left some small comments, but overall it looks good to me. Thanks @yerzhan7!

pkg/driver/node/mounter/pod_mounter.go
yerzhan7: Using `os.Chown` requires root privileges (or `CAP_CHOWN` capability) and adding root privileges to unit tests is not a best practice.

pkg/util/file.go
yerzhan7: Will add unit tests in the next revision, thanks for reviewing!

pkg/util/file.go
yerzhan7: Will change in the next revision. Do not have better naming suggestion.

pkg/driver/node/mounter/pod_mounter.go
unexge: Agh, makes sense to keep it as is then!

pkg/util/file_test.go
```diff
@@ -0,0 +1,29 @@
...
+	"errors"
+	"io/fs"
+	"os"
+	"testing"
+
+	"github.com/awslabs/aws-s3-csi-driver/pkg/util/testutil/assert"
+)
+
+func TestFileGroupIDReturnsProcessGid(t *testing.T) {
+	tmpFile, err := os.CreateTemp("", "testfile")
```
unexge, line 13: FYI, you can use [`t.TempDir()`](https://pkg.go.dev/testing#T.TempDir) to get a unique temporary directory, and after the test case it will be cleaned up automatically.

yerzhan7: Thanks, I didn't know.

PR #397, opened 2025-02-26
Disable provenance on Docker images

unexge:
Our CI recently broken due to fact platform images being manifest lists rather than single images: https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/13540744547/job/37843704971?pr=393#step:7:769. According to https://github.com/docker/buildx/issues/1509#issuecomment-1378538197, disabling provenance might ensure that single images being built instead of manifest lists, also, seems like our current `./scripts/provenance.sh` was trying to disable provenance via some weird way?

Just giving CI a go to see if this fix the problem. 

---


unexge:
Properly fixed with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/398

PR #401, opened 2025-03-01
node/mounter: add a more robust check for corrupted mounts

psarna:
Checking for corrupted mounts with kubernetes/mount-utils package has an issue with networked filesystems, especially ones based on FUSE. Namely, a check based on os.Stat() often returns results cached by the OS kernel, and thus return a valid result even if the underlying filesystem is not responsive anymore, because its userspace part hang.

There are at least two approaches to fix it:
  1. Use a more powerful `statx` syscall on linux with a AT_STATX_FORCE_SYNC flag set. That explicitly asks for not getting results from cache and is more likely to detect a corrupted mount.
  2. Use ReadDir. That one I simply verified empirically to work better than os.Stat. It tends to not serve results from cache, and has an additional perk than it trips if the mount point is not a directory.

<strike>
This patch implements solution (2.). The 1st one is theoretically even more robust, but includes relying on an OS-specific syscall that is also not supported for all kernels as it's relatively new.

This check was proven to work in practice when s3 fuse mountpoint breaks and starts returning an ENOTCONN error!

P.S. kubernetes/mount-utils already relies on statx too, except it explicitly asks for cached results whenever possible... which is less likely to work for networked filesystems.
 ref: https://github.com/kubernetes/mount-utils/blob/fa592385ec1796a726846a2a26f61e3968b5a184/mount_linux.go#L432
</strike>

    edit: since ReadDir might bring undesired side effects of reading large
    directiories periodically, this PR now implements option (1.) -- statx.

### DIFF
--- a/pkg/driver/node/mounter/pod_mounter.go
+++ b/pkg/driver/node/mounter/pod_mounter.go
@@ -319,7 +319,7 @@ func (pm *PodMounter) closeFUSEDevFD(fd int) {
 // If the target dir does not exists it tries to create it.
 // If the target dir is corrupted (decided with `mount.IsCorruptedMnt`) it tries to unmount it to have a clean mount.
 func (pm *PodMounter) verifyOrSetupMountTarget(target string) error {
-	_, err := os.Stat(target)
+	err := verifyMountPointStatx(target)
 	if err == nil {
 		return nil
 	}
--- a/pkg/driver/node/mounter/pod_mounter_darwin.go
+++ b/pkg/driver/node/mounter/pod_mounter_darwin.go
@@ -2,10 +2,17 @@ package mounter
 
 import (
 	"errors"
+	"os"
 
 	"github.com/awslabs/aws-s3-csi-driver/pkg/mountpoint"
 )
 
 func (pm *PodMounter) mountSyscallDefault(_ string, _ mountpoint.Args) (int, error) {
 	return 0, errors.New("Only supported on Linux")
+}
+
+func verifyMountPointStatx(path string) error {
+	// statx is a Linux-specific syscall, let's simulate with os.Stat
+	_, err := os.Stat(path)
+	return err
 }
--- a/pkg/driver/node/mounter/pod_mounter_linux.go
+++ b/pkg/driver/node/mounter/pod_mounter_linux.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"syscall"
 
+	"golang.org/x/sys/unix"
 	"k8s.io/klog/v2"
 
 	"github.com/awslabs/aws-s3-csi-driver/pkg/mountpoint"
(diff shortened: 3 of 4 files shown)


psarna:
Is the CI failure related to my changes?
```
  [FAIL] [sig-storage] CSI Volumes [Driver: s3.csi.aws.com] [Testpattern: Pre-provisioned PV (default fs)] cache Cache Local [It] two containers in the same pod using the same cache [sig-storage]
```
sounds a tad random

pkg/driver/node/mounter/pod_mounter.go
```diff
@@ -319,7 +319,7 @@ func (pm *PodMounter) closeFUSEDevFD(fd int) {
 // If the target dir does not exists it tries to create it.
 // If the target dir is corrupted (decided with `mount.IsCorruptedMnt`) it tries to unmount it to have a clean mount.
 func (pm *PodMounter) verifyOrSetupMountTarget(target string) error {
-	_, err := os.Stat(target)
+	_, err := os.ReadDir(target)
```
unexge, line 322: If the `target` is already mounted with Mountpoint, this could cause listing the bucket root. Not sure if this would be desired. Maybe option 1) with `statx` would be more suitable?

psarna: sure, I'm fine with statx, but what's the repository's approach to unix-specific and in general OS-specific code? Kubernetes repo just assumes unix, [brings it in](https://github.com/kubernetes/mount-utils/blob/fa592385ec1796a726846a2a26f61e3968b5a184/mount_linux.go#L36C2-L36C25), and then handles the case of unsupported syscall. Is it acceptable here as well?

unexge: So the `Mount` method [called in `NodePublishVolume`](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/pkg/driver/node/node.go#L142), and `NodePublishVolume` is called every minute or so because we [enable `requiresRepublish`](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/charts/aws-mountpoint-s3-csi-driver/templates/csidriver.yaml#L15). The reason for this repeated call is to ensure credentials are up-to-date. After updating the credentials, if the `target` is a mount point, we [return early](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/pkg/driver/node/mounter/pod_mounter.go#L116-L127) before doing the whole mount procedure again, but this code and the part changed in `SystemdMounter` called in each `Mount` call regardless if there is a previous mount or not.

We usually put platform specific code in `filename_[platform].go` and do the platform specific handling there. See [pod_mounter_linux.go](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/pkg/driver/node/mounter/pod_mounter_linux.go) and `mountSyscallDefault` method as an example. We only support Linux at the moment, we add macOS specific code just for ease of development, so you can just put a no-op implementation for `filename_darwin.go` similar to [pod_mounter_darwin.go](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/pkg/driver/node/mounter/pod_mounter_darwin.go) if you want to go that direction.

psarna: Ah, I didn't know it gets periodically called, makes sense. I'll try the statx way then

psarna: @unexge done, actually looks quite decent for Darwin too since it can just fall back straight to os.Stat

unexge:
Thanks @psarna! Please feel free to open an issue if you still experience problems with corrupted mounts.

pkg/driver/node/mounter/pod_mounter.go
unexge: Looks good to me! Thanks @psarna

unexge (approved):

PR #402, opened 2025-03-03
Run Mountpoint Pod as non-root

yerzhan7:
### Description
- **Problem**: Want to enable `"restricted"` [Pod Security Standard](https://kubernetes.io/docs/concepts/security/pod-security-standards/)and run Mountpoint Pod as non-root user
- **Solution**: Add `"restricted"` label to `mount-s3` namespace and run mountpoint pod as non-root user
- **Implementation**:
  - Set `RunAsNonRoot` of Mountpoint Pod to `true` via Security Context.
  - Set `SeccompProfile` to `RuntimeDefault`
  - Set `RunAsUser`  to `1000` .
    - Note that we can't set USER in the docker image itself because currently we re-use the same image for CSI Node Driver which needs to run as root user.
  - Do not set `RunAsUser` in OpenShift because it assigns uid automatically (it fails otherwise)
    - Created helper method to determine whether cluster is OpenShift or not by checking for existence of OpenShift API Groups
  - Add `pod-security.kubernetes.io/enforce: restricted` label to `mount-s3` namespace
  - Reverted https://github.com/awslabs/mountpoint-s3-csi-driver/pull/388 because now we always need group read permission for credentials (no need for if-else logic). Changed default cred permissions from 600 to 640 and from 700 to 750.
  - Set `--allow-root` if customer did not provide `--allow-other` (i.e. customer is running as root uid). Otherwise, customer container with uid 0 won't be able to access mountpoint.
  - Adjusted non-root E2E tests to use different default test user
     
- **Manual Testing**: Run E2E tests also in OpenShift cluster. Additionally, verified manually with both `restricted` and `privileged` namespace/SCC.
- **Issue**: https://github.com/awslabs/mountpoint-s3-csi-driver/issues/108


cmd/aws-s3-csi-controller/main.go
```diff
@@ -52,6 +56,7 @@ func main() {
 			ImagePullPolicy: corev1.PullPolicy(*mountpointImagePullPolicy),
 		},
 		CSIDriverVersion: version.GetVersion().DriverVersion,
+		IsOpenShift:      isOpenShift(client, log),
```
muddyfish, line 59: This should be more general, similar to how Mountpoint does [S3 personalities](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3/src/s3.rs).

yerzhan7: Thanks for reviewing!

Would something like this make sense? let me know if you have better suggestions regarding naming

```
// ClusterVariant represents different Kubernetes distributions
type ClusterVariant int

const (
	DefaultKubernetes ClusterVariant = iota // Vanilla K8s
	OpenShift                               // OpenShift K8s
)

// MountpointPodUserID returns the appropriate RunAsUser for Mountpoint Pod based on the cluster variant.
func (c ClusterVariant) MountpointPodUserID() *int64 {
	if c == OpenShift {
		// OpenShift clusters automatically assign uid from predefined namespace range
		// https://www.redhat.com/en/blog/a-guide-to-openshift-and-uids
		return nil
	}

	return ptr.To(int64(1000)) // Default non-root Mountpoint user
}
```

cmd/aws-s3-csi-controller/main.go
```diff
@@ -63,3 +69,30 @@ func main() {
...
+	apiGroups, err := discoveryClient.ServerGroups()
+	if err != nil {
+		log.Error(err, "Failed to get API groups to determine cluster variant. Assuming this is Default Kubernetes variant")
+		return defaultVariant
+	}
+
+	// Check if the cluster is an OpenShift cluster by detecting the "config.openshift.io" API group
+	for _, group := range apiGroups.Groups {
+		if group.Name == "config.openshift.io" {
+			return util.OpenShift
```
unexge, line 93:
```suggestion
			log.Info("Detected OpenShift cluster variant")
			return util.OpenShift
```

Logging here that we detected OpenShift variant would be useful for debugging

pkg/driver/node/credentialprovider/awsprofile/awsprofiletest/aws_profile.go
```diff
@@ -12,24 +11,15 @@ import (
 	"github.com/awslabs/aws-s3-csi-driver/pkg/util/testutil/assert"
 )
 
-func AssertCredentialsFromAWSProfile(t *testing.T, profileName string, filePerm os.FileMode, configFile, credentialsFile, accessKeyID, secretAccessKey, sessionToken string) {
+func AssertCredentialsFromAWSProfile(t *testing.T, profileName, configFile, credentialsFile, accessKeyID, secretAccessKey, sessionToken string) {
 	t.Helper()
 
-	assertCredentialFilePermissions(t, configFile, filePerm)
```
unexge, line 18: It seems useful to assert that credential files has the correct permissions, should we keep it?

pkg/util/cluster_variant.go
```diff
@@ -0,0 +1,24 @@
+package util
```
unexge, line 1: It's not a good practice to overuse packages like `util`, `model` etc. in Go. Package names are part of the API. I'd prefer if we put this into a package name like `cluster`, and have `cluster.Variant` for enum and a method `cluster.DetectVariant()` to detect the variant.

See https://go.dev/blog/package-names

unexge (dismissed):
Left some small comments, but overall looks good to me! Feel free to address them as follow-up PRs.

unexge (approved):

PR #415, opened 2025-03-13
Ensure idempotence in post_test step of e2e-tests

renanmagagnin:
*Issue #, if available:*

*Description of changes:* The changes in the [previous PR](https://github.com/awslabs/mountpoint-s3-csi-driver/pull/410) correctly tagged the architecture specific images but failed when being executed more than once. This change fixes this by suppressing `ImageAlreadyExistsException`.

Tested the E2E tests workflow in my fork.

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -238,5 +238,6 @@ jobs:
               --repository-name ${REPOSITORY_NAME} \
               --image-tag ${TAG_PASSED}-linux-${ARCH}-amazon \
               --image-manifest "$MANIFEST" \
-              --region ${AWS_REGION}
+              --region ${AWS_REGION} 2>&1 | grep -q "ImageAlreadyExistsException" || \
+              { err=$?; [ $err -eq 1 ] || exit $err; }
           done


.github/workflows/e2e-tests.yaml
```diff
@@ -238,5 +238,6 @@ jobs:
               --repository-name ${REPOSITORY_NAME} \
               --image-tag ${TAG_PASSED}-linux-${ARCH}-amazon \
               --image-manifest "$MANIFEST" \
-              --region ${AWS_REGION}
-          done
\ No newline at end of file
+              --region ${AWS_REGION} 2>&1 | grep -q "ImageAlreadyExistsException" || \
+              { err=$?; [ $err -eq 1 ] || exit $err; }
```
muddyfish, line 242: What does this bit do?

unexge (approved):

.github/workflows/e2e-tests.yaml
unexge: Seems a bit cryptic, but I think it just tries to ignore `ImageAlreadyExistsException` errors otherwise just propagates the error from the original command. I'm okay with the change to fix the CI but would be nice if we can add a comment explaining what this bit does and why.

Issue #334, opened 2025-01-11
Feature - Support for Pod-Level Identity Using AWS Credentials


aminmr:
/feature

**Is your feature request related to a problem? Please describe.**
The Mountpoint for Amazon S3 CSI Driver currently supports pod-level credentials only through IRSA. This creates limitations for users of  S3-compatible storage solutions, such as Ceph. These users cannot utilize pod-level credentials effectively and are restricted to driver-level credential configurations.

**Describe the solution you'd like in detail**
I would like the ability to define pod-level credentials using the following AWS environment variables:

```AWS_ACCESS_KEY_ID```
```AWS_SECRET_ACCESS_KEY```
```AWS_SESSION_TOKEN```

For security purposes, these variables could be passed to the pods via Kubernetes secrets. This feature would improve flexibility for S3-compatible storage while allowing credentials to be managed at the pod level.

**Describe alternatives you've considered**
The only current alternative is to configure credentials at the driver level. However, this approach lacks the granularity needed for pod-level configurations and does not meet the requirements for non-EKS Kubernetes environments or S3-compatible backends.

I am also willing to contribute to the development of this feature if guidance and support from the maintainers are available.

vladem:
Hi @aminmr, thanks for the feature request! We'll be happy to gather more feedback on it before starting the implementation. Generally, the use of static credentials is [not recommended](https://docs.aws.amazon.com/IAM/latest/UserGuide/best-practices.html#bp-workloads-use-roles), but we do recognize that there are [use cases](https://docs.aws.amazon.com/IAM/latest/UserGuide/best-practices.html#update-access-keys) where their usage is required.

With IRSA, service accounts are [annotated](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/examples/kubernetes/static_provisioning/pod_level_identity.yaml#L80) with the role they need to assume. If we decide to proceed with the implementation, we'll need to define a way of associating static credentials with the service account.

aminmr:
Hi @vladem,

Thanks for the response! I understand the concerns about static credentials. If you decide to move forward, I'm happy to help define how to associate them with the service account through annotations.

Thanks!

pawanpnvda:
We have successfully used the pod-level IRSA to mount S3 buckets on EKS clusters. But as mentioned before, some S3-compatible storage solutions, such as Ceph, or OCI buckets can not use this approach.

[...]

PR #420, opened 2025-03-25
Feature/volume attributes level identity

duanhongyi:
More granular setting of accessKeyId and secretAccessKey

```
apiVersion: v1
kind: PersistentVolume
metadata:
  name: s3-pv
spec:
  capacity:
    storage: 1200Gi # ignored, required
  accessModes:
    - ReadWriteMany # supported options: ReadWriteMany / ReadOnlyMany
  storageClassName: "" # Required for static provisioning
  claimRef: # To ensure no other PVCs can claim this PV
    namespace: default # Namespace is required even though it's in "default" namespace.
    name: s3-pvc # Name of your PVC
  mountOptions:
    - allow-delete
    - force-path-style
    - region us-east-1
    - endpoint-url https://play.min.io
  csi:
    driver: s3.csi.aws.com # required
    volumeHandle: s3-csi-driver-volume # Must be unique
    volumeAttributes:
      bucketName: builds
      accessKeyId: Q3AM3UQ867SPQQA43P2F
      secretAccessKey: zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG
      authenticationSource: attr
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: s3-pvc
spec:
  accessModes:
    - ReadWriteMany # Supported options: ReadWriteMany / ReadOnlyMany
  storageClassName: "" # Required for static provisioning
  resources:
    requests:
      storage: 1200Gi # Ignored, required
  volumeName: s3-pv # Name of your PV
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: s3-app
  labels:
    app: s3-app
spec:
  replicas: 1
  selector:
    matchLabels:
      app: s3-app
  template:
    metadata:
      labels:
        app: s3-app
    spec:
      containers:
      - name: s3-app
        image: ubuntu
        command: ["/bin/sh"]
        args: ["-c", "echo 'Hello from the container!' >> /data/$(date -u).txt; tail -f /dev/null"]
        volumeMounts:
        - name: persistent-storage
          mountPath: /data
        ports:
        - containerPort: 80
      volumes:
      - name: persistent-storage
        persistentVolumeClaim:
          claimName: s3-pvc
```

### DIFF
--- a/pkg/driver/node/credentialprovider/provider.go
+++ b/pkg/driver/node/credentialprovider/provider.go
@@ -34,6 +34,7 @@ const (
 	AuthenticationSourceUnspecified AuthenticationSource = ""
 	AuthenticationSourceDriver      AuthenticationSource = "driver"
 	AuthenticationSourcePod         AuthenticationSource = "pod"
+	AuthenticationSourceAttr        AuthenticationSource = "attr"
 )
 
 // A Provider provides methods for accessing AWS credentials.
@@ -66,6 +67,10 @@ type ProvideContext struct {
 	PodNamespace         string
 	ServiceAccountTokens string
 	ServiceAccountName   string
+	// Authentication values in volume attributes
+	AccessKeyID     string
+	SecretAccessKey string
+	SessionToken    string
 	// StsRegion is the `stsRegion` parameter passed via volume attribute.
 	StsRegion string
 	// BucketRegion is the `--region` parameter passed via mount options.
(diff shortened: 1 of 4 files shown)


unexge:
Hey @duanhongyi, thanks for the pull request! I've some security concerns with this approach, putting long-term AWS credentials on `volumeAttributes` as plain-text could cause these values to be leaked. In driver-level, we ask people to store these in Kubernetes Secrets, and then we [pass these credentials using AWS Profile files](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/pkg/driver/node/credentialprovider/awsprofile/aws_profile.go) with appropriate credentials to make sure they're not world-readable. I think the same approach should be used here as well. I think it would be more suitable to use `nodePublishSecretRef` and Kubernetes Secrets as mentioned [here](https://github.com/awslabs/mountpoint-s3-csi-driver/issues/334#issuecomment-2613552946) alongside AWS Profiles.

One other thing is that, we'd need to add documentation, end-to-end tests, security tests etc. on top of the implementation. I prefer if the PR adds any changes with an opt-in flag to not enable this feature until we address other things.

duanhongyi:
OK

PR #422, opened 2025-03-28
Ensure CloudFormation stacks are deleted in cluster clean up

unexge:
The CI [seems to be broken](https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/14110780306/job/39538720856?pr=421) because for `s3-csi-cluster-eksctl-amazonlinux2-x86` cluster, the EKS cluster does not exist but the CloudFormation stack exists and stuck in `ROLLBACK_COMPLETE` status. Current CI script doesn't try to remove the CF stack if the cluster does not exist, this PR updates `eksctl_delete_cluster` to try to delete the CF stack even if the cluster does not exist.

---

### DIFF
--- a/tests/e2e-kubernetes/scripts/eksctl.sh
+++ b/tests/e2e-kubernetes/scripts/eksctl.sh
@@ -115,6 +115,9 @@ function eksctl_delete_cluster() {
   FORCE=${4:-false}
 
   if ! eksctl_cluster_exists "${BIN}" "${CLUSTER_NAME}"; then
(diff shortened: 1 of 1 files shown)


yerzhan7 (approved):
> this PR updates eksctl_delete_cluster to try to delete the CF stack even if the cluster does not exist.

I assume you've deleted the bad stack using this PR? (as a test that it works)

What was the error message in CFN stack that caused the rollback?

PR #423, opened 2025-03-30
PodMounter: Fix ReadOnlyMany access mode

yerzhan7:
*Issue #, if available:* N/A

*Description of changes:* Currently, PodMounter fails to work with `ReadOnlyMany` access mode with error:

```
Error: Mount options: --read-only are ignored with FUSE fd mount point.Mount options should be passed while performing `mount` syscall in the caller process.
```

Fix: Remove `--read-only` flag when it is set

Also, added e2e tests for ReadOnlyMany case to validate that the fix worked.

### DIFF
--- a/pkg/driver/node/mounter/pod_mounter_linux.go
+++ b/pkg/driver/node/mounter/pod_mounter_linux.go
@@ -47,8 +47,6 @@ func (pm *PodMounter) mountSyscallDefault(target string, args mountpoint.Args) (
 
 	if args.Has(mountpoint.ArgReadOnly) {
 		flags |= syscall.MS_RDONLY
-		// Remove the read-only argument from the list as mount-s3 does not support it when using FUSE file descriptor
-		args.Remove(mountpoint.ArgReadOnly)
 	}
 
 	if args.Has(mountpoint.ArgAllowOther) || args.Has(mountpoint.ArgAllowRoot) {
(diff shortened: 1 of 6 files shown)


pkg/driver/node/mounter/pod_mounter_linux.go
```diff
@@ -47,6 +47,8 @@ func (pm *PodMounter) mountSyscallDefault(target string, args mountpoint.Args) (
 
 	if args.Has(mountpoint.ArgReadOnly) {
 		flags |= syscall.MS_RDONLY
+		// Remove the read-only argument from the list as mount-s3 does not support it when using FUSE file descriptor
+		args.Remove(mountpoint.ArgReadOnly)
```
unexge, line 51: Should we move this to `PodMounter.Mount` function after calling this method? Seems like it would be better there rather than mutating the `args`. For example the caller might pass you a copy of `args` and this would stop working

tests/e2e-kubernetes/testsuites/accessmode.go
```diff
@@ -0,0 +1,89 @@
+package custom_testsuites
```
unexge, line 1: Nice!

pkg/driver/node/mounter/pod_mounter_linux.go
yerzhan7: Good point! Moved to `PodMounter.Mount`

unexge (approved):
Thanks for catching and fixing this @yerzhan7!

Issue #427, opened 2025-04-08
Provided resource requests and limits for initContainers in aws-mountpoint-s3-csi-driver


sc-harish:
/feature

**Is your feature request related to a problem? Please describe.**
Our cluster requires resource requests/limits on initContainers, but we cannot set them right now.

**Describe the solution you'd like in detail**
I'll submit a PR

**Additional context**
As part of this [PR](https://github.com/awslabs/mountpoint-s3-csi-driver/pull/116), I noticed that resource limits are set for the other containers but not for the init containers. If the cluster enforces resource constraints, this could prevent the pod from being created.

sc-harish:
Requesting review for the mentioned [PR](https://github.com/awslabs/mountpoint-s3-csi-driver/pull/428)

dannycjones:
This change was released in [Mountpoint CSI Driver v1.14.1](https://github.com/awslabs/mountpoint-s3-csi-driver/releases/tag/v1.14.1). Please let us know if there's any issues.

Thank you for both the feature request and the contribution of the feature, @sc-harish!

PR #428, opened 2025-04-08
Init container resource limits

sc-harish:
*Issue:*
https://github.com/awslabs/mountpoint-s3-csi-driver/issues/427

*Description of changes:*
Added resource constraints to the init container and parameterized the values using templates.

### DIFF
--- a/charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
@@ -70,7 +70,7 @@ spec:
           volumeMounts:
             - name: mp-install
               mountPath: /target
-          {{- with default .Values.node.resources .Values.initContainer.install_mountpoint.resources }}
+          {{- with default .Values.node.resources .Values.initContainer.installMountpoint.resources }}
           resources: {{- toYaml . | nindent 12 }}
           {{- end }}
       {{- end }}
--- a/charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
@@ -70,7 +70,7 @@ spec:
           volumeMounts:
             - name: mp-install
               mountPath: /target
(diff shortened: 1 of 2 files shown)


charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
```diff
@@ -70,6 +70,9 @@ spec:
           volumeMounts:
             - name: mp-install
               mountPath: /target
+          {{- with default .Values.node.resources .Values.initContainer.install-mountpoint.resources }}
```
unexge, line 73: The idea here is to use `node.resources` as a default value if `initContainer.install-mountpoint.resources` is not specified right? Since we set a default value in `values.yaml` for `initContainer.install-mountpoint.resources`, I feel like it would always be non-empty and never use `node.resources`. Could you check if that's the case by running `helm template ./charts/aws-mountpoint-s3-csi-driver | grep initContainers -A 30` please? Maybe we should just leave `initContainer.install-mountpoint.resources` commented in `values.yaml`?

sc-harish: Thanks for taking the time to review this. From what I understand, if resources: {} is explicitly set in values.yaml, Helm treats it as an empty value—so the with block gets skipped. As a result, it falls back to using node.resources instead. I tested this using the command you shared, and it seems to behave that way. Please correct me if I'm wrong here.

**resources set to {} in values.yaml, it uses resources value set for node**
Screenshot 2025-04-09 at 10 03 25 PM

**resources set as non-empty in values.yaml for initContainers**
image
image

unexge: Sorry for late response, yes looks like you're right! As one last thing, could you please rename Helm variable to `initContainer.installMountpoint` to have consistent naming with other variables?

unexge:
Thanks for the contribution @sc-harish!

unexge (approved):

PR #430, opened 2025-04-10
Add Mountpoint debugging flags to all E2E tests

renanmagagnin:


tests/e2e-kubernetes/testsuites/util.go
```diff
@@ -105,6 +106,14 @@ func createVolumeResource(ctx context.Context, config *storageframework.PerTestC
 	pvName := "s3-e2e-pv-" + uuid.New().String()
 	pvcName := "s3-e2e-pvc-" + uuid.New().String()
 
+	// Add debug options if they're not already present
+	debugOptions := []string{"debug", "debug-crt"}
+	for _, option := range debugOptions {
+		if !slices.Contains(mountOptions, option) {
```
unexge, line 112: Maybe we should parse `mountOptions` with [`mountpoint.ParseArgs`](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/pkg/mountpoint/args.go#L54) here and use that to query presence of an option? If the caller uses `--debug` to enable debug logs, this would duplicate it for example.

tests/e2e-kubernetes/testsuites/mountoptions.go
```diff
@@ -83,6 +83,41 @@ func (t *s3CSIMountOptionsTestSuite) DefineTests(driver storageframework.TestDri
 		ginkgo.DeferCleanup(cleanup)
 	})
 
+	ginkgo.FIt("should add debug MountOptions in createVolumeResource", func(ctx context.Context) {
```
unexge, line 86:
```suggestion
	ginkgo.It("should add debug MountOptions in createVolumeResource", func(ctx context.Context) {
```

unexge (changes requested):

tests/e2e-kubernetes/testsuites/util.go
```diff
@@ -105,6 +106,15 @@ func createVolumeResource(ctx context.Context, config *storageframework.PerTestC
 	pvName := "s3-e2e-pv-" + uuid.New().String()
 	pvcName := "s3-e2e-pvc-" + uuid.New().String()
 
+	// Add debug options if they're not already present
+	normalizedOptions := mountpoint.ParseArgs(mountOptions)
```
unexge, line 110: I think we can simply this code block by using methods of `mountpoint.Args`:
```go
options := mountpoint.ParseArgs(mountOptions)
options.Set(mountpoint.ArgDebug, mountpoint.ArgNoValue) 
options.Set(mountpoint.ArgDebugCRT, mountpoint.ArgNoValue)
// ..


pv := &v1.PersistentVolume{
	ObjectMeta: metav1.ObjectMeta{
		Name: pvName,
	},
	Spec: v1.PersistentVolumeSpec{
		// ...
		MountOptions: options.SortedList(),
	},
}
```

`ArgDebug` and `ArgDebugCRT` needs to be added as constants to `pkg/mountpoint/args.go`.

renanmagagnin: Done, thank you!

unexge (dismissed):

unexge (approved):

Issue #353, opened 2025-01-21
Support sharing Mountpoint instances between multiple Pods


unexge:
Currently, the CSI Driver spawns one Mountpoint instance per Pod. This is good for isolation, but also in some cases sharing Mountpoint instances between multiple Pods - if the Pods are using same volume and credentials - would improve prefetch/cache efficiency of Mountpoint and overall reduce memory usage as Mountpoint instances are not free.

This should be ideally and opt-in flag, in addition to [access mode](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes) of the volume.

unexge:
Starting with v2.0.0, the CSI Driver automatically shares Mountpoint Pods between multiple workload Pods if they're running on the same host, using the same credentials and the same volume. See [Mountpoint Pod sharing](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/docs/MOUNTPOINT_POD_SHARING.md) for more details.

See https://github.com/awslabs/mountpoint-s3-csi-driver/issues/504 for more details about v2, and upgrading to it.

PR #439, opened 2025-04-23
PodMounter: Add Mountpoint Pod Sharing Support

yerzhan7:
*Issue:* https://github.com/awslabs/mountpoint-s3-csi-driver/issues/353

*Description of changes:*  This PR adds support for sharing Mountpoint Pods between multiple workload pods when they have compatible configurations. This optimization reduces resource usage by reusing existing Mountpoint Pods instead of creating a new one for each workload pod.

*Key Changes:*
- Custom Resource Definition
  - Added MountpointS3PodAttachment CRD to track workload-to-Mountpoint Pod attachments

- Controller Changes
  - Added reconciler logic to create/manage MountpointS3PodAttachment resources
  - Added stale attachment cleaner to handle orphaned attachments
  - Implemented expectations pattern to handle eventual consistency
  - Added cleanup of resources when pods are terminated

- Node Changes
  - Modified PodMounter to use shared source mount points and perform bind mounts on target path
  - Added dedicated PodUnmounter to handle clean unmounting
  - Added `needs-unmount` annotation handling for graceful pod cleanup
  - Added locking mechanisms to prevent race conditions

- Testing
  - Added e2e and controller tests covering various pod sharing scenarios
  - Manual testing performed on personal EKS cluster


cmd/aws-s3-csi-controller/csicontroller/reconciler.go
```diff
@@ -163,56 +177,269 @@ func (r *Reconciler) spawnOrDeleteMountpointPodIfNeeded(
...
+	}
+
+	return r.addWorkloadToS3PodAttachment(ctx, s3pa, workloadUID, log)
+}
+
+func (r *Reconciler) addWorkloadToS3PodAttachment(ctx context.Context, s3pa *crdv1.MountpointS3PodAttachment, workloadUID string, log logr.Logger) (bool, error) {
+	log.Info("Adding workload UID to MountpointS3PodAttachment", "workloadUID", workloadUID)
+
+	for key := range s3pa.Spec.MountpointS3PodToWorkloadPodUIDs {
+		s3pa.Spec.MountpointS3PodToWorkloadPodUIDs[key] = append(s3pa.Spec.MountpointS3PodToWorkloadPodUIDs[key], workloadUID)
```
unexge, line 304: Why do we add to the first (or rather a random element as [map iteration is undefined in Go](https://go.dev/blog/maps#iteration-order)) element here? Why this is a map in the first place?

cmd/aws-s3-csi-controller/csicontroller/reconciler.go
```diff
@@ -163,56 +177,269 @@ func (r *Reconciler) spawnOrDeleteMountpointPodIfNeeded(
...
+	}
+	s3pa := &crdv1.MountpointS3PodAttachment{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: "s3pa-",
+			Labels: map[string]string{
+				LabelCSIDriverVersion: r.mountpointPodConfig.CSIDriverVersion,
+			},
+		},
+		Spec: crdv1.MountpointS3PodAttachmentSpec{
+			NodeName:             workloadPod.Spec.NodeName,
```
unexge, line 414: Should we also include bucket name here – even though it's immutable?

cmd/aws-s3-csi-controller/csicontroller/reconciler.go
```diff
@@ -163,56 +177,269 @@ func (r *Reconciler) spawnOrDeleteMountpointPodIfNeeded(
...
+			},
+		},
+	}
+	if authSource == "pod" {
+		s3pa.Spec.WorkloadNamespace = workloadPod.Namespace
+		s3pa.Spec.WorkloadServiceAccountName = getServiceAccountName(workloadPod)
+
+		roleARN, err := r.findIRSAServiceAccountRole(ctx, workloadPod)
+		if err != nil {
+			return err
```
unexge, line 431: Maybe we should perform this check before spawning MP Pod? If we run into this error, it would leave things in partial state (i.e., there is an MP Pod but no attachment)

cmd/aws-s3-csi-controller/csicontroller/reconciler.go
```diff
@@ -163,56 +177,269 @@ func (r *Reconciler) spawnOrDeleteMountpointPodIfNeeded(
...
+		}
+		s3pa.Spec.WorkloadServiceAccountIAMRoleARN = roleARN
+	}
+
+	err = r.Create(ctx, s3pa)
+	if err != nil {
+		log.Error(err, "Failed to create MountpointS3PodAttachment")
```
unexge, line 438: Should we delete spawned MP Pod here?

cmd/aws-s3-csi-controller/main.go
```diff
@@ -66,3 +87,25 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+func IndexMountpointS3PodAttachmentFields(log logr.Logger, mgr manager.Manager) {
```
unexge, line 91: Would be nice to add a comment for what this function does and why its needed

pkg/api/v1/groupversion_info.go
```diff
@@ -0,0 +1,20 @@
+// Package v1 contains API Schema definitions for the s3.csi.aws.com v1 API group.
```
unexge, line 1: Maybe we should start with `v1beta` or `v1alpha` version here?

pkg/api/v1/mountpoints3podattachment_types.go
```diff
@@ -0,0 +1,79 @@
...
+	FieldAuthenticationSource             = "spec.authenticationSource"
+	FieldWorkloadFSGroup                  = "spec.workloadFSGroup"
+	FieldWorkloadServiceAccountName       = "spec.workloadServiceAccountName"
+	FieldWorkloadNamespace                = "spec.workloadNamespace"
+	FieldWorkloadServiceAccountIAMRoleARN = "spec.workloadServiceAccountIAMRoleARN"
+)
+
+// MountpointS3PodAttachmentSpec defines the desired state of MountpointS3PodAttachment.
+type MountpointS3PodAttachmentSpec struct {
+	// Important: Run "make generate" to regenerate code after modifying this file
```
unexge, line 22: Would be nice to enforce this as part of our CI if there is an easy way to do

cmd/aws-s3-csi-controller/csicontroller/reconciler.go
```diff
@@ -163,56 +177,269 @@ func (r *Reconciler) spawnOrDeleteMountpointPodIfNeeded(
...
+
+	if s3paContainsWorkload(s3pa, workloadUID) {
+		log.Info("MountpointS3PodAttachment already has this workload UID")
+		return false, nil
+	}
+
+	return r.addWorkloadToS3PodAttachment(ctx, s3pa, workloadUID, log)
+}
+
+func (r *Reconciler) addWorkloadToS3PodAttachment(ctx context.Context, s3pa *crdv1.MountpointS3PodAttachment, workloadUID string, log logr.Logger) (bool, error) {
```
unexge, line 300: Would be nice to have some sanity checks in this function (e.g., checking if `s3pa`s and `workloadPod`'s node, pvName etc. fields match)

pkg/driver/node/mounter/mppod_lock.go
```diff
@@ -0,0 +1,55 @@
...
+// releaseMPPodLock decrements the reference count for a pod's lock.
+// When the reference count reaches zero, the lock is removed from the map.
+// If the lock doesn't exist, the function returns silently.
+func releaseMPPodLock(mpPodUID string) {
+	mpPodLocksMutex.Lock()
+	defer mpPodLocksMutex.Unlock()
+
+	lock, exists := mpPodLocks[mpPodUID]
+	if !exists {
+		// Should never happen
```
unexge, line 46: Maybe we can emit an error log?

pkg/driver/node/mounter/pod_mounter.go
```diff
@@ -69,36 +73,96 @@ func NewPodMounter(podWatcher *watcher.Watcher, credProvider *credentialprovider
...
+		if err != nil {
+			klog.Errorf("Failed to find source mount point for %q: %v", target, err)
+			return fmt.Errorf("Failed to find source mount point for %q: %w", target, err)
+		}
+		mpPodUID := filepath.Base(source)
+		podPath := pm.podPath(mpPodUID)
+
+		mpPodLock := getMPPodLock(mpPodUID)
+		mpPodLock.mutex.Lock()
+		defer func() {
```
unexge, line 104: Go's `defer`s are function-scoped rather than block-scoped, it would be better if we can have dedicated functions that acquire lock and release on return. This could result in hard to debug deadlocks as its a long function and hard to reason about when the lock will be released

pkg/driver/node/mounter/pod_mounter.go
```diff
@@ -384,3 +432,48 @@ func (pm *PodMounter) volumeNameFromTargetPath(target string) (string, error) {
 func (pm *PodMounter) helpMessageForGettingMountpointLogs(pod *corev1.Pod) string {
 	return fmt.Sprintf("You can see Mountpoint logs by running: `kubectl logs -n %s %s`. If the Mountpoint Pod already restarted, you can also pass `--previous` to get logs from the previous run.", pod.Namespace, pod.Name)
 }
+
+func (pm *PodMounter) getS3PodAttachmentWithRetry(ctx context.Context, volumeName string, credentialCtx credentialprovider.ProvideContext, fsGroup, pvMountOptions string) (*crdv1.MountpointS3PodAttachment, string, error) {
+	fieldFilters := client.MatchingFields{
+		crdv1.FieldNodeName:             os.Getenv("CSI_NODE_NAME"), // TODO
```
unexge, line 438: We have this information in `S3NodeServer.NodeID`

pkg/driver/node/mounter/pod_mounter.go
```diff
@@ -384,3 +432,48 @@ func (pm *PodMounter) volumeNameFromTargetPath(target string) (string, error) {
...
+		crdv1.FieldPersistentVolumeName: volumeName,
+		crdv1.FieldVolumeID:             credentialCtx.VolumeID,
+		crdv1.FieldMountOptions:         pvMountOptions,
+		crdv1.FieldWorkloadFSGroup:      fsGroup,
+		crdv1.FieldAuthenticationSource: credentialCtx.AuthenticationSource,
+	}
+	if credentialCtx.AuthenticationSource == credentialprovider.AuthenticationSourcePod {
+		fieldFilters[crdv1.FieldWorkloadNamespace] = credentialCtx.PodNamespace
+		fieldFilters[crdv1.FieldWorkloadServiceAccountName] = credentialCtx.ServiceAccountName
```
unexge, line 447: Is it intentional that this one doesn't include role ARN? Ideally, this logic should be extracted into a common module and re-used between the node and the controller

pkg/driver/node/mounter/pod_mounter.go
```diff
@@ -69,36 +73,96 @@ func NewPodMounter(podWatcher *watcher.Watcher, credProvider *credentialprovider
...
+
+		podCredentialsPath, err := pm.ensureCredentialsDirExists(podPath)
+		if err != nil {
+			klog.Errorf("Failed to create credentials directory for %q: %v", target, err)
+			return fmt.Errorf("Failed to create credentials directory for %q: %w", target, err)
+		}
+
+		credentialCtx.SetWriteAndEnvPath(podCredentialsPath, mppod.PathInsideMountpointPod(mppod.KnownPathCredentials))
+
+		_, _, err = pm.credProvider.Provide(ctx, credentialCtx)
```
unexge, line 117: Since this procedure of refreshing credentials are done in two places, we should ideally extract it into a separate function

pkg/driver/node/mounter/pod_mounter.go
```diff
@@ -107,126 +171,98 @@ func (pm *PodMounter) Mount(ctx context.Context, bucketName string, target strin
...
 	}
 
-	if isMountPoint {
-		klog.V(4).Infof("Target path %q is already mounted", target)
-		return nil
-	}
-
-	env := envprovider.Default()
-	env.Merge(credEnv)
+	if !isSourceMountPoint {
```
unexge, line 178: Would be nice to extract inside of this if block into a dedicated function

pkg/driver/node/mounter/pod_unmounter.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+func (u *PodUnmounter) HandleS3PodAttachmentUpdate(old, new any) {
+	s3pa := new.(*crdv1.MountpointS3PodAttachment)
+	if s3pa.Spec.NodeName != u.nodeID {
+		return
+	}
+
+	for mpPodName, uids := range s3pa.Spec.MountpointS3PodToWorkloadPodUIDs {
+		if len(uids) == 0 {
```
unexge, line 51: Wouldn't this cause race conditions if the node decides to unmount an MP Pod and the controller decides to assign a new workload Pod to this MP Pod? We should ideally mark this MP Pod as deleted from the controller and check that here

pkg/driver/node/mounter/pod_unmounter.go
```diff
@@ -0,0 +1,197 @@
...
+		WritePath: filepath.Join(util.KubeletPath(), "pods", mpPodUID),
+	})
+	if err != nil {
+		klog.Errorf("Failed to clean up credentials for %s: %v", source, err)
+		return err
+	}
+	return nil
+}
+
+func (u *PodUnmounter) CleanupDanglingMounts() {
```
unexge, line 119: Ideally this should be a goroutine that runs periodically

cmd/aws-s3-csi-controller/csicontroller/reconciler.go
```diff
@@ -4,34 +4,47 @@ import (
...
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	crdv1 "github.com/awslabs/aws-s3-csi-driver/pkg/api/v1"
+	"github.com/awslabs/aws-s3-csi-driver/pkg/driver/node/volumecontext"
 	"github.com/awslabs/aws-s3-csi-driver/pkg/podmounter/mppod"
+	"github.com/go-logr/logr"
```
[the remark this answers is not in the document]
yerzhan7: I think yes because `logf.FromContext(ctx).WithValues(...)` returns `logr.Logger` interface.

anurag4DSB:
What is the motivation for transitioning from a systemd-based mounter to a pod-based mounter?
Is the primary reason scalability, or are there other factors driving this change?

unexge:
Hey @anurag4DSB, see https://github.com/awslabs/mountpoint-s3-csi-driver/issues/279 and the linked issues for more details on the motivation why we're moving away from `systemd` mounter. This PR specifically makes progress towards https://github.com/awslabs/mountpoint-s3-csi-driver/issues/353.

anurag4DSB:
> Hey @anurag4DSB, see #279 and the linked issues for more details on the motivation why we're moving away from `systemd` mounter. This PR specifically makes progress towards #353.

Thank you so much. Appreciate the prompt response.

unexge (approved):
This is a great work @yerzhan7! Left some comments, but feel free to address them as follow-up PRs. I think we can merge this PR as-is.

pkg/api/v1/mountpoints3podattachment_types.go
unexge: Would be still nice to have a way to detect if this is "dirty" meaning it requires us to run `make generate`, but it's not urgent

yerzhan7 (commented):
Thanks for reviewing! Fixed merge conflicts and resolved small/quick comments in https://github.com/awslabs/mountpoint-s3-csi-driver/pull/439/commits/eeb1699f3bdd8681bad7d67b1a6a489ad05d32d0

Will address other comments in separate PRs.

unexge (approved):

cmd/aws-s3-csi-controller/csicontroller/reconciler.go
```diff
@@ -163,56 +183,323 @@ func (r *Reconciler) spawnOrDeleteMountpointPodIfNeeded(
...
+	return fieldFilters
+}
+
+// getAuthSource returns authentication source from given PV.
+// Defaults to `driver` if `authenticationSource` is not found in volume attributes.
+func (r *Reconciler) getAuthSource(pv *corev1.PersistentVolume) string {
+	volumeAttributes := mppod.ExtractVolumeAttributes(pv)
+	authSource := volumeAttributes[volumecontext.AuthenticationSource]
+	if authSource == "" {
+		return "driver"
```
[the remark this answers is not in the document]
yerzhan7: Yeah, this logic is duplicated, will address it separately.

cmd/aws-s3-csi-controller/csicontroller/reconciler.go
```diff
@@ -163,56 +183,323 @@ func (r *Reconciler) spawnOrDeleteMountpointPodIfNeeded(
...
+	fieldFilters := client.MatchingFields{
+		crdv1beta.FieldNodeName:             workloadPod.Spec.NodeName,
+		crdv1beta.FieldPersistentVolumeName: pv.Name,
+		crdv1beta.FieldVolumeID:             pv.Spec.CSI.VolumeHandle,
+		crdv1beta.FieldMountOptions:         strings.Join(pv.Spec.MountOptions, ","),
+		crdv1beta.FieldWorkloadFSGroup:      fsGroup,
+		crdv1beta.FieldAuthenticationSource: authSource,
+	}
+
+	if authSource == "pod" {
```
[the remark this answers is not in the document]
yerzhan7: Fixed. Will address moving to different package separately.

cmd/aws-s3-csi-controller/csicontroller/reconciler.go
```diff
@@ -163,56 +183,323 @@ func (r *Reconciler) spawnOrDeleteMountpointPodIfNeeded(
...
+		return &s3paList.Items[0], nil
+	default:
+		return nil, fmt.Errorf("found %d MountpointS3PodAttachments when expecting 0 or 1", len(s3paList.Items))
+	}
+}
+
+// handleInactivePod handles inactive workload pod.
+func (r *Reconciler) handleInactivePod(ctx context.Context, s3pa *crdv1beta.MountpointS3PodAttachment, workloadUID string, log logr.Logger) (bool, error) {
+	if s3pa == nil {
+		log.Info("Workload pod is not active. Did not find any MountpointS3PodAttachments.")
```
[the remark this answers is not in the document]
yerzhan7: We can add pending expectation check, but most likely if we requeue inactive pod event next time reconciler would just skip it completely because by that time Workload Pod won't be found. I.e. cleanup will be handled by periodic stale attachment cleanup job.

Since `handleInactivePod()` is "best effort" anyway at the moment I don't think we need this extra check. 

Maybe even we don't need `handleInactivePod()` method at all, and we should rely fully on stale attachment cleaner instead.

cmd/aws-s3-csi-controller/csicontroller/reconciler.go
```diff
@@ -163,56 +183,323 @@ func (r *Reconciler) spawnOrDeleteMountpointPodIfNeeded(
...
+
+	return r.addWorkloadToS3PodAttachment(ctx, s3pa, workloadUID, log)
+}
+
+// addWorkloadToS3PodAttachment adds workload UID to the first Mountpoint Pod in the map
+// TODO: We will later add extra logic for selecting/creating MPPod if existing MP Pods are using old CSI Driver version or have some "no-new-attachments" annotation
+func (r *Reconciler) addWorkloadToS3PodAttachment(ctx context.Context, s3pa *crdv1beta.MountpointS3PodAttachment, workloadUID string, log logr.Logger) (bool, error) {
+	log.Info("Adding workload UID to MountpointS3PodAttachment")
+
+	for key := range s3pa.Spec.MountpointS3PodAttachments {
```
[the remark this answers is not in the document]
yerzhan7: Good point - we might need to handle this case together when we fix the TODO above.

PR #441, opened 2025-04-24
Check allowed licenses in GitHub CI

unexge:
This PR adds a new `make` target `make check_licenses` to ensure all Go dependencies has compatible licenses using [`go-licenses`](https://github.com/google/go-licenses) tool. Also adds this new target to our unit test GitHub workflow.

---


dannycjones (approved):
Thanks Burak, this is great!

I'm a bit sad, it looks like `go-licenses` doesn't have a way to clarify the licenses for those packages with warnings. This PR is already a big improvement though.

PR #443, opened 2025-04-25
Move `mountoptions` package to `github.com/awslabs/aws-s3-csi-driver/pkg/mountpoint`

unexge:
This is part of the effort of consolidating Mountpoint related functionalities in `github.com/awslabs/aws-s3-csi-driver/pkg/mountpoint` package.

---


pkg/mountpoint/mountoptions/mount_options.go
yerzhan7, line 1:
```
	"k8s.io/apimachinery/pkg/util/wait"
	"k8s.io/klog/v2"
```

Should these be removed?

yerzhan7 (approved):

pkg/mountpoint/mountoptions/mount_options.go
unexge: I expect these packages to work outside of Kubernetes, but will replace them if we found out if that's not the case.

PR #444, opened 2025-04-28
Move `mount`, `unmount` and `isMountPoint` functionalities to `github.com/awslabs/aws-s3-csi-driver/pkg/mountpoint/mounter`

unexge:
This is part of the effort of consolidating Mountpoint related
functionalities in `github.com/awslabs/aws-s3-csi-driver/pkg/mountpoint`
package.

---

### DIFF
--- a/pkg/driver/node/mounter/pod_mounter.go
+++ b/pkg/driver/node/mounter/pod_mounter.go
@@ -76,7 +76,7 @@ func (pm *PodMounter) Mount(ctx context.Context, bucketName string, target strin
 
 	podID := credentialCtx.PodID
 
-	isMountPoint, err := pm.mount.IsMountPoint(target)
+	isMountPoint, err := pm.IsMountPoint(target)
 	if err != nil {
 		err = pm.verifyOrSetupMountTarget(target, err)
 		if err != nil {
@@ -229,7 +229,7 @@ func (pm *PodMounter) Unmount(ctx context.Context, target string, credentialCtx 
 
 // IsMountPoint returns whether given `target` is a `mount-s3` mount.
 func (pm *PodMounter) IsMountPoint(target string) (bool, error) {
-	return pm.mount.IsMountPoint(target)
+	return pm.mount.CheckMountPoint(target)
 }
 
 // waitForMountpointPod waints until Mountpoint Pod for given `podID` and `volumeName` is in `Running` state.
--- a/pkg/driver/node/mounter/pod_mounter.go
+++ b/pkg/driver/node/mounter/pod_mounter.go
@@ -7,18 +7,17 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
-	"syscall"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
-	"k8s.io/mount-utils"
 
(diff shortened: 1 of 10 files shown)


muddyfish:
LGTM, though I'd want someone with more context on this to approve as well

pkg/driver/node/mounter/pod_mounter.go
```diff
@@ -77,14 +76,12 @@ func (pm *PodMounter) Mount(ctx context.Context, bucketName string, target strin
 
 	podID := credentialCtx.PodID
 
-	err = pm.verifyOrSetupMountTarget(target)
+	isMountPoint, err := pm.mount.IsMountPoint(target)
```
yerzhan7, line 79: Any reason for changing order between `verifyOrSetupMountTarget()` and `IsMountPoint()`?

The rest LGTM.

unexge: The `verifyOrSetupMountTarget` was doing a `statx`  syscall and checking the returned error to understand if its needs to create the target folder or re-mount it. Now the idea is to move `statx` call to `IsMountPoint` and if it returns an error just propagate it, so `verifyOrSetupMountTarget` can still handle any returned error from `statx`. `IsMountPoint` was doing regular `stat` syscall before and was only retuning error on "not found" error. 

I realized I forgot to use `statx` in `IsMountPoint`, will do it now. And I can also change the name of `IsMountPoint` to `CheckMountPoint` maybe, as now it will report the errors that `verifyOrSetupMountTarget` will fix instead of simply checking whether the target is mount point.

unexge: Addressed with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/444/commits/0bdc09aaed2a3954a7c91bd2ab992db6c6c15957 also updated doc comments on `CheckMountPoint`.

pkg/driver/node/mounter/systemd_mounter.go
```diff
@@ -85,25 +88,21 @@ func (m *SystemdMounter) Mount(ctx context.Context, bucketName string, target st
...
 			if mntErr := m.Unmount(ctx, target, credentialprovider.CleanupContext{
 				WritePath: credentialCtx.WritePath,
 				PodID:     credentialCtx.PodID,
 				VolumeID:  credentialCtx.VolumeID,
 			}); mntErr != nil {
-				return fmt.Errorf("Unable to unmount the target %q : %v, %v", target, mountpointErr, mntErr)
+				return fmt.Errorf("Unable to unmount the target %q : %v, %v", target, err, mntErr)
 			}
+		} else {
+			return fmt.Errorf("Could not check if %q is a mount point: %v, %v", target, err, err)
```
yerzhan7, line 102: We are printing the same error twice here?

unexge: I think it looks like that due to diff, I double-checked the file, and it doesn't seem to be the case https://github.com/awslabs/mountpoint-s3-csi-driver/blob/0bdc09aaed2a3954a7c91bd2ab992db6c6c15957/pkg/driver/node/mounter/systemd_mounter.go#L75-L104

unexge: Sorry misunderstood, fixed with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/444/commits/48e3115c97e7310189af2a9bcbe1616de10d7885

yerzhan7 (approved):

PR #453, opened 2025-05-06
Pod sharing fixes and improvements

yerzhan7:
*Issue #, if available:* N/A

*Description of changes:* This is follow-up to previous PR https://github.com/awslabs/mountpoint-s3-csi-driver/pull/439 to address remaining feedback:
- Makefile: Install `controller-gen` tool
- Stale Attachment Cleaner: Change periodicity to 2min
- Reconciler: Add empty annotation check
- Node: 
  - Fix bug where during NPV call for existing target mountpoint we do not provide IRSA Role ARN to credential context
    - Now even if target is mountpoint we would still get PodAttachment in order to get that Role ARN from reconciler  
    - Remove `findSourceMountPoint` method as it's no longer needed (and to simplify logic)
  - PodUnmounter:
     - Add log message
     - Remove unnecessary mutex
     - Fix bug where we provide incorrect WritePath to CleanupCredential context


Makefile
```diff
@@ -207,18 +207,26 @@ generate:
 	cat $(TMP_POD_ATTACHMENT_CRD_FILE) >> $(TEST_POD_ATTACHMENT_CRD_FILE)
 	rm $(TMP_POD_ATTACHMENT_CRD_FILE)
 
-## Binaries used in tests.
+## Tool Binaries
 
-TESTBIN ?= $(shell pwd)/tests/bin
-$(TESTBIN):
-	mkdir -p $(TESTBIN)
+TOOLS_BIN ?= $(shell pwd)/tools/bin
```
unexge, line 212: We should probably add `tools/bin` to `.gitignore`

unexge (approved):

PR #458, opened 2025-05-08
EKS Pod Identity support for pod level credentials

renanmagagnin:
*Description of changes:* With this change, we will support the configuration of pod level credentials with EKS Pod Identity. In EKS clusters, this is an alternative to IRSA with an easier configuration process.

*Key Changes:*

- Credential Provider
  - In `provider_pod.go` the precedence is given to IRSA credentials. If they are not configured, then EKS Pod Identity credentials are forwarded to the Mountpoint process instead.
  - Added unit tests

- Testing
  - Added E2E tests covering EKS Pod Identity setup scenarios with IAM roles of varying levels of S3 access
  - Added one pod level IRSA E2E test to make sure driver level EKS Pod Identity is not used instead
  - Added E2E tests covering setup scenario of IRSA and EKS Pod Identity
  - Manual testing performed on personal EKS cluster

### DIFF
--- a/charts/aws-mountpoint-s3-csi-driver/templates/csidriver.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/csidriver.yaml
@@ -12,6 +12,8 @@ spec:
   tokenRequests:
     - audience: "sts.amazonaws.com"
       expirationSeconds: 3600
+    {{- if .Values.experimental.podMounter }}
     - audience: "pods.eks.amazonaws.com"
       expirationSeconds: 3600
+    {{- end }}
   requiresRepublish: true
--- a/charts/aws-mountpoint-s3-csi-driver/templates/csidriver.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/csidriver.yaml
@@ -12,4 +12,6 @@ spec:
   tokenRequests:
     - audience: "sts.amazonaws.com"
       expirationSeconds: 3600
+    - audience: "pods.eks.amazonaws.com"
+      expirationSeconds: 3600
   requiresRepublish: true
--- a/pkg/driver/node/credentialprovider/provider_pod.go
+++ b/pkg/driver/node/credentialprovider/provider_pod.go
@@ -48,56 +48,51 @@ func (c *Provider) provideFromPod(ctx context.Context, provideCtx ProvideContext
 		return nil, status.Errorf(codes.InvalidArgument, "Failed to parse service account tokens: %v", err)
 	}
 
-	// stsToken := tokens[serviceAccountTokenAudienceSTS]
-	stsToken := tokens[serviceAccountTokenAudiencePodIdentity]
-	if stsToken == nil {
-		klog.Errorf("credentialprovider: `authenticationSource` configured to `pod` but no service account tokens for %s received. Please make sure to enable `podInfoOnMountCompat`, see "+podLevelCredentialsDocsPage, serviceAccountTokenAudiencePodIdentity)
-		return nil, status.Errorf(codes.InvalidArgument, "Missing service account token for %s", serviceAccountTokenAudiencePodIdentity)
-	}
-
-	// roleARN, err := c.findPodServiceAccountRole(ctx, provideCtx)
-	// if err != nil {
-	// 	return nil, err
-	// }
-
-	// region, err := c.stsRegion(provideCtx)
-	// if err != nil {
-	// 	return nil, status.Errorf(codes.InvalidArgument, "Failed to detect STS AWS Region, please explicitly set the AWS Region, see "+stsConfigDocsPage)
-	// }
-
-	// defaultRegion := os.Getenv(envprovider.EnvDefaultRegion)
-	// if defaultRegion == "" {
-	// 	defaultRegion = region
-	// }
-
 	podID := provideCtx.PodID
 	volumeID := provideCtx.VolumeID
 	if podID == "" {
 		return nil, status.Error(codes.InvalidArgument, "Missing Pod info. Please make sure to enable `podInfoOnMountCompat`, see "+podLevelCredentialsDocsPage)
 	}
 
-	tokenName := podLevelServiceAccountTokenName(podID, volumeID)
-
-	err := renameio.WriteFile(filepath.Join(provideCtx.WritePath, tokenName), []byte(stsToken.Token), CredentialFilePerm)
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "Failed to write service account token: %v", err)
+	stsToken := tokens[serviceAccountTokenAudienceSTS]
+	if stsToken == nil {
+		klog.Errorf("credentialprovider: `authenticationSource` configured to `pod` but no service account tokens for %s received. Please make sure to enable `podInfoOnMountCompat`, see "+podLevelCredentialsDocsPage, serviceAccountTokenAudienceSTS)
+		return nil, status.Errorf(codes.InvalidArgument, "Missing service account token for %s", serviceAccountTokenAudienceSTS)
 	}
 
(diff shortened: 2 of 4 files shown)


charts/aws-mountpoint-s3-csi-driver/templates/csidriver.yaml
```diff
@@ -12,4 +12,6 @@ spec:
   tokenRequests:
     - audience: "sts.amazonaws.com"
       expirationSeconds: 3600
+    - audience: "pods.eks.amazonaws.com"
+      expirationSeconds: 3600
```
yerzhan7, line 16: Should we wrap it with `{{- if .Values.experimental.podMounter -}}`?

Similar also in `provideFromPod()` method below with `usePodMounter`

pkg/driver/node/credentialprovider/provider_pod.go
```diff
@@ -19,8 +19,10 @@ import (
 )
 
 const (
-	serviceAccountTokenAudienceSTS = "sts.amazonaws.com"
-	serviceAccountRoleAnnotation   = "eks.amazonaws.com/role-arn"
+	serviceAccountTokenAudienceSTS         = "sts.amazonaws.com"
+	serviceAccountTokenAudiencePodIdentity = "pods.eks.amazonaws.com"
+	serviceAccountRoleAnnotation           = "eks.amazonaws.com/role-arn"
+	podIdentityCredURI                     = "http://169.254.170.23/v1/credentials"
```
yerzhan7, line 25: As a comment, could you please provide doc link where it says that URI should be this value?

pkg/driver/node/credentialprovider/provider_pod.go
```diff
@@ -48,56 +55,68 @@ func (c *Provider) provideFromPod(ctx context.Context, provideCtx ProvideContext
...
-		klog.Errorf("credentialprovider: `authenticationSource` configured to `pod` but no service account tokens for %s received. Please make sure to enable `podInfoOnMountCompat`, see "+podLevelCredentialsDocsPage, serviceAccountTokenAudienceSTS)
+		klog.Errorf("credentialprovider: `authenticationSource` configured to `pod` but no service account token for %s received. Please make sure to enable `podInfoOnMountCompat`, see "+podLevelCredentialsDocsPage, serviceAccountTokenAudienceSTS)
 		return nil, status.Errorf(codes.InvalidArgument, "Missing service account token for %s", serviceAccountTokenAudienceSTS)
 	}
 
-	roleARN, err := c.findPodServiceAccountRole(ctx, provideCtx)
-	if err != nil {
-		return nil, err
+	eksToken := tokens[serviceAccountTokenAudiencePodIdentity]
+	if eksToken == nil {
```
yerzhan7, line 63: Just checking: Are we still going to receive Pod Identity token even if Pod Identity Agent is not installed and/or no associations are created? 

In other words, are we always expecting to receive 2 tokens (STS and EKS) in the request?

pkg/driver/node/credentialprovider/provider_pod.go
```diff
@@ -141,7 +160,53 @@ func (c *Provider) findPodServiceAccountRole(ctx context.Context, provideCtx Pro
...
+	}
+
+	tokenName := podLevelServiceAccountTokenName(provideCtx.PodID, provideCtx.VolumeID)
+	tokenFile := filepath.Join(provideCtx.EnvPath, tokenName)
+
+	return envprovider.Environment{
+		envprovider.EnvRoleARN:              roleARN,
+		envprovider.EnvWebIdentityTokenFile: tokenFile,
+		envprovider.EnvRegion:               region,
+		envprovider.EnvDefaultRegion:        defaultRegion,
```
yerzhan7, line 199: Curious why we are passing `region` and `defaultRegion` for IRSA, but not for Pod Identity?

pkg/driver/node/credentialprovider/provider_pod.go
yerzhan7, line 1: I think we also need to update `cleanupFromPod()` method to cleanup Pod Identity token?

pkg/driver/node/credentialprovider/provider_pod.go
```diff
@@ -48,56 +55,68 @@ func (c *Provider) provideFromPod(ctx context.Context, provideCtx ProvideContext
...
+		err := renameio.WriteFile(filepath.Join(provideCtx.WritePath, tokenName), []byte(stsToken.Token), CredentialFilePerm)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "Failed to write service account STS token: %v", err)
+		}
 
-		envprovider.EnvRegion:        region,
-		envprovider.EnvDefaultRegion: defaultRegion,
+		env.Merge(irsaCredentialsEnvironment)
+	} else {
+		klog.Error("Error providing credentials from pod with STS Web Identity provider (IRSA)")
```
yerzhan7, line 105: Don't think we need to log this as error log. Info level should be fine since it's expected if customer only configured Pod Identity credentials.

pkg/driver/node/credentialprovider/provider_pod.go
```diff
@@ -141,7 +160,53 @@ func (c *Provider) findPodServiceAccountRole(ctx context.Context, provideCtx Pro
...
+// podLevelEksPodIdentityServiceAccountTokenName returns service account token name for Pod-level identity with EKS Pod Identity.
+// It escapes from slashes to make this token name path-safe.
+func podLevelEksPodIdentityServiceAccountTokenName(podID string, volumeID string) string {
+	id := escapedVolumeIdentifier(podID, volumeID)
+	return id + "-eks-pod-identity.token"
+}
+
+// createIRSACredentialsEnvironment creates an environment with the environment variables needed for pod-level authentication with IRSA
+func (c *Provider) createIRSACredentialsEnvironment(ctx context.Context, provideCtx ProvideContext) (envprovider.Environment, error) {
+	roleARN, err := c.findPodServiceAccountRole(ctx, provideCtx)
```
yerzhan7, line 177: I think here there is a small issue related to my Pod Sharing change where currently `findPodServiceAccountRole()` will make API call to get ServiceAccount if `ServiceAccountEKSRoleARN` from MountpointS3PodAttachment custom resource is empty string. 

https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/pkg/driver/node/credentialprovider/provider_pod.go#L116-L119



We should not make this API call and instead always return `provideCtx.ServiceAccountEKSRoleARN` if we are using Pod Mounter. And if it's empty string we should return `Missing role annotation` error.

(In Pod Mounter (v2), we've decided to get role ARN from this custom resource instead of making Get `ServiceAccount` API call on Node)

pkg/driver/node/credentialprovider/provider_pod.go
yerzhan7: Seems like EKS Pod Identity Webhook provides `region` and `defaultRegion` variables for both IRSA and Pod Identity: https://github.com/philljie/amazon-eks-pod-identity-webhook/blob/master/pkg/handler/handler.go#L211-L246

I think we should do the same to be consistent.

pkg/driver/node/credentialprovider/provider_pod.go
renanmagagnin: Done, thanks! The handling of custom addresses will be implemented in a follow-up PR

yerzhan7 (dismissed):

unexge (dismissed):

tests/e2e-kubernetes/testsuites/credentials.go
unexge, line 1: There is a top-level comment on `DefineTests` functions in this file describing different tests, we should probably add EKS PI there as well

unexge (approved):

PR #459, opened 2025-05-08
Ensure to not assign new workload to Mountpoint Pod if its annotated with "needs-unmount"

unexge:


cmd/aws-s3-csi-controller/csicontroller/reconciler.go
```diff
@@ -637,7 +741,7 @@ func (r *Reconciler) addNeedsUnmountAnnotation(ctx context.Context, mpPodName st
 	mpPod.Annotations[mppod.AnnotationNeedsUnmount] = "true"
 
 	// Update the pod
-	err = r.Update(ctx, mpPod)
+	err = r.Update(ctx, mpPod) // TODO: This probably needs to be a patch as we might've get a stale Mountpoint Pod.
```
yerzhan7, line 744: +1 

`Patch` is better here

yerzhan7 (approved):

PR #464, opened 2025-05-12
Add IsMountpoint check in PodUnmounter

yerzhan7:
*Issue #, if available:* N/A

*Description of changes:* Currently, there is error message during unmount when there are multiple `HandleMountpointPodUpdate` method invocations. Fix this by first checking if `sourcePath` is mountpoint or not and handle special cases accordingly.

Also, refactor `SourceMountDir` to be dependent on `kubeletpath`

### DIFF
--- a/pkg/driver/node/mounter/pod_unmounter.go
+++ b/pkg/driver/node/mounter/pod_unmounter.go
@@ -21,12 +21,11 @@ const cleanupInterval = 2 * time.Minute
 
 // PodUnmounter handles unmounting of Mountpoint Pods and cleanup of associated resources
 type PodUnmounter struct {
-	nodeID         string
-	mount          *mpmounter.Mounter
-	kubeletPath    string
-	sourceMountDir string
-	podWatcher     *watcher.Watcher
-	credProvider   *credentialprovider.Provider
+	nodeID       string
+	mount        *mpmounter.Mounter
+	kubeletPath  string
+	podWatcher   *watcher.Watcher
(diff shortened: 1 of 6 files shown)


pkg/driver/node/mounter/pod_unmounter.go
```diff
@@ -68,25 +65,82 @@ func (u *PodUnmounter) unmountSourceForPod(mpPod *corev1.Pod) {
...
+	source := filepath.Join(SourceMountDir(u.kubeletPath), mpPod.Name)
+
+	// Check mountpoint status and handle special cases
+	isMountpoint, err := u.checkMountpointAndCleanup(source)
+	if err != nil {
+		klog.Errorf("Error during handling mountpoint check for %s: %v", source, err)
+		return
+	}
+	if !isMountpoint {
+		return
```
unexge, line 77: Should we emit a log here?

unexge (approved):

pkg/driver/node/mounter/pod_unmounter.go
yerzhan7: I originally had it, but then removed as it produced a lot of noise in logs because `HandleMountpointPodUpdate` is invoked around 4-5 times during unmount.

I can add it back with V(5) level? `klog.V(5).Infof()`

unexge: Oh okay, I think it's fine to not add now then

PR #474, opened 2025-05-16
Improvements on Mountpoint Pod cleanup with Pod sharing

unexge:
- [Do not mount plugin-dir in CSI Driver container](https://github.com/awslabs/mountpoint-s3-csi-driver/commit/8af73383d96162b8ed6e64d4be1e7c5e71c8ec0b) 
The plugin directory (`/var/lib/kubelet/plugins/s3.csi.aws.com/` by default) was mounted in the CSI Driver container at `/csi`.
However, the kubelet directory (`/var/lib/kubelet` by default) is also mounted in the CSI Driver at `/var/lib/kubelet` which already includes the plugin dir.
This was okay so far, but with Pod sharing we started mounting Mountpoint instances in the plugin dir, and also bind mounts from the kubelet dir for the workloads.
This was causing Mountpoint mounts to have references to themselves in both the plugin dir and the kubelet dir, and due that when we try to unmount them,
we were getting "resource is busy" errors, as the kernel was thinking there are still references to the mount

- [Exit with success exit code from Mountpoint Pod if it fail to receive mount options and `mount.exit` file is present](https://github.com/awslabs/mountpoint-s3-csi-driver/commit/5b9d26417f034660c1448eebef22c560e9bccbcd) 

- [Make sure to wait until all references are gone before unmounting Mountpoint in PodUnmounter](https://github.com/awslabs/mountpoint-s3-csi-driver/commit/a1885211fdfe7a9ea9dd93422ddbc60f38f4f4ce) 

---

### DIFF
--- a/pkg/driver/node/mounter/pod_unmounter.go
+++ b/pkg/driver/node/mounter/pod_unmounter.go
@@ -1,7 +1,9 @@
 package mounter
 
 import (
+	"context"
 	"errors"
+	"fmt"
 	"io/fs"
 	"os"
 	"path/filepath"
@@ -14,10 +16,19 @@ import (
 	"github.com/awslabs/aws-s3-csi-driver/pkg/util"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
 )
 
-const cleanupInterval = 2 * time.Minute
+const (
+	danglingMountpointCleanupInterval = 2 * time.Minute
+
+	waitUntilMountpointIsUnmountedTimeout  = 30 * time.Second
+	waitUntilMountpointIsUnmountedInterval = 5 * time.Second
+
+	waitUntilMountpointIsUnusedTimeout  = 30 * time.Second
+	waitUntilMountpointIsUnusedInterval = 5 * time.Second
+)
 
 // PodUnmounter handles unmounting of Mountpoint Pods and cleanup of associated resources
 type PodUnmounter struct {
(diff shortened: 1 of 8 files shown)


pkg/driver/node/mounter/pod_unmounter.go
```diff
@@ -230,30 +109,155 @@ func (u *PodUnmounter) CleanupDanglingMounts() error {
...
+	u.cleanUnmount(mpPod)
+}
+
+// cleanUnmount performs a clean unmount for `mpPod`.
+func (u *PodUnmounter) cleanUnmount(mpPod *corev1.Pod) {
+	source := u.mountpointPodSourcePath(mpPod.Name)
+	podPath := u.podPath(string(mpPod.UID))
+
+	// First, write `mount.exit` file to signal a clean exit to Mountpoint Pod, so it exists with zero code.
+	if err := u.writeExitFile(podPath); err != nil {
```
yerzhan7, line 146: Could this fail to write exit file if we already unmounted and deleted MPPod in previous invocation? 

I.e. should we check if `source` is mountpoint before writing exit file?

yerzhan7 (commented):
Thanks for fixing this!

pkg/driver/node/mounter/pod_unmounter.go
unexge: It might fail, but I think it wouldn't be a fatal failure. We already know Mountpoint Pod exists here, and we'd need to signal a clean exit to the Mountpoint Pod even it's not mounted. I also updated `aws-s3-csi-mounter` to check this file even if Mountpoint Pod is not mounted.

pkg/driver/node/mounter/pod_unmounter.go
```diff
@@ -230,30 +109,155 @@ func (u *PodUnmounter) CleanupDanglingMounts() error {
...
+	}
+
+	unlockMountpointPod := lockMountpointPod(mpPod.Name)
+	defer unlockMountpointPod()
+
+	u.cleanUnmount(mpPod)
+}
+
+// cleanUnmount performs a clean unmount for `mpPod`.
+func (u *PodUnmounter) cleanUnmount(mpPod *corev1.Pod) {
```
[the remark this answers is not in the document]
unexge: Added various logs with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/474/commits/a3901d5c41e62b955718b423c82c880ab7393458

pkg/driver/node/mounter/pod_unmounter.go
yerzhan7: > We already know Mountpoint Pod exists here

It might not exist as subsequent invocations wait for MP Pod lock release from previous invocation. Although, the chance of MP Pod not existing right away is very low as it takes take to cleanup. So, I am ok to keeping it as it is for now.

yerzhan7 (approved):

pkg/driver/node/mounter/pod_unmounter.go
unexge: Yeah at the worst case it should just emit an error and not try it again on the next run

PR #478, opened 2025-05-22
Create an `emptyDir` volume in Mountpoint Pod to use as local-cache if enabled

unexge:

### DIFF
--- a/cmd/aws-s3-csi-mounter/csimounter/csimounter.go
+++ b/cmd/aws-s3-csi-mounter/csimounter/csimounter.go
@@ -4,12 +4,14 @@ import (
 	"fmt"
 	"io/fs"
 	"os"
+	"path/filepath"
 
 	"k8s.io/klog/v2"
 
 	"github.com/awslabs/aws-s3-csi-driver/pkg/mountpoint"
 	"github.com/awslabs/aws-s3-csi-driver/pkg/mountpoint/mountoptions"
 	"github.com/awslabs/aws-s3-csi-driver/pkg/mountpoint/runner"
+	"github.com/awslabs/aws-s3-csi-driver/pkg/podmounter/mppod"
 )
 
 var mountErrorFileperm = fs.FileMode(0600) // only owner readable and writeable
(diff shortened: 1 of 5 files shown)


cmd/aws-s3-csi-mounter/csimounter/csimounter.go
```diff
@@ -37,11 +39,16 @@ func Run(options Options) (int, error) {
...
+	localCacheDir := filepath.Join("/", mppod.LocalCacheDirName)
+	_, localCacheEnabledViaMountOptions := mountpointArgs.Remove(mountpoint.ArgCache)
+	localCacheMounted := checkIfDirExists(localCacheDir)
+
+	if localCacheEnabledViaMountOptions && !localCacheMounted {
+		return 0, fmt.Errorf("local cache enabled via mount options but cache folder is not mounted at %q", localCacheDir)
+	}
+
+	if localCacheMounted {
+		mountpointArgs.Set(mountpoint.ArgCache, localCacheDir)
```
yerzhan7, line 51: np: Not sure I like to modify `mountpointArgs` in `csimounter` in addition to modifying them in `node`. Is there a reason we can't adjust `--cache` argument on the Node side?

yerzhan7 (approved):

cmd/aws-s3-csi-mounter/csimounter/csimounter.go
unexge: I think we don't have this information in the node component, maybe we can move that logic from the node component to here as well?

yerzhan7: By "this information" you mean `--cache /local-cache` flag? Node should have this in mountOptions from PV, no?

unexge: It's the fact that `local-cache` volume is mounted to Mountpoint Pod. We can also understand by looking into mount options and volume attributes to understand if caching is enabled, but that would mean we need to duplicate the same logic in both controller and node. That's why it feels easier to make that decision on Mountpoint Pod by looking whether the cache volume is mounted at `/local-cache` or not.

cmd/aws-s3-csi-mounter/csimounter/csimounter.go
```diff
@@ -37,11 +39,16 @@ func Run(options Options) (int, error) {
...
-	// ideally we should create a volume (`emptyDir` by default) in the Mountpoint Pod and use that.
-	mountpointArgs, err := createCacheDir(mountpointArgs)
-	if err != nil {
-		return 0, fmt.Errorf("failed to create cache dir: %w", err)
+	localCacheDir := filepath.Join("/", mppod.LocalCacheDirName)
+	_, localCacheEnabledViaMountOptions := mountpointArgs.Remove(mountpoint.ArgCache)
+	localCacheMounted := checkIfDirExists(localCacheDir)
+
+	if localCacheEnabledViaMountOptions && !localCacheMounted {
+		return 0, fmt.Errorf("local cache enabled via mount options but cache folder is not mounted at %q", localCacheDir)
```
yerzhan7, line 47: Shouldn't we return non-zero exit code in this case?

unexge: We basically ignore exit-code if this function returns an error (we call `klog.Fatal` which exists with `255`). By convention, you return zero values in case of an error, and they ignored in the case of a non-nil error

PR #488, opened 2025-05-29
Use Mountpoint Pod UID as cache key with `PodMounter`

unexge:

### DIFF
--- a/pkg/driver/node/credentialprovider/provider.go
+++ b/pkg/driver/node/credentialprovider/provider.go
@@ -186,20 +186,32 @@ func New(client k8sv1.CoreV1Interface, regionFromIMDS func() (string, error)) *P
 // Depending on the configuration, it either returns driver-level or pod-level credentials.
 func (c *Provider) Provide(ctx context.Context, provideCtx ProvideContext) (envprovider.Environment, AuthenticationSource, error) {
 	if provideCtx.MountKind == MountKindUnspecified {
-		return nil, "", fmt.Errorf("MountKind must be specified on credential ProvideContext struct.")
+		return nil, AuthenticationSourceUnspecified, errors.New("Missing `MountKind` from `ProvideContext`")
+	}
+
+	if provideCtx.IsPodMountpoint() && provideCtx.MountpointPodID == "" {
+		return nil, AuthenticationSourceUnspecified, errors.New("Missing `MountpointPodID` from `ProvideContext`")
 	}
 
 	authenticationSource := provideCtx.AuthenticationSource
+	var env envprovider.Environment
+	var err error
 	switch authenticationSource {
 	case AuthenticationSourcePod:
-		env, err := c.provideFromPod(ctx, provideCtx)
-		return env, AuthenticationSourcePod, err
+		env, err = c.provideFromPod(ctx, provideCtx)
 	case AuthenticationSourceUnspecified, AuthenticationSourceDriver:
-		env, err := c.provideFromDriver(provideCtx)
-		return env, AuthenticationSourceDriver, err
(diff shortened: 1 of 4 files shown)


pkg/driver/node/credentialprovider/provider.go
```diff
@@ -186,20 +186,32 @@ func New(client k8sv1.CoreV1Interface, regionFromIMDS func() (string, error)) *P
...
-		return env, AuthenticationSourceDriver, err
+		authenticationSource = AuthenticationSourceDriver
+		env, err = c.provideFromDriver(provideCtx)
 	default:
 		return nil, AuthenticationSourceUnspecified, fmt.Errorf("unknown `authenticationSource`: %s, only `driver` (default option if not specified) and `pod` supported", authenticationSource)
 	}
+
+	if provideCtx.IsPodMountpoint() {
+		env.Set(envprovider.EnvMountpointCacheKey, provideCtx.MountpointPodID)
+	}
```
yerzhan7, line 211: Should we set cache key somewhere else? E.g. in `pod_mounter.go`?

It's kinda not related to credentials at all.

unexge: Yeah I also thought the same, but since we set it for PLI in `provider_pod.go` I wanted to put this one into same place as well to make it consistent. I think we can move it to `PodMounter` once we remove the systemd fallback here.

yerzhan7: I think we can move it to `PodMounter` today since this code path and systemd path are already not overlapping with each other, but it's up to you.

Can we have at least `TODO` comment to mention that this needs to be moved to `PodMounter`?

yerzhan7: Also, this means we are setting cache key always (not only when cache flag is present) which I think is unnecessary.

unexge: I can move it to `PodMounter` but we don't know whether case is enabled or not in `PodMounter`, so either we'd need to move it into `aws-s3-csi-mounter` or just set it always

yerzhan7: Oh, yeah, I forgot that decision about cache is now done in `aws-s3-csi-mounter`. Maybe that's another reason to move decision about cache to `PodMounter`?

This way we set all environmental variables and mount options in single place (`PodMounter`) and then `aws-s3-csi-mounter` would simply pass these 'as is' to mount-s3 process? I think this would be much cleaner/simpler.

Currently, the logic about cache is spread out in multiple places (`CredentialsProvider`, `aws-s3-csi-mounter`) which makes it difficult to follow.

unexge: Yeah that's true it's not nice, but it's also not that trivial to be honest. The controller decides whether cache should be enabled or not, and then the mounter knows that by looking at whether cache volume is mounted or not. The node doesn't have any information at the moment.

I can just move this cache key to the mounter I think, that would be the best way for now. What do you think?

yerzhan7: > The controller decides whether cache should be enabled or not

I think we can duplicate controller logic about cache into node. Node should have all the information as controller (from `volume_context` and `mount_flags`). Are there any disadvantages of duplicating this logic? We can even let both controller and node use the same helper/util method `isCacheEnabled(...)` to avoid code duplication.

> I can just move this cache key to the mounter I think, that would be the best way for now. What do you think?

This can also work, thanks.

unexge:
Closing in favor of https://github.com/awslabs/mountpoint-s3-csi-driver/pull/496

PR #492, opened 2025-06-03
e2e tests: retry on conflict when updating service account

yerzhan7:
*Issue #, if available:* N/A

*Description of changes:* Some credentials e2e tests fail on OpenShift clusters.

Need to retry on conflict because OpenShift automatically adds some annotations and labels to service accounts after creation.


tests/e2e-kubernetes/testsuites/credentials.go
```diff
@@ -1243,11 +1244,25 @@ func deleteCredentialSecret(ctx context.Context, f *framework.Framework) error {
...
+		if err != nil {
+			return err
+		}
+
+		if sa.Annotations == nil {
+			sa.Annotations = make(map[string]string)
+		}
+		sa.Annotations[roleARNAnnotation] = roleARN
+
+		_, err = client.Update(ctx, sa, metav1.UpdateOptions{})
```
unexge: A `Patch` probably better here, but I'm also okay with this one.

tests/e2e-kubernetes/testsuites/credentials.go
```diff
@@ -1256,25 +1271,12 @@ func annotateServiceAccountWithRole(sa *v1.ServiceAccount, roleARN string) {
...
-				return nil
-			}
-			return err
-		}
-
 		framework.Logf("Restoring ServiceAccount %s's role", sa.Name)
-		annotateServiceAccountWithRole(sa, originalRoleARN)
-		_, err = client.Update(ctx, sa, metav1.UpdateOptions{})
-		return err
+		return annotateServiceAccountWithRole(ctx, f, sa, originalRoleARN)
```
unexge: I think we should probably keep 
```go
if apierrors.IsNotFound(err) {
     return nil
}
```
here just in case SA is already deleted

yerzhan7: Fixed

tests/e2e-kubernetes/testsuites/credentials.go
yerzhan7: Going to keep as is. Simple patching is not working in my testing.

unexge (dismissed):

PR #494, opened 2025-06-03
Remove `SystemdMounter`

unexge:
This PR is a preparation for v2 branch. It removes `SystemdMounter` and makes `PodMounter` the default mounter. And also enables `node.{tolerateAllTaints, podInfoOnMount}` by default in our Helm chart.

---


.github/workflows/e2e-tests.yaml
```diff
@@ -112,87 +112,28 @@ jobs:
...
         run: |
           tests/e2e-kubernetes/scripts/run.sh
-      - name: Run E2E Tests (SystemdMounter)
-        if: matrix.selinux-mode != 'enforcing'
+      - name: Run E2E Tests
         env:
           ACTION: "run_tests"
         run: |
           tests/e2e-kubernetes/scripts/run.sh
-      - name: Run Performance Tests
```
yerzhan7, line 130: No longer need to run Performance Tests?

Dockerfile
```diff
@@ -60,9 +60,7 @@ ENV MOUNTPOINT_BIN_DIR=/mountpoint-s3/bin
 
 # Copy Mountpoint binary
 COPY --from=mp_builder /mountpoint-s3 /mountpoint-s3
-# TODO: These won't be necessary with containerization.
 COPY --from=mp_builder /lib64/libfuse.so.2 /mountpoint-s3/bin/
```
yerzhan7, line 63: Just checking: This one is still needed in v2?

yerzhan7 (approved):

Dockerfile
unexge: I tried removing it, but it failed, my assumption is that Mountpoint is linked with `libfuse2` and the base image only contains `libfuse3` or doesn't contain any `libfuse`. So, we need to put a compatible `libfuse` there.

.github/workflows/e2e-tests.yaml
unexge: We currently don't utilize performance tests, and they also emit some warnings https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/15398283220/job/43324461325#step:13:44. We probably need to take a look and fix the problems. But again we don't utilize them in this repo, there are some better up-to-date benchmarks in the Mountpoint repo.

yerzhan7: Ok, cool.

np: you can also remove benchmark related environmental variables above

unexge: Will do in a follow-up PR

PR #496, opened 2025-06-04
Support `ephemeral` volume type as local-cache instead of `persistentVolumeClaim` in Mountpoint Pod

unexge:
This PR replaces `persistentVolumeClaim` cache type with `ephemeral`. If configured, the Mountpoint Pod will use a [generic ephemeral volume](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes) to ensure each Mountpoint Pod gets its own cache volume. 

Given the following configuration:
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: mp-cache-ebs-sc
provisioner: ebs.csi.aws.com
reclaimPolicy: Delete
volumeBindingMode: WaitForFirstConsumer
parameters:
  type: io2
  iopsPerGB: "256000"
  blockExpress: "true"
---
apiVersion: v1
kind: PersistentVolume
metadata:
  name: s3-pv
spec:
  # ...
  csi:
    driver: s3.csi.aws.com
    volumeHandle: s3-pv
    volumeAttributes:
      bucketName: amzn-s3-demo-bucket
      cache: ephemeral
      cacheEphemeralStorageClassName: mp-cache-ebs-sc # required
      cacheEphemeralStorageResourceRequest: 4Gi       # required
```

The spawned Mountpoint Pod will have the following PVC definition:
```yaml
Name:          mp-rpg92-local-cache
Namespace:     mount-s3
StorageClass:  mp-cache-ebs-sc
Status:        Bound
Volume:        pvc-2450cff6-d694-4e5d-a968-c1373068c9aa
Labels:        type=mountpoint-csi-driver-local-cache
Annotations:   pv.kubernetes.io/bind-completed: yes
               # ...
Finalizers:    [kubernetes.io/pvc-protection]
Capacity:      4Gi
Access Modes:  RWO 
VolumeMode:    Filesystem
Used By:       mp-rpg92
```

The PV will be dynamically provisioned by the EBS CSI Driver. The created PVC and PV will be automatically deleted (due to `reclaimPolicy: Delete` in SC) alongside Mountpoint Pod when its no longer in use.

---

### DIFF
--- a/pkg/podmounter/mppod/creator.go
+++ b/pkg/podmounter/mppod/creator.go
@@ -197,7 +197,7 @@ func (c *Creator) configureLocalCache(mpPod *corev1.Pod, mpContainer *corev1.Con
 
 	if cacheEnabledViaOptions {
 		if cacheType != "" {
-			return errors.New("Cache configured with both `mountOptions` and `volumeAttributes`, please remove the deprecated cache configuragion in `mountOptions`")
+			return errors.New("Cache configured with both `mountOptions` and `volumeAttributes`, please remove the deprecated cache configuration in `mountOptions`")
 		}
 
 		// TODO: Create and link `CACHING.md`.
@@ -277,7 +277,7 @@ func (c *Creator) createCacheVolumeSourceForEphemeral(volumeAttributes map[strin
 			VolumeClaimTemplate: &corev1.PersistentVolumeClaimTemplate{
 				ObjectMeta: metav1.ObjectMeta{
 					Labels: map[string]string{
-						"type": "mountpoint-csi-driver-local-cache",
+						"s3.csi.aws.com/type": "local-ephemeral-cache",
 					},
 				},
 				Spec: corev1.PersistentVolumeClaimSpec{
--- a/pkg/podmounter/mppod/creator.go
+++ b/pkg/podmounter/mppod/creator.go
@@ -45,7 +45,7 @@ const (
 	// cacheTypeEmptyDir creates a cache unique to the Mountpoint Pod by using `emptyDir` volume type.
 	// See https://kubernetes.io/docs/concepts/storage/volumes/#emptydir.
 	cacheTypeEmptyDir = "emptyDir"
-	// cacheTypeEphemeral creates a generic ephemeral volume unique to the Mountpoint Pod (with `ReadWriteOncePod` access mode) by using `ephemeral` volume type.
+	// cacheTypeEphemeral creates a generic ephemeral volume unique to the Mountpoint Pod by using `ephemeral` volume type.
 	// See https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes.
 	cacheTypeEphemeral = "ephemeral"
 )
@@ -281,10 +281,7 @@ func (c *Creator) createCacheVolumeSourceForEphemeral(volumeAttributes map[strin
 					},
 				},
 				Spec: corev1.PersistentVolumeClaimSpec{
-					AccessModes: []corev1.PersistentVolumeAccessMode{
-						// Make sure the Mountpoint Pods is the exclusive owner of this volume
-						corev1.ReadWriteOncePod,
-					},
+					AccessModes:      []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce},
 					StorageClassName: &storageClassName,
 					VolumeMode:       ptr.To(corev1.PersistentVolumeFilesystem),
 					Resources: corev1.VolumeResourceRequirements{
--- a/pkg/podmounter/mppod/creator.go
+++ b/pkg/podmounter/mppod/creator.go
@@ -42,8 +42,12 @@ const (
 )
 
 const (
-	cacheTypeEmptyDir              = "emptyDir"
-	cacheTypePersistentVolumeClaim = "persistentVolumeClaim"
+	// cacheTypeEmptyDir creates a cache unique to the Mountpoint Pod by using `emptyDir` volume type.
+	// See https://kubernetes.io/docs/concepts/storage/volumes/#emptydir.
+	cacheTypeEmptyDir = "emptyDir"
+	// cacheTypeEphemeral creates a generic ephemeral volume unique to the Mountpoint Pod (with `ReadWriteOncePod` access mode) by using `ephemeral` volume type.
+	// See https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes.
+	cacheTypeEphemeral = "ephemeral"
 )
 
 const CommunicationDirSizeLimit = 10 * 1024 * 1024 // 10MB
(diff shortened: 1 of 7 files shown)


yerzhan7:
Since we no longer support sharing local cache in v2 (at least at the moment), could we please set `EnvMountpointCacheKey` only if `if provideCtx.IsSystemDMountpoint()` in https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/pkg/driver/node/credentialprovider/provider_pod.go#L74?

pkg/podmounter/mppod/creator.go
```diff
@@ -251,16 +255,47 @@ func (c *Creator) createCacheVolumeSourceForEmptyDir(volumeAttributes map[string
...
+			VolumeClaimTemplate: &corev1.PersistentVolumeClaimTemplate{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						"type": "mountpoint-csi-driver-local-cache",
+					},
+				},
+				Spec: corev1.PersistentVolumeClaimSpec{
+					AccessModes: []corev1.PersistentVolumeAccessMode{
+						// Make sure the Mountpoint Pods is the exclusive owner of this volume
+						corev1.ReadWriteOncePod,
```
yerzhan7: I think `ReadWriteOncePod` is unnecessary as each pod gets its own unique ephemeral volume instance as per https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes

I am just concerned in case customers want to use different CSI driver that does not support `ReadWriteOncePod` access mode.

unexge: So the PVC created by this mechanism is just a regular PVC, and it can be mounted by a different Pod technically. The `ReadWriteOncePod` prevents that. It's probably not a big issue if we don't specify it as we ensure each Mountpoint Pods gets its own PVC.

EBS CSI Driver supports it, and it's the main use-case we're targeting with this feature. I can look it up what's the support looks like among other CSI Drivers to have some data that can help us to decide.

unexge: Well. The CI failures on Kubernetes < 1.32 seems to be related to https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1982. Maybe we can drop `ReadWriteOncePod` 😄

unexge: Changed to `ReadWriteOnce` with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/496/commits/166f5bf73d04785782d66286a5bcd35513256d7b to see if it will fix the CI.

pkg/podmounter/mppod/creator.go
```diff
@@ -251,16 +255,44 @@ func (c *Creator) createCacheVolumeSourceForEmptyDir(volumeAttributes map[string
 	return corev1.VolumeSource{EmptyDir: emptyDir}, nil
 }
 
-// createCacheVolumeSourceForEmptyDir creates an `persistentVolumeClaim` volume source to use as local-cache.
-func (c *Creator) createCacheVolumeSourceForPVC(volumeAttributes map[string]string) (corev1.VolumeSource, error) {
-	claimName := volumeAttributes[volumecontext.CachePersistentVolumeClaimName]
-	if claimName == "" {
-		return corev1.VolumeSource{}, fmt.Errorf("%q must be provided with %q cache type", volumecontext.CachePersistentVolumeClaimName, cacheTypePersistentVolumeClaim)
+// createCacheVolumeSourceForEphemeral creates an `ephemeral` volume source to use as local-cache.
```
yerzhan7, line 258: np: Just noticed a typo on line 200 above: `configuragion in mountOptions")` -> `configuration`

pkg/podmounter/mppod/creator_test.go
```diff
@@ -286,13 +288,69 @@ func createAndVerifyPod(t *testing.T, clusterVariant cluster.Variant, expectedRu
...
+								volumecontext.CacheEphemeralStorageResourceRequest: "1Gi",
+							},
+						},
+					},
+				},
+			})
+			assert.Equals(t, cmpopts.AnyError, err)
+		})
+
+		t.Run("With ephemeral cache but missing resource request", func(t *testing.T) {
```
yerzhan7, line 333: What is the reason we are making resource request limit a required parameter? 

It is currently optional for `emptyDir` case, why `ephemeral` is different?

pkg/podmounter/mppod/creator.go
```diff
@@ -251,16 +255,44 @@ func (c *Creator) createCacheVolumeSourceForEmptyDir(volumeAttributes map[string
...
+	if err != nil {
+		return corev1.VolumeSource{}, failedToParseQuantityError(err, volumecontext.CacheEphemeralStorageResourceRequest, storageResourceRequestStr)
+	}
+
+	return corev1.VolumeSource{
+		Ephemeral: &corev1.EphemeralVolumeSource{
+			VolumeClaimTemplate: &corev1.PersistentVolumeClaimTemplate{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						"type": "mountpoint-csi-driver-local-cache",
```
yerzhan7: Is this label used for anything? We usually set some prefix like this for our labels/annotations `s3.csi.aws.com/`?

pkg/podmounter/mppod/creator.go
yerzhan7: > So the PVC created by this mechanism is just a regular PVC, and it can be mounted by a different Pod technically. 

According to the docs CSI Drivers should not mount the same volume to several Pods. So, it should be safe to use `ReadWriteOnce` for now.

pkg/podmounter/mppod/creator_test.go
unexge: It's because `ephemeral` creates PVC under the hood which requires storage requests, for example:

```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: s3-cache-pvc
spec:
  accessModes:
    - ReadWriteMany
  storageClassName: "gp2"
```

this fails with:
```
The PersistentVolumeClaim "s3-cache-pvc" is invalid: spec.resources[storage]: Required value
```

pkg/podmounter/mppod/creator.go
unexge: It's just for tracking purposes really, might be handy to list volumes used as cache by the CSI Driver. But you're right about the prefix, I'll update it to: `s3.csi.aws.com/type: local-ephemeral-cache`

pkg/podmounter/mppod/creator.go
unexge: Fixed with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/496/commits/b5216c7b894ba5b2c52f65a2a9e4e7d519eeb501

pkg/podmounter/mppod/creator.go
unexge: Updated with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/496/commits/b5216c7b894ba5b2c52f65a2a9e4e7d519eeb501

PR #500, opened 2025-06-05
Update docs for Outposts

muddyfish:
*Issue #, if available:* N/A

*Description of changes:*
Updates the text in the configuration guide for outposts and static provisioning.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -5,10 +5,13 @@ Mountpoint specific configuration.
 
 ## Static Provisioning
 
-The driver only supports Static Provisioning as of today, and you need an existing S3 Bucket to use.
+The CSI driver supports only static provisioning for an existing S3 bucket.
+Supported bucket types include general purpose, directory, and Outposts buckets.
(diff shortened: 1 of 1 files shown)


docs/CONFIGURATION.md
```diff
@@ -5,10 +5,13 @@ Mountpoint specific configuration.
 
 ## Static Provisioning
 
-The driver only supports Static Provisioning as of today, and you need an existing S3 Bucket to use.
+The CSI driver supports only static provisioning for an existing S3 bucket.
+Supported bucket types include general purpose, directory, and Outposts buckets.
+In the 'bucketName' field, provide the full S3 bucket name.
+For Outposts buckets, only access point ARN or alias is supported.
```
yerzhan7, line 11: np: What about Object Lambda and Access Points?

yerzhan7 (approved):

docs/CONFIGURATION.md
muddyfish: I don't think OL & Access Points are types of bucket right? Though we don't mention them at all in the CSI docs

PR #502, opened 2025-06-05
Add `CACHING.md` to explain possible caching configurations the CSI Driver provides

unexge:
[Rendered](https://github.com/unexge/mountpoint-s3-csi-driver/blob/8ce319057d26d7b6092856a8f20820375a0ed549/docs/CACHING.md)

---


docs/CACHING.md
```diff
@@ -0,0 +1,353 @@
+# Caching Configuration of Mountpoint for Amazon S3 CSI Driver
+
+Mountpoint supports caching file system metadata and object content to reduce cost and improve performance for repeated reads to the same file. The CSI Driver allows you to configure caching of Mountpoint in your PersistentVolume (PV) definition. See [Mountpoint's caching configuration]((https://github.com/awslabs/mountpoint-s3/blob/main/doc/CONFIGURATION.md#caching-configuration)) for more details about caching.
```
yerzhan7: np: `Mountpoint` - not clear what you mean by this. How about `[Mountpoint for Amazon S3](https://github.com/awslabs/mountpoint-s3)`?

docs/CACHING.md
```diff
@@ -0,0 +1,353 @@
+# Caching Configuration of Mountpoint for Amazon S3 CSI Driver
```
yerzhan7, line 1: We already have `FEATURES.md` file [here](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/v2/docs/FEATURES.md) and caching seems like a feature. Should we move it into that file or extract Pod Sharing into separate `MOUNTPOINT_POD_SHARING.md` doc? I think I prefer the latter.

docs/CACHING.md
```diff
@@ -0,0 +1,353 @@
...
+          EOF
+      - udevadm control --reload && udevadm trigger
+```
+
+The `i3.8xlarge` instance type provides four NVMe instance storage disks. After applying your changes using `eksctl`, you can install the example EKS NVMe manifest:
+```bash
+$ kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/sig-storage-local-static-provisioner/refs/heads/master/helm/generated_examples/eks-nvme-ssd.yaml
+```
+
+This should create a StorageClass named `nvme-ssd`, and four PVs for each local NVMe instance storage disk attached to the instance:
```
yerzhan7: > `four PVs for each local NVMe instance`

Not super clear to me. So, this means we will have 4 PVs for each EC2 Instance? What if customer has workload across many nodes? Could you please clarify many node scenario in the doc?

docs/CACHING.md
```diff
@@ -0,0 +1,353 @@
...
+
+```yaml
+apiVersion: eksctl.io/v1alpha5
+kind: ClusterConfig
+metadata:
+  name: cluster-with-storage
+  region: eu-central-1
+managedNodeGroups:
+  - name: storage-nvme
+    desiredCapacity: 1
```
yerzhan7: I would prefer to have working example for more than 1 EC2 Instance (e.g. 2) as it's far more popular use-case.

docs/CACHING.md
```diff
@@ -0,0 +1,353 @@
...
+  name: s3-pv
+spec:
+  # ...
+  csi:
+    driver: s3.csi.aws.com
+    # ...
+    volumeAttributes:
+      bucketName: amzn-s3-demo-bucket
+      cache: ephemeral
+      cacheEphemeralStorageClassName: nvme-ssd # required
```
yerzhan7: np: `nvme-ssd` -> `gp2` 

At this part of the doc it's not clear where `nvme-ssd` StorageClass is created. Suggesting to use default pre-installed EKS StorageClass `gp2` here or move `nvme-ssd` StorageClass definition before PV definition.

docs/CACHING.md
```diff
@@ -0,0 +1,353 @@
...
+Annotations:   pv.kubernetes.io/bind-completed: yes
+               pv.kubernetes.io/bound-by-controller: yes
+Finalizers:    [kubernetes.io/pvc-protection]
+Capacity:      1769Gi
+Access Modes:  RWO
+VolumeMode:    Filesystem
+Used By:       mp-xt6c4
+```
+
+Note that if there is no local NVMe available in the scheduled node, Mountpoint Pod would fail to schedule and your workload would hang in `Pending` state.
```
yerzhan7: > `would hang in `Pending` state`

np: With what error message?

docs/CACHING.md
```diff
@@ -0,0 +1,353 @@
...
+  mountOptions:
+    - cache /cache/folder/on/host
+  csi:
+    driver: s3.csi.aws.com
+    # ...
+    volumeAttributes:
+      bucketName: amzn-s3-demo-bucket
+```
+
+will translate the following automatically by the CSI Driver v2:
```
yerzhan7: This kinda implies that S3 CSI Driver will actually modify customer PV automatically which is not the case. We can just say that cache folder parameter from MountOptions will be ignored and we will use `emptyDir` for cache instead.

docs/CACHING.md
unexge: I think it makes sense to move Pod sharing into its own file. I feel like if we put all features into one file it would be very big. I'll do this as a separate PR.

docs/CACHING.md
unexge: I'll address this with the next comment by increasing the example to have two nodes. But yeah each instance will have 4 PVs due to their type `i3.8xlarge` provides 4 local NVMe.

Issue #504, opened 2025-06-06
Mountpoint for Amazon S3 CSI Driver v2


unexge:
> [!IMPORTANT] 
> Mountpoint for Amazon S3 CSI Driver v2 is available as Helm chart, EKS add-on and Kustomization manifest! See [Upgrading Mountpoint for Amazon S3 CSI Driver from v1 to v2](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/docs/UPGRADING_TO_V2.md) for more details on upgrading to v2, and [Installing Mountpoint for Amazon S3 CSI Driver](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/docs/INSTALL.md) for a fresh installation!

---

We’re planning to release version 2 of the Mountpoint for Amazon S3 CSI driver in the upcoming months. Today, we are announcing a beta release of v2 to enable wider community testing. Please note that this beta release is intended for experimental use and __its adoption in production environments is not recommended__. This Beta release is under active development and has not completed testing and review for performance, reliability, and security. Please refer to [Unstable Releases](https://github.com/awslabs/mountpoint-s3-csi-driver?tab=readme-ov-file#unstable-releases) for more information. Any testing is appreciated, however, we are especially interested in getting your feedback on how the beta CSI v2.0 driver performs with large scale workloads. The v2 beta should be tested on a fresh cluster.

With version 2 of the Mountpoint for Amazon S3 CSI driver, you can use the CSI driver on Security-Enhanced Linux (SELinux) enabled environments like [Red Hat OpenShift Service on AWS (ROSA)](https://aws.amazon.com/rosa/), share existing Mountpoint instances between multiple workload pods with a common credential on the same node to further optimize performance, caching and resource utilization, automating cache folder creation for you when you configure caching.

To deliver these improvements, v2 introduces breaking changes because your Mountpoint processes will now spawn inside dedicated and unprivileged Kubernetes pods (see [Run Mountpoint inside container](https://github.com/awslabs/mountpoint-s3-csi-driver/issues/279) and the linked issues for more detail) and not on the host unlike v1.x.

## New features with Mountpoint for Amazon S3 CSI Driver v2

[...]

PR #507, opened 2025-06-09
Doc updates about caching and MP Pod resource limits

yerzhan7:
*Issue #, if available:* https://github.com/awslabs/mountpoint-s3-csi-driver/issues/504

*Description of changes:*
- Fix caching example as it no longer requires --cache flag
- In configuration doc have example that shows all options for volume attributes including MP Pod resource limits which was not documented


docs/CONFIGURATION.md
```diff
@@ -16,26 +16,60 @@ kind: PersistentVolume
...
+    volumeHandle: s3-csi-driver-volume  # Must be unique
     volumeAttributes:
-      bucketName: amzn-s3-demo-bucket # Replace with your bucket name
+      # ----- BUCKET CONFIGURATION -----
+      bucketName: amzn-s3-demo-bucket   # Required: S3 bucket name or ARN (for Outpost)
+
+      # ----- AUTHENTICATION CONFIGURATION -----
+      # See more details in https://github.com/awslabs/mountpoint-s3-csi-driver/blob/v2/docs/CONFIGURATION.md#aws-credentials
+      authenticationSource: driver      # Optional: Authentication source [driver (default) | pod]
+      stsRegion: us-east-1              # Optional: Region for AWS STS endpoint when using pod-level identity with IRSA
```
renanmagagnin, line 39: We are using driver-level and mentioning `stsRegion` for pod-level, should we change this to avoid confusion?

examples/kubernetes/static_provisioning/caching.yaml
```diff
@@ -12,15 +12,17 @@ spec:
     namespace: default # Namespace is required even though it's in "default" namespace.
     name: s3-pvc # Name of your PVC
   mountOptions:
-    - cache /tmp/s3-pv1-cache # specify cache directory, relative to root host filesystem
-                              # the directory must be unique per mount on a host
-    - metadata-ttl 3 # 3 second time to live
+    # See more details in https://github.com/awslabs/mountpoint-s3/blob/main/doc/CONFIGURATION.md#caching-configuration
+    - metadata-ttl 3 # 3 second time to live for metadata cache
```
renanmagagnin, line 16: np: 3-second ttl or 3 seconds ttl

docs/CONFIGURATION.md
yerzhan7: I think it's fine as we already mention it's optional. Plus, I wanted to mention driver because this is the default.

Thanks for reviewing!

PR #509, opened 2025-06-13
[DRAFT] Testing rosa ci

yerzhan7:
*Issue #, if available:* N/A

*Description of changes:* Testing CI


.github/workflows/e2e-test-trusted.yaml
```diff
@@ -14,7 +14,7 @@ permissions:
 jobs:
   e2e:
     name: E2E Tests
-    if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.id == github.event.pull_request.base.repo.id }}
```
yerzhan7, line 17: TODO: Revert this

tests/e2e-kubernetes/scripts/helm.sh
```diff
@@ -19,8 +19,19 @@ function helm_uninstall_driver() {
   KUBECTL_BIN=${2}
   RELEASE_NAME=${3}
   KUBECONFIG=${4}
+  CLUSTER_TYPE=${5}
+
   if driver_installed ${HELM_BIN} ${RELEASE_NAME} ${KUBECONFIG}; then
-    $HELM_BIN uninstall $RELEASE_NAME --namespace kube-system --kubeconfig $KUBECONFIG
+    if [[ "${CLUSTER_TYPE}" == "openshift" ]]; then
+      echo "OpenShift cluster detected - using graceful Helm uninstall as ClusterRoleBindings cannot be deleted due to admission webhooks."
```
vladem: how does it work now, do we keep those ClusterRoleBindings in-between tests? any other resources that deletion skips?

tests/e2e-kubernetes/scripts/run.sh
```diff
@@ -39,11 +40,11 @@ K8S_VERSION_EKSCTL=${K8S_VERSION_EKSCTL:-${K8S_VERSION%.*}}
 
 # We need to ensure that we're using all testing matrix variables in the cluster name
 # because they all run in parallel and conflicting name would break other tests.
-CLUSTER_NAME="s3-csi-cluster-${CLUSTER_TYPE}-${AMI_FAMILY,,}-${ARCH}"
-
-if [[ "${CLUSTER_TYPE}" == "eksctl" ]]; then
+if [[ "${CLUSTER_TYPE}" == "openshift" ]]; then
+    CLUSTER_NAME=${CLUSTER_NAME:-"s3-csi-rosa"}
```
vladem, line 44: just acknowledging that we're testing on one version of openshift (when do we update it? use latest?) and nodes have x86 architecture (no arm testing), I think that's ok for a start

.github/workflows/e2e-rosa-tests.yaml
```diff
@@ -0,0 +1,153 @@
...
+  BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
+  BENCHMARK_BUCKET: "s3://${{ vars.BENCHMARK_BUCKET }}"
+  TAG_UNTESTED: "untested_${{ inputs.ref }}"
+  TAG_PASSED: "test_passed_${{ inputs.ref }}"
+  CLUSTER_TYPE: "openshift"
+  CLUSTER_NAME: "s3-csi-${{ inputs.environment }}"
+jobs:
+  build:
+    runs-on: ubuntu-22.04 # FIXME - https://github.com/actions/runner-images/issues/11471
+    environment: ${{ inputs.environment }}
```
vladem, line 28: what's the purpose of trusted env in the context of openshift?

.github/workflows/e2e-rosa-tests.yaml
```diff
@@ -0,0 +1,153 @@
...
+    inputs:
+      environment:
+        required: true
+        type: string
+      ref:
+        required: true
+        type: string
+
+concurrency:
+  group: e2e-cluster-${{ inputs.environment }}
```
vladem, line 14: is it possible to use a separate concurrency group for rosa tests? I bet otherwise rosa tests->eks tests will run sequentially?

.github/workflows/e2e-rosa-tests.yaml
```diff
@@ -0,0 +1,153 @@
...
+        run: tests/e2e-kubernetes/scripts/run.sh
+      - name: Install OpenShift CLI
+        uses: redhat-actions/openshift-tools-installer@v1
+        with:
+          oc: "latest"
+      - name: Get ROSA Token
+        uses: aws-actions/aws-secretsmanager-get-secrets@v2
+        with:
+          secret-ids: |
+            RosaToken
```
vladem, line 99: we need to document somewhere (not on gh) how do we initialize this, TERRAFORM_BUCKET and other variables

.github/workflows/e2e-test-rosa-untrusted.yaml
```diff
@@ -0,0 +1,27 @@
...
+  pull_request_target:
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  approval:
+    name: Approval Gate
+    environment: "approval-gate"
```
vladem, line 13: just acknowledging approval gate for invocations from forks

.github/workflows/e2e-rosa-tests.yaml
```diff
@@ -0,0 +1,153 @@
...
+        working-directory: tests/e2e-kubernetes/rosa/
+        run: terraform validate
+      - name: Create/Update ROSA cluster
+        working-directory: tests/e2e-kubernetes/rosa/
+        env:
+          TF_VAR_aws_region: "${{ vars.AWS_REGION }}"
+          TF_VAR_cluster_name: "s3-csi-${{ inputs.environment }}"
+          TF_VAR_rhcs_token: "${{ env.ROSATOKEN }}"
+          TF_VAR_billing_account_id: ${{ secrets.ROSA_BILLING_ACCOUNT_ID }}
+        run: terraform apply --auto-approve
```
vladem, line 163: consider re-creating cluster once in a while to mitigate the risk of dangling resources piling up (e.g. lost pods?)

.github/workflows/e2e-rosa-tests.yaml
```diff
@@ -0,0 +1,153 @@
...
+      - name: Get OpenShift credentials
+        uses: aws-actions/aws-secretsmanager-get-secrets@v2
+        with:
+          secret-ids: |
+            OPENSHIFT_CREDS,s3-csi-${{ inputs.environment }}-openshift-credentials
```
vladem, line 168: is it possible to derive short-lived credentials (e.g. with expiration 2hr) from long lived ones?

tests/e2e-kubernetes/testsuites/credentials.go
```diff
@@ -306,14 +309,26 @@ func (t *s3CSICredentialsTestSuite) DefineTests(driver storageframework.TestDriv
...
+
 		cleanClusterWideResources := func(ctx context.Context) {
 			// Since we're using cluster-wide resources and we're running multiple tests in the same cluster,
 			// we need to clean up all credential related resources before each test to ensure we've a
 			// clean starting point in each test.
 			By("Cleaning up cluster-wide resources")
 
 			sa := csiDriverServiceAccount(ctx, f)
-			overrideServiceAccountRole(ctx, f, sa, "")
+			overrideServiceAccountRole(ctx, f, sa, driverBaselineIRSARoleARN)
```
vladem, line 332: > // The returned function restored Service Account's EKS Role ARN annotation to it's original value.

why don't we use the returned function to restore the annotation?

.github/workflows/e2e-rosa-tests.yaml
yerzhan7: I am using different concurrency groups: `e2e-cluster-rosa-trusted` and `e2e-cluster-rosa-untrusted`.

EKS and ROSA tests are run in parallel.

.github/workflows/e2e-rosa-tests.yaml
yerzhan7: Same as in EKS, no? Or are you saying in EKS we had it for a different purpose that is not applicable for OpenShift?

.github/workflows/e2e-rosa-tests.yaml
yerzhan7: Not that I am aware of.

.github/workflows/e2e-test-rosa-untrusted.yaml
yerzhan7: This is copy/paste from `e2e-test-untrusted.yaml` file

tests/e2e-kubernetes/scripts/helm.sh
yerzhan7: > do we keep those ClusterRoleBindings in-between tests?

Yes. I did not find out any workarounds about this. This is due to the fact that we install privileged Helm chart not via OpenShift Operator. 

The other alternative is to perform cluster create and delete on each run (which takes 1h).

> any other resources that deletion skips?

Service Accounts as well. I will update the comment.

tests/e2e-kubernetes/scripts/run.sh
yerzhan7: > nodes have x86 architecture (no arm testing)

Correct.

> (when do we update it? use latest?

We can create custom scheduled GitHub Action to periodically update OpenShift and TerraForm versions. There is no "latest" alias.

tests/e2e-kubernetes/testsuites/credentials.go
yerzhan7: Primarily because we had it previously with `""` parameter. 

I think this is just a fail-safe to ensure each test starts with known deterministic baseline state in case someone forgets to do `deferCleanup` properly in the individual test.

.github/workflows/e2e-rosa-tests.yaml
yerzhan7: Will try to add check if cluster is too old and re-create it. Similar to what we have for EKS clusters right now.

.github/workflows/e2e-rosa-tests.yaml
```diff
@@ -110,6 +110,50 @@ jobs:
...
+
+          if [ $AGE_DAYS -ge $MAX_AGE_DAYS ]; then
+            echo "Cluster is $AGE_DAYS days old (>= $MAX_AGE_DAYS days). Destroying for fresh deployment."
+            echo "destroy_cluster=true" >> $GITHUB_OUTPUT
+          else
+            echo "Cluster is $AGE_DAYS days old (< $MAX_AGE_DAYS days). Will reuse existing cluster."
+            echo "destroy_cluster=false" >> $GITHUB_OUTPUT
+          fi
+      - name: Destroy old cluster
+        if: steps.check_cluster_age.outputs.destroy_cluster == 'true'
```
vladem: nit: destroy be default, opt-out if previous step added the variable:
```
if: steps.check_cluster_age.outputs.destroy_cluster != 'false'
```
otherwise I guess this is somewhat untested: have we ever seen the output being set to `true` and the step being run?

tests/e2e-kubernetes/testsuites/credentials.go
```diff
@@ -132,11 +133,13 @@ func (t *s3CSICredentialsTestSuite) DefineTests(driver storageframework.TestDriv
...
 	//
-	// In our test environment we add "AmazonS3FullAccess" policy to our EC2 instances
-	// (see "eksctl-patch.json") which allows Driver-level 3) to work.
+	// In our EKS test environment we add "AmazonS3FullAccess" policy to our EC2 instances
+	// (see "eksctl-patch.json") which allows Driver-level 4) to work.
 	// In order to test if other driver-level and pod-level credentials correctly work,
 	// we're trying to set a more restricted role (e.g. with "AmazonS3ReadOnlyAccess" policy),
-	// in these test cases to ensure it does not fallback to Driver-level 3) credentials.
+	// in these test cases to ensure it does not fallback to Driver-level 4) credentials.
+	// NOTE: In OpenShift clusters as IMDS is not available we rely on Driver-level IRSA 2)
```
vladem, line 141: As far as I understand only IRSA (driver / pod-level) and K8s secrets are the available credential provider methods on openshift. Subsequently the asks:
1. Could you please document that and make clear that other tests are skipped, i.e Pod Identity and Instance Profile suites.
1. Do we test `K8s secrets`, or we always fallback to IRSA? 
1. Do we test `pod-level IRSA` or always fallback to driver-level?

tests/e2e-kubernetes/scripts/run.sh
vladem: > We can create custom scheduled GitHub Action to periodically update OpenShift and TerraForm versions. There is no "latest" alias.

Can we? As far as I understand the version is fixed in `variables.tf`:
```
variable "openshift_version" {
  type    = string
  default = "4.20.11"
}
```

Since it's a code change, I think manual process is ok. May be we should check for new openshift versions at the same time when we're updating EKS version in our CI?

yerzhan7: Yeah, I've meant GitHub Action that will submit PRs for our manual review. Similar to this attempt with sidecar version updates: https://github.com/awslabs/mountpoint-s3-csi-driver/pull/605

tests/e2e-kubernetes/testsuites/credentials.go
yerzhan7: > Could you please document that and make clear that other tests are skipped, i.e Pod Identity and Instance Profile suites.

I will adjust this comment further to mention Pod Identity and Instance Profile tests are skipped. 

We also already log this in tests. E.g. `[SKIPPED] IAM instance profiles not available (no IMDS)`

Or you have meant documenting it elsewhere?

> Do we test K8s secrets, or we always fallback to IRSA?

Yes, we do test K8s secrets. See https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/21869706445/job/63121939783?pr=509#step:17:5345

> Do we test pod-level IRSA or always fallback to driver-level?

Yes - we test pod-level IRSA. See https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/21869706445/job/63121939783?pr=509#step:17:5775

.github/workflows/e2e-rosa-tests.yaml
yerzhan7: Will change default to destroy.

Will test it.

vladem: I see that destruction didn't go too well during the test, should we expect such failures every 3 days?😕
- https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/21901030424/job/63237612564

I am ok with fixing it as a follow up, but yet another reason for CI to fail is something that we'd ideally avoid.

tests/e2e-kubernetes/scripts/run.sh
vladem: Sounds good, we may address as a follow up.

.github/workflows/e2e-test-trusted.yaml
vladem: Let me know once this is reverted, otherwise LGTM.

.github/workflows/e2e-rosa-tests.yaml
yerzhan7: Yeah, I will follow-up separately on this. In my previous testing this didn't occur - VPC was always deleted properly.

I will trigger manual delete-rosa-cluster.yaml workflow after merge to test again.

Issue #309, opened 2024-12-05
Pod cannot be deleted if the S3 PVC fails to mount


geniass:
/kind bug
> **_NOTE:_**  If this is a filesystem related bug, please take a look at the [Mountpoint repo](https://github.com/awslabs/mountpoint-s3) to submit a [bug report](https://github.com/awslabs/mountpoint-s3/issues/new?assignees=&labels=bug&projects=&template=bug-report.yml)

**What happened?**
Mounting an s3 volume into a pod fails (FailedMount k8s event). Then the kubelet is unable to delete the pod because the vol_data.json file is missing. The pod hangs in the Terminating state forever.

Kubelet logs get flooded with:
```
[...]

PR #512, opened 2025-06-18
Recreate dbus connection

vladem:
*Issue #, if available:*  https://github.com/awslabs/mountpoint-s3-csi-driver/issues/309, https://github.com/awslabs/mountpoint-s3-csi-driver/issues/515, https://github.com/awslabs/mountpoint-s3-csi-driver/issues/516

*Description of changes:* 

Recreate dbus connection after receiving a "connection closed" error.


charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
```diff
@@ -236,8 +237,8 @@ spec:
             type: DirectoryOrCreate
         - name: systemd-bus
           hostPath:
-            path: /run/systemd/private
-            type: Socket
+            path: /run/systemd
+            type: DirectoryOrCreate
```
muddyfish: Wouldn't `Directory` be better here as we get early failure if systemd isn't installed?

pkg/system/systemd.go
```diff
@@ -75,15 +76,21 @@ func NewSystemdOsConnection() (*SystemdOsConnection, error) {
...
-		Conn:   conn,
-		Object: systemd,
+		Conn:     conn,
+		Object:   systemd,
+		isClosed: false,
 	}, nil
 }
 
 func (sc *SystemdOsConnection) Close() error {
+	// TODO: set isClosed flag?
```
muddyfish: Any reason why this TODO wasn't done?

pkg/system/systemd.go
```diff
@@ -254,11 +280,16 @@ func (s *SystemdSupervisor) RemoveServiceWatcher(serviceName string, ch chan<- *
...
 	signals := make(chan *dbus.Signal, signalBufferSize)
+	s.conn_mu.Lock()
 	s.conn.Signal(signals)
+	s.conn_mu.Unlock()
 
 	for sig := range signals {
 		s.dispatchSignal(sig)
 	}
+
+	klog.V(4).ErrorS(nil, "pollSignals finished") // goroutine terminates when underlying connection gets closed
```
muddyfish: Why is this an error?

pkg/system/systemd.go
```diff
@@ -252,18 +293,21 @@ func (s *SystemdSupervisor) RemoveServiceWatcher(serviceName string, ch chan<- *
...
 	s.conn.Signal(signals)
+	return signals
+}
 
+func (s *OsSystemdSupervisor) processSignals(signals <-chan *dbus.Signal) {
+	// dbus package closes the [signals] channel when the [s.conn] gets closed
 	for sig := range signals {
 		s.dispatchSignal(sig)
 	}
+	klog.V(5).Info("Systemd D-Bus signal channel closed — OsSystemdSupervisor stopped processing systemd signals")
```
muddyfish, line 306: Should we also call `.close()` here given we know it's closed?

pkg/system/systemd.go
```diff
@@ -7,14 +7,17 @@ import (
...
 
 	"github.com/godbus/dbus/v5"
 	"k8s.io/klog/v2"
 )
 
 const (
 	systemdSocket    = "unix:path=/run/systemd/private"
-	signalBufferSize = 4096 // Messages are dropped if the buffer fills, so make it large
+	signalBufferSize = 4096            // Messages are dropped if the buffer fills, so make it large
+	runUnitTimeout   = 5 * time.Minute // Maximum time to wait for unit updates
```
muddyfish: 5 minutes feels very long for waiting for an unmount or a backgrounded MP to start up

pkg/system/systemd_runner.go
```diff
@@ -0,0 +1,107 @@
...
+//
+// The replacement of the supervisor is synchronized via an internal mutex
+// to ensure only one reconnection attempt occurs at a time, while allowing
+// lock-free reads for all other callers.
+//
+// Method calls on the underlying SystemdSupervisor are safe to invoke
+// concurrently; however, reconnection logic is protected to avoid redundant
+// supervisor creation or resource leaks.
+type SystemdRunner struct {
+	mu                sync.Mutex // protects write path only
```
muddyfish: `systemdSupervisorMutex`?

pkg/system/systemd_runner.go
```diff
@@ -0,0 +1,107 @@
...
+	return runner, nil
+}
+
+func (s *SystemdRunner) StartService(ctx context.Context, config *ExecConfig) (string, error) {
+	err := s.ensureConnectionOpen()
+	if err != nil {
+		return "", nil
+	}
+
+	return s.systemdSupervisor.Load().StartService(ctx, config)
```
muddyfish: Could we add a comment here saying why it's fine to run `StartService/RunOneshot` without retry logic in case the connection closes partway through after ensuring it's open?

pkg/system/systemd_runner.go
```diff
@@ -0,0 +1,107 @@
...
+		return nil, err
+	}
+	runner.systemdSupervisor.Store(supervisor)
+	return runner, nil
+}
+
+func (s *SystemdRunner) StartService(ctx context.Context, config *ExecConfig) (string, error) {
+	err := s.ensureConnectionOpen()
+	if err != nil {
+		return "", nil
```
muddyfish: Why aren't we passing through the error?

pkg/system/systemd_runner.go
```diff
@@ -0,0 +1,107 @@
...
+	// Reload to avoid race: someone else may have swapped it already
+	current = s.systemdSupervisor.Load()
+	if !current.IsConnectionClosed() {
+		return nil // another goroutine already fixed it
+	}
+
+	klog.V(5).Info("SystemdRunner: systemd connection was closed, re-creating systemd supervisor")
+	current.Stop()
+
+	newSup, err := s.supervisorFactory.StartSupervisor()
```
muddyfish: This could be a function reused in `StartSystemdRunner`

pkg/system/systemd_runner.go
```diff
@@ -0,0 +1,107 @@
...
+	if err != nil {
+		return err
+	}
+	s.systemdSupervisor.Store(newSup)
+
+	return nil
```
muddyfish: Would it be useful to make this return `newSup`? Then we could avoid calling `load` again

pkg/system/systemd_runner_test.go
```diff
@@ -0,0 +1,86 @@
...
+	ctx := context.Background()
+	config := &system.ExecConfig{
+		Name:        "test.service",
+		ExecPath:    "/bin/test",
+		Args:        []string{"arg1", "arg2"},
+		Description: "Test service",
+	}
+
+	// First supervisor reports closed connection and gets stopped
+	mockSupervisor1.EXPECT().IsConnectionClosed().Return(true)
```
muddyfish: Why is this called twice?

muddyfish: Could we also include the order of calls expected so it's clear

pkg/system/systemd_runner_test.go
```diff
@@ -0,0 +1,86 @@
...
+	mockSupervisor1.EXPECT().IsConnectionClosed().Return(true)
+	mockSupervisor1.EXPECT().Stop()
+
+	// Create second supervisor that will handle the service start
+	mockSupervisor2 := mock_system.NewMockSystemdSupervisor(ctrl)
+	mockFactory.EXPECT().StartSupervisor().Return(mockSupervisor2, nil)
+
+	// Second supervisor handles service start
+	mockSupervisor2.EXPECT().StartService(ctx, config).Return("test output", nil)
```
muddyfish: Can we have a assert not called `StartService` on `mockSupervisor`

pkg/system/systemd.go
vladem: Technically this is not required for this fix to function: we rely on `OsSystemdSupervisor::Close` being called after detecting the closed connection which will result in closure of all watcher channels and subsequent termination of goroutines, thus no resources will be leaked even with no timeout. 

I don't want to risk reporting a mount failure in case of `mount-s3` taking a bit longer to start without a good reason. **I am going to remove this timeout.** I think we can rely on `<-ctx.Done()` to complete at some point (when NodePublishRequest is abandoned by kubelet).

UPD. we already specify a timeout: https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/pkg/driver/node/mounter/systemd_mounter.go#L68

pkg/system/systemd_runner.go
vladem: adding to the doc comment just before `SystemdRunner struct` definition

pkg/system/systemd_runner.go
vladem: this, combined with `ensureConnectionOpen` returning `SystemdSupervisor`, becomes a bit hard to follow

I tried both:
1. creating and reusing `SystemdRunner::startSystemdSupervisor` in `StartSystemdRunner`
2. making `ensureConnectionOpen` to return `SystemdSupervisor`

and liked \2 better since:
- fewer `Load()` from atomic => better readability, slightly more efficient
- no nested calls for a simple operation like create `SystemdSupervisor` and `Store` it

does it sound ok?

pkg/system/systemd_runner_test.go
vladem: added a comment with explanation:
```
// First supervisor reports closed connection twice (fast + slow path)
```

pkg/system/systemd_runner_test.go
vladem: failure on unexpected call is the default behaviour of gomock, added a comment with explanation:
```
// If mockSupervisor1.StartService(...) is called instead, the test will fail.
```

pkg/system/systemd_runner_test.go
```diff
@@ -64,17 +64,25 @@ func TestSystemdRunnerRecreateConnection(t *testing.T) {
...
+		mockSupervisor1.EXPECT().IsConnectionClosed().Return(true),
+
+		// First supervisor is stopped
+		mockSupervisor1.EXPECT().Stop(),
+
+		// Second supervisor created
+		mockFactory.EXPECT().StartSupervisor().Return(mockSupervisor2, nil),
+
+		// This expectation ensures that only mockSupervisor2 is allowed to handle StartService.
+		// If mockSupervisor1.StartService(...) is called instead, the test will fail.
```
muddyfish, line 83: What if both are called?

pkg/system/systemd_runner_test.go
```diff
@@ -0,0 +1,94 @@
+package system_test
```
muddyfish, line 1: Is it possible to add some integration tests for this?

vladem: It's complicated, I think we should rely on unit tests and a manual one-off integration test with `sudo systemctl daemon-reexec` command used to disrupt the dbus connection.

Complications are:
- we use a single driver in all tests and we don't want closed dbus connection to affect other tests => a separate step for this test in a workflow is required
- triggering closed dbus connection from test would require to spawn a privileged pod like this (if something goes wrong, we'll have a broken node and all following tests will fail):

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: systemd-reexec
spec:
  nodeName: <node_name>
  hostPID: true
  containers:
    - name: reexec
      image: ubuntu:20.04
      securityContext:
        privileged: true
      command:
        - bash
        - -c
        - |
          chroot /host /bin/bash -c 'systemctl daemon-reexec'
      volumeMounts:
        - name: host-root
          mountPath: /host
  volumes:
    - name: host-root
      hostPath:
        path: /
        type: Directory
  restartPolicy: Never
```

pkg/system/systemd_runner_test.go
vladem: the test will fail as well

charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
```diff
@@ -155,7 +155,8 @@ spec:
               mountPropagation: HostToContainer
               {{- end }}
             - name: systemd-bus
-              mountPath: /run/systemd/private
+              mountPath: /run/systemd
```
unexge, line 158: We also need to apply the same changes to `deploy/kubernetes/base/node-daemonset.yaml` for Kustomization deployments

unexge:
There is also a similar PR in runc to re-create/-try when the dbus connection gets closed, might be helpful to take some insperation: https://github.com/opencontainers/runc/pull/2923

pkg/system/systemd.go
```diff
@@ -131,6 +138,10 @@ func (sc *SystemdOsConnection) callDbus(ctx context.Context, method string, ret
 	select {
 	case call := <-ch:
 		if call.Err != nil {
+			if call.Err == dbus.ErrClosed {
```
unexge: We should probably use `errors.Is(...)` here just in case the error is wrapped

PR #528, opened 2025-07-01
Remove kOps testing clusters

unexge:


muddyfish:
PR description is missing the justification

.github/workflows/e2e-tests.yaml
```diff
@@ -127,32 +126,6 @@ jobs:
           ACTION: "run_tests"
         run: |
           tests/e2e-kubernetes/scripts/run.sh
-      - name: Run Performance Tests
```
muddyfish, line 130: Are these anything we care about?

unexge: We were not using them to my knowledge. We can add them back properly in the future if we find value on it.

PR #531, opened 2025-07-09
Upgrade installation and upgrading documentations

unexge:
Rendered [`INSTALL.md`](https://github.com/unexge/mountpoint-s3-csi-driver/blob/6626482c8e751eec8d3c169321f949f41f105f9b/docs/INSTALL.md) and [`UPGRADING_TO_V2.md`](https://github.com/unexge/mountpoint-s3-csi-driver/blob/6626482c8e751eec8d3c169321f949f41f105f9b/docs/UPGRADING_TO_V2.md).

---

### DIFF
--- a/charts/aws-mountpoint-s3-csi-driver/values.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/values.yaml
@@ -76,7 +76,9 @@ sidecars:
     resources: {}
 
 controller:
-  # Consider deploying the controller component to special nodes for controller components, e.g.
+  # Consider deploying the controller component to special nodes for controller components.
+  # See https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/docs/INSTALL.md#configuring-nodeSelector-for-the-controller-component for more details.
+  # For example:
   # nodeSelector:
   #   kubernetes.io/role: control-plane
   serviceAccount:
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -12,7 +12,7 @@
 
 ### Cluster setup (optional)
 
-If you don't have an existing cluster, you can follow these steps to setup an Amazon EKS cluster using [eksctl](https://eksctl.io/). See [eksctl documentation](https://eksctl.io/getting-started/) for more details.
+If you don't have an existing cluster, you can follow these steps to setup an Amazon EKS cluster using [eksctl](https://eksctl.io/).
 
 - Set a cluster-name and a region:
 
@@ -27,10 +27,11 @@ export REGION=us-west-2
 eksctl create cluster \
   --name $CLUSTER_NAME \
   --region $REGION \
-  --with-oidc \
-  --ssh-access \
-  --ssh-public-key <my-key>
+  --with-oidc \ # To enable IAM Roles for Service Accounts (IRSA)
+  ...
 ```
+
+See [eksctl documentation](https://eksctl.io/getting-started/) for more details on the configuration you can use while creating your cluster.
 
 - Setup kubectl context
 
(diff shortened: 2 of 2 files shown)


docs/INSTALL.md
```diff
@@ -28,48 +32,43 @@ eksctl create cluster \
   --ssh-public-key <my-key>
```
renanmagagnin: Would it be worth it to add some additional info on ssh-public-key?

charts/aws-mountpoint-s3-csi-driver/values.yaml
```diff
@@ -76,6 +76,9 @@ sidecars:
     resources: {}
 
 controller:
+  # Consider deploying the controller component to special nodes for controller components, e.g.
+  # nodeSelector:
+  #   kubernetes.io/role: control-plane
```
renanmagagnin: Should we mention the part of the documentation (`INSTALL.md`) that covers this? Similar to what you did for the SA above

docs/INSTALL.md
```diff
@@ -28,48 +32,43 @@ eksctl create cluster \
...
 ```
 
-### Configure access to S3
+### Configure access to Amazon S3
 
 Assuming you have a cluster created, the next step is to ensure the CSI Driver will be able to access Amazon S3.
 
 There's a few different options for providing credentials for the Mountpoint CSI Driver to use when accessing Amazon S3.
-See the [AWS credentials section of CONFIGURATION.md](./CONFIGURATION.md#aws-credentials) for instructions on how to setup IAM principals for accessing Amazon S3.
+See the [AWS Credentials](CONFIGURATION.md#aws-credentials) for instructions and different ways of configuring AWS credentials for accessing Amazon S3.
```
renanmagagnin:
```suggestion
See [AWS Credentials](CONFIGURATION.md#aws-credentials) for instructions and different ways of configuring AWS credentials for accessing Amazon S3.
```

docs/INSTALL.md
```diff
@@ -28,48 +32,43 @@ eksctl create cluster \
...
-You may deploy the Mountpoint for Amazon S3 CSI driver via Kustomize, Helm, or as an [Amazon EKS managed add-on](https://docs.aws.amazon.com/eks/latest/userguide/eks-add-ons.html#workloads-add-ons-available-eks).
+### Deploy Mountpoint for Amazon S3 CSI Driver
 
-#### Kustomize
-```sh
-kubectl apply -k "github.com/awslabs/mountpoint-s3-csi-driver/deploy/kubernetes/overlays/stable/"
-```
-> [!WARNING]
-> Using the main branch to deploy the driver is not supported. The main branch may contain upcoming features incompatible with the currently released stable version of the driver.
+You may deploy the Mountpoint for Amazon S3 CSI Driver via [Helm](https://helm.sh/), [Kustomize](https://github.com/kubernetes-sigs/kustomize), or as an [Amazon EKS managed add-on](https://docs.aws.amazon.com/eks/latest/userguide/eks-add-ons.html#workloads-add-ons-available-eks).
```
renanmagagnin: We are mentioning the EKS add-on here but only having sections for the first two. I think we should cover it too. One option would be to reuse the content of the warning we added to the start. Another option could be to move what we have in the warning to a section here and simplify the warning by just referencing this introduced section. We could also consider having it as the first in the order instead of the last.

docs/INSTALL.md
unexge: Removed mention of the `--ssh-access` and `--ssh-public-key` and just added a link to official documentation for all configuration options.

docs/INSTALL.md
unexge: Added section for EKS managed add-on with the same content as the banner.

renanmagagnin (approved):

PR #532, opened 2025-07-11
Add end-to-end test for upgrading the CSI Driver

unexge:
## Change

This new test suite verifies that upgrading the CSI Driver to a new version doesn't break existing workloads that created with the previous version. The test involves the following steps for testing the upgrade from X to Y:
- Uninstall the CSI Driver if it's installed with any version
- Install version X
- Start three workloads
   - A Pod using `S3ReadOnlyAccess` policy with driver-level IRSA
   - A Pod using `S3FullAccess` policy with pod-level IRSA
   - A Pod using `S3ReadOnlyAccess` policy with pod-level IRSA
- Write a test file on writable workloads, and assert read-only workloads can list but cannot write
- Upgrade to version Y
- Start three workloads
   - A Pod using `S3ReadOnlyAccess` policy with driver-level IRSA
   - A Pod using `S3FullAccess` policy with pod-level IRSA
   - A Pod using `S3ReadOnlyAccess` policy with pod-level IRSA
- Write a test file on writable workloads, and assert read-only workloads can list but cannot write
- Every minute for an hour repeat:
   - Ensure all writable pods can read the original file, and can do basic file operations (i.e., writing, reading, listing and deleting)
   - Ensure all read-only pods can still list but cannot write
- Ensure all workloads can be successfully terminated

## Testing

Verified that this test fails if I disable `systemd` mount support by setting `supportLegacySystemDMounts` to `false`.  This might happen after 2 hours or so because:
  1. We ask for service account tokens that valid for 1 hour (see `CSIDriver` object)
  2. Session duration of the IAM roles we assume is 1 hour

So, it might take ~2 hours at worst case to hit both cycles, and get the failure. That's why we run the upgrade test for 150 minutes.

## TODOs:

- [x] Add a GitHub workflow to manually trigger upgrade tests between two given versions - Added a workflow with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/532/commits/2348101b9c99567904bd80d63fa69c342ff68d83 but couldn't test it. It needs to be in main to be able to triggered manually, so, I'll send follow-up PRs if I made a mistake on the workflow file – which is very likely 😅 

---


tests/e2e-kubernetes/testsuites/upgrade.go
```diff
@@ -0,0 +1,376 @@
...
+	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
+	storageframework "k8s.io/kubernetes/test/e2e/storage/framework"
+	admissionapi "k8s.io/pod-security-admission/api"
+)
+
+// This value defines how long the upgrade test should take.
+//
+// This needs to be at least more than 2 hours because
+//  1. We ask for service account tokens that valid for 1 hour (see `CSIDriver` object)
+//  2. Session duration of the IAM roles we assume is 1 hour
```
renanmagagnin, line 30: Out of curiosity, would it be possible to decrease these durations from one hour to speed the tests without sacrificing what is being tested?

tests/e2e-kubernetes/testsuites/upgrade.go
```diff
@@ -0,0 +1,376 @@
...
+
+		// Make sure to start from a clean state
+		uninstallCSIDriver(cfg)
+
+		// Install the previously released version from our Helm repo
+		chartPath := pullCSIDriver(settings, cfg, helmChartPreviousVersion)
+		installCSIDriver(cfg, helmChartPreviousVersion, chartPath)
+
+		// Configure driver-level IRSA with "S3ReadOnlyAccess" policy
+		updateCSIDriversServiceAccountRole(ctx, oidcProvider, iamPolicyS3ReadOnlyAccess)
```
renanmagagnin, line 190: Would it be worth to testing with driver-level full access?

tests/e2e-kubernetes/testsuites/upgrade.go
```diff
@@ -0,0 +1,376 @@
...
+		// Create two SAs for pod-level IRSA with "S3FullAccess" and "S3ReadOnlyAccess" policies
+		pliFullAccessSA, pliReadOnlyAccessSA := createServiceAccountWithPolicy(ctx, iamPolicyS3FullAccess), createServiceAccountWithPolicy(ctx, iamPolicyS3ReadOnlyAccess)
+
+		// Create three workloads with different SAs
+		dliReadOnlyAccessPod := createPod(ctx, "default")
```
renanmagagnin, line 195: np: refactor usages of the default SA into a variable

tests/e2e-kubernetes/testsuites/upgrade.go
```diff
@@ -0,0 +1,376 @@
...
+		fullAccessPods = append(fullAccessPods, pliFullAccessPodNewVersion)
+		readOnlyAccessPods = append(readOnlyAccessPods, dliReadOnlyAccessPodNewVersion, pliReadOnlyAccessPodNewVersion)
+		for _, pod := range []*v1.Pod{pliFullAccessPodNewVersion} {
+			checkWriteToPath(f, pod, testFile, testWriteSize, seed)
+			checkReadFromPath(f, pod, testFile, testWriteSize, seed)
+		}
+		for _, pod := range readOnlyAccessPods {
+			checkListingPath(f, pod, e2epod.VolumeMountPath1)
+			checkWriteToPathFails(f, pod, testFile, testWriteSize, seed)
+		}
```
renanmagagnin, line 239: I think we can reduce repetition in the creation and validation of the workloads.
How about we refactor the workload creation into a function that returns the two lists of pods (fullAccess and readOnly)? Then, could also refactor the checking of access into a function that takes the two lists and performs appropriate validation (line 232-239).

tests/e2e-kubernetes/testsuites/upgrade.go
```diff
@@ -0,0 +1,376 @@
...
+		upgradeCSIDriver(cfg, helmChartNewVersion, chartPath)
+
+		// Create new workloads after the upgrade
+		dliReadOnlyAccessPodNewVersion := createPod(ctx, "default")
+		pliFullAccessPodNewVersion := createPod(enablePLI(ctx), pliFullAccessSA.Name)
+		pliReadOnlyAccessPodNewVersion := createPod(enablePLI(ctx), pliReadOnlyAccessSA.Name)
+		fullAccessPods = append(fullAccessPods, pliFullAccessPodNewVersion)
+		readOnlyAccessPods = append(readOnlyAccessPods, dliReadOnlyAccessPodNewVersion, pliReadOnlyAccessPodNewVersion)
+		for _, pod := range []*v1.Pod{pliFullAccessPodNewVersion} {
```
renanmagagnin, line 232: Why are we checking only the new version full access here instead of the entire fullAccessPods list similar to what we do for the read only validation?

tests/e2e-kubernetes/testsuites/upgrade.go
```diff
@@ -0,0 +1,376 @@
...
+		if oidcProvider == "" {
+			Fail("Please configure OIDC provider for the testing cluster")
+		}
+
+		framework.Logf("Testing upgrade from %q to %q...", helmChartPreviousVersion, helmChartNewVersion)
+
+		settings, cfg := initHelmClient()
+
+		// Make sure to start from a clean state
+		uninstallCSIDriver(cfg)
```
renanmagagnin, line 183: Just to make sure: would it make sense to also uninstall the EKS ad-don here to make sure it's in a clean state? I'm not sure how unclean state could happen so don't know if this makes sense.

renanmagagnin (approved):

tests/e2e-kubernetes/testsuites/upgrade.go
unexge: For 1) this value is currently hard-coded, and I didn't want to add a new Helm chart variable for just the tests. We can consider patching the value in-place
For 2) unfortunately for IAM role this is the minimum duration

I can take a look if 1) is possible to do

tests/e2e-kubernetes/testsuites/upgrade.go
unexge: Since we can't run them in parallel, it would require having another test case for the whole duration. The read-only one provides more guarantee as we have "S3FullAccess" on the instance profiles, asserting read-only kinda captures we correctly refresh the credentials at driver-level.

tests/e2e-kubernetes/testsuites/upgrade.go
unexge: I didn't add `--allow-overwrite` to mount arguments and this would fail without that. But I think it's a good point, I can just add `--allow-overwrite` and simply this part.

tests/e2e-kubernetes/testsuites/upgrade.go
unexge: It would fail at installation or uninstall phase if the EKS add-on is installed due to resource ownership – Helm would check if the Kubernetes Resource originally created by Helm, and fail if not unless we pass `--take-ownership`.

Issue #534, opened 2025-07-11
Karpenter Not Considering Mountpoint Pod Resources


kevarr:
/feature

**Is your feature request related to a problem? Please describe.**

Hello, I've been kicking the tires a little on `v2.0.0-beta` and really liking the new enhancements!

While testing, I've identified the following scheduling problem when using the AWS Mountpoint S3 CSI Driver with Karpenter. The sequence of events is:

1.  An application `Pod` requiring an S3 volume is submitted to the cluster.
2.  Karpenter sees this `Pod` and provisions a new, right-sized `Node` based *only* on the application `Pod`'s resource requests.
3.  Once the application `Pod` is scheduled to the new `Node`, the S3 CSI Driver controller creates a *separate* `Mountpoint` Pod to manage the volume mount.
4.  This `Mountpoint` Pod has hard node affinity for the application's `Node`, but its resource requirements were not included in Karpenter's initial calculation.
5.  As a result, the `Mountpoint` Pod often fails to schedule due to `Insufficient cpu/memory` on the node that was provisioned for the application. This leaves the application `Pod` stuck in a `ContainerCreating` state, waiting for a volume that can never be mounted.

**Describe the solution you'd like in detail**

One possible solution would be to refactor the CSI driver to inject the Mountpoint container directly into the application pod as a sidecar container. This way the `kube-scheduler` would see the *combined* resource requests of the application container and the injected Mountpoint sidecar. Karpenter would then receive a single Pod with the total resource requirements and provision a correctly-sized node from the beginning. This should eliminate the race condition and align the lifecycle of the Mountpoint container with the application pod it serves.

**Describe alternatives you've considered**
We have considered two primary workarounds, both of which have drawbacks:

1.  **Manual Resource Padding:** We can manually increase the CPU and memory requests of every application pod to account for the future `Mountpoint` Pod.
2.  **Preemptive "Pause" Pods:** We have experimented with deploying low-priority "pause" pods via a `DaemonSet` to reserve space on nodes. The `Mountpoint` Pod (with a higher priority) can then preempt these pause pods to acquire the needed resources.

unexge:
Hey @kevarr, thanks for the report. We are aware of this problem, and was going to create a tracking ticket – but you were faster :) 

[...]

PR #536, opened 2025-07-15
Add support for reserving headroom for Mountpoint Pods

unexge:
This PR adds an experimental feature for reserving headroom for Mountpoint Pods to help with issues like https://github.com/awslabs/mountpoint-s3-csi-driver/issues/534.

## Why?

The CSI Driver v2 spawns Mountpoint Pods into the same nodes as the workloads to provide volumes. The CSI Driver spawns these Mountpoint Pods _after_ the workloads are assigned to nodes. This approach is taken because we might be able to [share an existing Mountpoint Pod](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/v2/docs/MOUNTPOINT_POD_SHARING.md) on a specific node, eliminating the need to spawn a new one. Therefore, the CSI Driver needs to wait until a workload is scheduled to a specific node before deciding whether to share an existing Mountpoint Pod or spawn a new one.

This dynamic spawning of Mountpoint Pods on specific nodes creates a problem with node autoscalers like Karpenter. Since autoscalers are not aware of Mountpoint Pod requirements when making scaling decisions, they may make suboptimal choices. As a result, there may not be enough space for a Mountpoint Pod on the specific node where the workload is deployed.

This feature aims to mitigate this issue by reserving headroom for upcoming Mountpoint Pods, ensuring autoscalers create instances large enough to host Mountpoint Pods alongside workloads.

## How it's used?

Opting into this feature requires two steps.

### 1. Enable this feature during Helm chart installation (one-time setup)

```bash
$ helm upgrade --install aws-mountpoint-s3-csi-driver \
    --set experimental.reserveHeadroomForMountpointPods=true \
    ...
    aws-mountpoint-s3-csi-driver/aws-mountpoint-s3-csi-driver
```

This is behind a feature flag because it adds cluster-wide `patch` permissions on pods to the CSI Driver Controller's Service Account. This permission is needed for:
  1. Adding labels to Workload Pods for use in inter-pod affinity rules in Headroom Pods
  2. Removing scheduling gates from Workload Pods after creating Headroom Pods to make the Workload Pod ready for scheduling

### 2. Add a scheduling gate for each Workload Pod where reserving headroom for Mountpoint Pods is desired

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: workload
spec:
  schedulingGates:
    - name: s3.csi.aws.com/reserve-headroom-for-mppod # <-- HERE
  containers:
    # ...
  volumes:
    - name: vol
      persistentVolumeClaim:
        claimName: s3-pvc
```

This is currently done manually, and we may move this logic into a mutating admission webhook in the future to make configuration easier.

## How it works?

When the CSI Driver detects a Workload Pod using the scheduling gate to enable this feature, it performs the following steps:

  1. Labels the Workload Pod to use inter-pod affinity rules in the Headroom Pods
  2. Creates Headroom Pods using a pause container with inter-pod affinity rule to the Workload Pod - since node autoscalers like [Karpenter supports inter-pod affinity rules](https://karpenter.sh/docs/concepts/scheduling/#pod-affinityanti-affinity), this should help them to choose a right instance type
  3. Ungates the scheduling gate from the Workload Pod to let it scheduled - alongside the Headroom Pods if possible
  4. Schedules Mountpoint Pod if necessary (i.e., the CSI Driver cannot share an existing Mountpoint Pod) into the same node as the Workload and Headroom Pods using a preempting priority class
  5. Mountpoint Pod most likely preempts the Headroom Pods if there is no space in the node - as the Headroom Pods uses a negative priority -, or just gets scheduled if there is enough space for all pods
  6. Deletes the Headroom Pods as soon as the Workload Pod is running or terminated - as Mountpoint Pods are already scheduled or no longer needed

## Limitations

This approach has some downsides:
- Overprovisioning: It may cause node autoscalers to create bigger instances than the cluster needs.
- Inter-pod affinity: Inter-pod affinity rules are compute-intensive and may cause significant slowdowns on scheduling in clusters with more than several hundred nodes
- Preemption: Mountpoint Pods would be spawned with preempting priority class if this feature is on, and they might preempt some other pods rather than Headroom Pods. Even though the CSI Driver sets a negative priority on Headroom Pods to make them the first target in case of preemption there are no guarantees that the Headroom Pods won't be evicted by other high priority pods and causing Mountpoint Pods to preempting other pods.

See [the documentation](https://github.com/unexge/mountpoint-s3-csi-driver/blob/2d2b6dd68f1b78bf8a458e170ba9cbb894f5f5a5/docs/HEADROOM_FOR_MPPOD.md) for more details. 

---

### DIFF
--- a/charts/aws-mountpoint-s3-csi-driver/templates/mount-s3-priority.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/mount-s3-priority.yaml
@@ -23,7 +23,7 @@ globalDefault: false
 description: >-
     Preempting priority class for Mountpoint Pods.
     It has the highest possible value for non-builtin PriorityClasses to ensure Mountpoint Pods get scheduled quickly and not evicted first in case of a resource pressure.
-    Mountpoint Pods spawned with this priority class by default if "Reserving headroom for Mountpoint Pods" feature is enabled.
+    Mountpoint Pods are spawned with this priority class by default if "Reserving headroom for Mountpoint Pods" feature is enabled.
     See https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/docs/HEADROOM_FOR_MPPOD.md for more details.
 
 ---
@@ -39,4 +39,3 @@ description: >-
     See https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/docs/HEADROOM_FOR_MPPOD.md for more details.
 
 {{- end }}
-
--- a/charts/aws-mountpoint-s3-csi-driver/templates/mount-s3-priority.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/mount-s3-priority.yaml
@@ -10,7 +10,7 @@ description: >-
     It has the highest possible value for non-builtin PriorityClasses to ensure Mountpoint Pods get scheduled quickly and not evicted first in case of a resource pressure.
     It will not cause other Pods to be preempted.
 
-{{- if .Values.experimental.reserveHeadroomForMountpointPods -}}
+{{ if .Values.experimental.reserveHeadroomForMountpointPods -}}
 
 ---
 apiVersion: scheduling.k8s.io/v1
@@ -35,5 +35,5 @@ description: >-
     Default priority class for Headroom Pods.
     It has a negative priority to allow Mountpoint Pods to evict them.
 
-{{- end -}}
+{{- end }}
 
--- a/charts/aws-mountpoint-s3-csi-driver/templates/mount-s3-priority.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/mount-s3-priority.yaml
@@ -23,6 +23,8 @@ globalDefault: false
 description: >-
     Preempting priority class for Mountpoint Pods.
     It has the highest possible value for non-builtin PriorityClasses to ensure Mountpoint Pods get scheduled quickly and not evicted first in case of a resource pressure.
+    Mountpoint Pods spawned with this priority class by default if "Reserving headroom for Mountpoint Pods" feature is enabled.
+    See https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/docs/HEADROOM_FOR_MPPOD.md for more details.
 
 ---
 apiVersion: scheduling.k8s.io/v1
@@ -34,6 +36,7 @@ globalDefault: false
 description: >-
     Default priority class for Headroom Pods.
     It has a negative priority to allow Mountpoint Pods to evict them.
+    See https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/docs/HEADROOM_FOR_MPPOD.md for more details.
 
 {{- end }}
 
--- a/charts/aws-mountpoint-s3-csi-driver/templates/mount-s3-priority.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/mount-s3-priority.yaml
@@ -9,3 +9,31 @@ description: >-
     Default priority class for Mountpoint Pods.
     It has the highest possible value for non-builtin PriorityClasses to ensure Mountpoint Pods get scheduled quickly and not evicted first in case of a resource pressure.
     It will not cause other Pods to be preempted.
+
+{{- if .Values.experimental.reserveHeadroomForMountpointPods -}}
+
+---
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: mount-s3-preempting-critical
+value: 1000000000
+preemptionPolicy: PreemptLowerPriority
+globalDefault: false
+description: >-
+    Preempting priority class for Mountpoint Pods.
+    It has the highest possible value for non-builtin PriorityClasses to ensure Mountpoint Pods get scheduled quickly and not evicted first in case of a resource pressure.
+
+---
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: mount-s3-headroom
+value: -1000
+globalDefault: false
+description: >-
+    Default priority class for Headroom Pods.
+    It has a negative priority to allow Mountpoint Pods to evict them.
+
+{{- end -}}
+
--- a/cmd/aws-s3-csi-controller/csicontroller/reconciler.go
+++ b/cmd/aws-s3-csi-controller/csicontroller/reconciler.go
@@ -170,8 +170,8 @@ func (r *Reconciler) reconcileWorkloadPod(ctx context.Context, pod *corev1.Pod) 
 			if i == 0 {
 				err = r.labelWorkloadPodForHeadroomPod(ctx, pod, log)
 				if err != nil {
-					errs = append(errs, err)
-					continue
+					log.Error(err, "Failed to label Workload Pod for Headroom Pods")
+					return reconcile.Result{}, err
 				}
 			}
 
--- a/cmd/aws-s3-csi-controller/csicontroller/reconciler.go
+++ b/cmd/aws-s3-csi-controller/csicontroller/reconciler.go
@@ -132,7 +132,7 @@ func (r *Reconciler) reconcileWorkloadPod(ctx context.Context, pod *corev1.Pod) 
 		return reconcile.Result{}, nil
 	}
 
-	scheduled := pod.Spec.NodeName != ""
+	scheduled := isPodScheduled(pod)
 	needsHeadroomForMountpointPod := mppod.ShouldReserveHeadroomForMountpointPod(pod)
 	hasHeadroomPods := mppod.WorkloadHasLabelPodForHeadroomPod(pod)
 	if !scheduled && !needsHeadroomForMountpointPod && !hasHeadroomPods {
@@ -145,35 +145,16 @@ func (r *Reconciler) reconcileWorkloadPod(ctx context.Context, pod *corev1.Pod) 
 		priorityClassKind = mppod.PreemptingPriorityClass
 	}
 
-	var requeue bool
+	volumes, requeue, err := r.getWorkloadVolumes(ctx, pod)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
 	var errs []error
+	numHeadroomPods, numRemovedHeadroomPods := 0, 0
 
-	numVolumes, numHeadroomPods, numRemovedHeadroomPods := 0, 0, 0
-
-	for _, vol := range pod.Spec.Volumes {
-		podPVC := vol.PersistentVolumeClaim
-		if podPVC == nil {
-			continue
-		}
-
-		// If PVC has no bound PVs yet, `getBoundPVForPodClaim` will return `errPVCIsNotBoundToAPV`.
-		// In this case we'll just return `reconcile.Result{Requeue: true}` here, which will bubble up to the
-		// original `Reconcile` call and will cause a retry for this Pod with an exponential backoff.
-		pvc, pv, err := r.getBoundPVForPodClaim(ctx, pod, podPVC)
-		if err != nil {
-			if errors.Is(err, errPVCIsNotBoundToAPV) {
-				requeue = true
-			} else {
-				errs = append(errs, err)
-			}
-			continue
-		}
-
-		csiSpec := extractCSISpecFromPV(pv)
-		if csiSpec == nil {
-			continue
-		}
-		numVolumes += 1
+	for i, vol := range volumes {
+		pv, pvc := vol.pv, vol.pvc
 
 		if scheduled {
 			log.V(debugLevel).Info("Found bound PV for PVC", "pvc", pvc.Name, "volumeName", pv.Name)
@@ -186,7 +167,7 @@ func (r *Reconciler) reconcileWorkloadPod(ctx context.Context, pod *corev1.Pod) 
 			}
 		} else if needsHeadroomForMountpointPod {
 			// Label the workload pod once for headroom pods
-			if numVolumes == 1 {
+			if i == 0 {
 				err = r.labelWorkloadPodForHeadroomPod(ctx, pod, log)
 				if err != nil {
 					errs = append(errs, err)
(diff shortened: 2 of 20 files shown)


tests/e2e-kubernetes/testsuites/headroom.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func (t *s3CSIHeadroomTestSuite) SkipUnsupportedTests(_ storageframework.TestDriver, pattern storageframework.TestPattern) {
+	if pattern.VolType != storageframework.PreprovisionedPV {
+		e2eskipper.Skipf("Suite %q does not support %v", t.tsInfo.Name, pattern.VolType)
+	}
+}
+
+func (t *s3CSIHeadroomTestSuite) DefineTests(driver storageframework.TestDriver, pattern storageframework.TestPattern) {
+	f := framework.NewFrameworkWithCustomTimeouts(NamespacePrefix+"cache", storageframework.GetDriverTimeouts(driver))
```
yerzhan7: `cache` -> `headroom`?

tests/e2e-kubernetes/testsuites/headroom.go
```diff
@@ -0,0 +1,157 @@
...
+			deferCleanup(vol.CleanupResource)
+
+			pod := e2epod.MakePod(f.Namespace.Name, nil, []*v1.PersistentVolumeClaim{vol.Pvc}, admissionapi.LevelBaseline, "")
+			pod.Spec.SchedulingGates = []v1.PodSchedulingGate{{Name: headroomSchedulingGate}}
+
+			pod, err := createPod(ctx, f.ClientSet, f.Namespace.Name, pod)
+			framework.ExpectNoError(err)
+			deferCleanup(func(ctx context.Context) error { return e2epod.DeletePodWithWait(ctx, f.ClientSet, pod) })
+
+			checkBasicFileOperations(pod, e2epod.VolumeMountPath1)
```
yerzhan7, line 151: np: Can we also validate that headroom pod does not exist?

docs/HEADROOM_FOR_MPPOD.md
```diff
@@ -0,0 +1,79 @@
...
+  3. Ungates the scheduling gate from the Workload Pod to let it scheduled - alongside the Headroom Pods if possible
+  4. Schedules Mountpoint Pod if necessary (i.e., the CSI Driver cannot share an existing Mountpoint Pod) into the same node as the Workload and Headroom Pods using a preempting priority class
+  5. Mountpoint Pod most likely preempts the Headroom Pods if there is no space in the node - as the Headroom Pods uses a negative priority -, or just gets scheduled if there is enough space for all pods
+  6. Deletes the Headroom Pods as soon as the Workload Pod is running or terminated - as Mountpoint Pods are already scheduled or no longer needed
+
+## What are the limitations of this feature?
+
+### Overprovisioning
+
+This feature may cause overprovisioning (i.e., autoscalers may create larger instance types than needed to host pending workloads), which could result in allocating more resources than the cluster requires. This occurs because the CSI Driver allocates one Headroom Pod per volume, but in some cases that capacity may not be needed due to the Mountpoint Pod sharing feature.
```
yerzhan7: > capacity may not be needed due to the Mountpoint Pod sharing feature

So, this also means that another limitation of the headroom approach is that "Mountpoint Pod sharing feature" won't be utilised properly (or even not at all in some scenarios). E.g. when there are 100 workloads scheduled at the same time Karpenter might schedule 100 nodes (with 100 MP Pods) for them instead of e.g. 50 nodes and 50 MP Pods.

Should we document that "Prevents/Undermines Mountpoint Pod Sharing Optimization" as another separate limitation in this section of the doc?

tests/e2e-kubernetes/testsuites/headroom.go
```diff
@@ -0,0 +1,157 @@
...
+		})
+
+		It("should get scheduled automatically after reserving headroom for multiple volumes", func(ctx context.Context) {
```
yerzhan7, line 154: Would be nice to also add a test with 2 workload pods for the MP pod sharing scenario.

tests/e2e-kubernetes/testsuites/headroom.go
```diff
@@ -0,0 +1,157 @@
...
+	return t.tsInfo
+}
+
+func (t *s3CSIHeadroomTestSuite) SkipUnsupportedTests(_ storageframework.TestDriver, pattern storageframework.TestPattern) {
+	if pattern.VolType != storageframework.PreprovisionedPV {
+		e2eskipper.Skipf("Suite %q does not support %v", t.tsInfo.Name, pattern.VolType)
+	}
+}
+
+func (t *s3CSIHeadroomTestSuite) DefineTests(driver storageframework.TestDriver, pattern storageframework.TestPattern) {
```
yerzhan7, line 53: These tests do not actually test with autoscaling schedulers, right? Could we mention this as a code comment?

cmd/aws-s3-csi-controller/csicontroller/reconciler.go
```diff
@@ -119,20 +127,114 @@ func (r *Reconciler) reconcileMountpointPod(ctx context.Context, pod *corev1.Pod
...
+				errs = append(errs, err)
+				continue
+			}
+		} else if needsHeadroomForMountpointPod {
+			// Label the workload pod once for headroom pods
+			if i == 0 {
+				err = r.labelWorkloadPodForHeadroomPod(ctx, pod, log)
+				if err != nil {
+					errs = append(errs, err)
+					continue
```
yerzhan7: It might make more sense to `return` early here with error if we fail to apply label and retry in the next reconciliation, instead of `continue` creating headroom pods?

pkg/podmounter/mppod/headroom.go
```diff
@@ -0,0 +1,170 @@
...
+// to reserve headroom for the Mountpoint Pod to serve volumes to workload.
+//
+// If this scheduling gate is used on a Workload Pod, the CSI Driver:
+//  1. Labels the Workload Pod to use inter-pod affinity rules in the Headroom Pods
+//  2. Creates Headroom Pods using a pause container with inter-pod affinity rule to the Workload Pod
+//  3. Ungates the scheduling gate from the Workload Pod to let it scheduled - alongside the Headroom Pods if possible
+//  4. Schedules Mountpoint Pod if necessary (i.e., the CSI Driver cannot share an existing Mountpoint Pod) into the same node as the Workload and Headroom Pods using a preempting priority class
+//  5. Mountpoint Pod most likely preempts the Headroom Pods if there is no space in the node - as the Headroom Pods uses a negative priority -, or just gets scheduled if there is enough space for all pods
+//  6. Deletes the Headroom Pods as soon as the Workload Pod is running or terminated - as Mountpoint Pods are already scheduled or no longer needed
+const SchedulingGateReserveHeadroomForMountpointPod = "experimental.s3.csi.aws.com/reserve-headroom-for-mppod"
```
yerzhan7: Not sure that we should use prefix `experimental` in labels and scheduling gate names.

E.g. if this feature would become stable in the future we might need to maintain 2 types of labels and support some kind of migration code to ensure we are not breaking anything during stable release.

I think even K8s never uses `experimental` in their labels for experimental features.

cmd/aws-s3-csi-controller/csicontroller/reconciler.go
```diff
@@ -119,20 +127,114 @@ func (r *Reconciler) reconcileMountpointPod(ctx context.Context, pod *corev1.Pod
...
+	err = errors.Join(errs...)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if needsHeadroomForMountpointPod &&
+		len(volumes) == numHeadroomPods {
+		// Spawned all Headroom Pods needed, now ungate the Workload Pod,
+		// so it can get scheduled (alongside Headroom Pods)
+		err = r.ungateHeadroomSchedulingGateForWorkloadPod(ctx, pod, log)
```
yerzhan7, line 208: If headroom pod creation fails, then workload pod will be permanently in gated state, right? I.e. it won't be automatically moved into `Failed` state like before during `ContainerCreation` state. 

Should we add some troubleshooting guide on how to find and delete this stuck gated pods?

pkg/podmounter/mppod/headroom.go
```diff
@@ -0,0 +1,170 @@
...
+	return slices.ContainsFunc(workloadPod.Spec.SchedulingGates, isHeadroomScheduligGate)
+}
+
+// UngateHeadroomSchedulingGateForWorkloadPod removes the [SchedulingGateReserveHeadroomForMountpointPod] scheduling gate from the `workloadPod`.
+func UngateHeadroomSchedulingGateForWorkloadPod(workloadPod *corev1.Pod) {
+	workloadPod.Spec.SchedulingGates = slices.DeleteFunc(workloadPod.Spec.SchedulingGates, isHeadroomScheduligGate)
+}
+
+// isHeadroomScheduligGate returns whether the `sg` equals to [SchedulingGateReserveHeadroomForMountpointPod].
+func isHeadroomScheduligGate(sg corev1.PodSchedulingGate) bool {
```
yerzhan7: `isHeadroomScheduligGate` -> `isHeadroomSchedulingGate`

pkg/driver/node/mounter/pod_mounter.go
```diff
@@ -510,7 +512,11 @@ func (pm *PodMounter) helpMessageForGettingMountpointLogs(pod *corev1.Pod) strin
 }
 
 // helpMessageForGettingMountpointPodStatus returns a help message to throubleshoot if Mountpoint Pod is not running.
-func (pm *PodMounter) helpMessageForGettingMountpointPodStatus(mpPodName string) string {
+func (pm *PodMounter) helpMessageForGettingMountpointPodStatus(err error, mpPodName string) string {
+	if errors.Is(err, watcher.ErrPodUnschedulable) {
+		return fmt.Sprintf("Seems like Mountpoint Pod is in 'Pending' status because it is unschedulable. This usually happens if there is no space for Mountpoint Pod in the node, see %s fore more details and some possible workarounds. You can see Mountpoint Pod's status and any potential failures by running: `kubectl describe pods -n %s %s`", unschedulableMountpointPodReference, mountpointPodNamespace, mpPodName)
```
yerzhan7: `fore more` -> `for more`

pkg/driver/node/mounter/pod_mounter.go
```diff
@@ -510,7 +512,11 @@ func (pm *PodMounter) helpMessageForGettingMountpointLogs(pod *corev1.Pod) strin
 }
 
 // helpMessageForGettingMountpointPodStatus returns a help message to throubleshoot if Mountpoint Pod is not running.
```
yerzhan7: `throubleshoot` -> `troubleshoot`

charts/aws-mountpoint-s3-csi-driver/templates/mount-s3-priority.yaml
```diff
@@ -9,3 +9,34 @@ description: >-
...
+kind: PriorityClass
+metadata:
+  name: mount-s3-preempting-critical
+value: 1000000000
+preemptionPolicy: PreemptLowerPriority
+globalDefault: false
+description: >-
+    Preempting priority class for Mountpoint Pods.
+    It has the highest possible value for non-builtin PriorityClasses to ensure Mountpoint Pods get scheduled quickly and not evicted first in case of a resource pressure.
+    Mountpoint Pods spawned with this priority class by default if "Reserving headroom for Mountpoint Pods" feature is enabled.
```
yerzhan7: `Mountpoint Pods spawned` -> `Mountpoint Pods are spawned`

pkg/podmounter/mppod/headroom.go
unexge: I can only see one experimental label in https://kubernetes.io/docs/reference/labels-annotations-taints/, but there are lots of alpha and beta labels/annotations.

But your point makes sense, I don't expect us to change label/scheduling-gate names between versions, and we can drop experimental there as we already have experimental feature-flag at Helm-level.

cmd/aws-s3-csi-controller/csicontroller/reconciler.go
unexge: I think both is fine, I expect order of things to not matter much in this case. We can create Headroom Pods first, and they would wait until we label the workloads. Since all other error cases just goes to next volume to make some progress, I used the same logic here.

docs/HEADROOM_FOR_MPPOD.md
unexge: Good callout. I mentioned this in "Overprovisioning" section.

docs/HEADROOM_FOR_MPPOD.md
```diff
@@ -0,0 +1,79 @@
...
+
+When the CSI Driver detects a Workload Pod using the scheduling gate to enable this feature, it performs the following steps:
+
+  1. Labels the Workload Pod to use inter-pod affinity rules in the Headroom Pods
+  2. Creates Headroom Pods using a pause container with inter-pod affinity rule to the Workload Pod - since node autoscalers like [Karpenter supports inter-pod affinity rules](https://karpenter.sh/docs/concepts/scheduling/#pod-affinityanti-affinity), this should help them to choose a right instance type
+  3. Ungates the scheduling gate from the Workload Pod to let it scheduled - alongside the Headroom Pods if possible
+  4. Schedules Mountpoint Pod if necessary (i.e., the CSI Driver cannot share an existing Mountpoint Pod) into the same node as the Workload and Headroom Pods using a preempting priority class
+  5. Mountpoint Pod most likely preempts the Headroom Pods if there is no space in the node - as the Headroom Pods uses a negative priority -, or just gets scheduled if there is enough space for all pods
+  6. Deletes the Headroom Pods as soon as the Workload Pod is running or terminated - as Mountpoint Pods are already scheduled or no longer needed
```
yerzhan7, line 24: > Deletes the Headroom Pods as soon as the Workload Pod is running or terminated

Sometimes Reconciler might miss that Workload Pod is terminated due to eventual consistency and we will log `Pod not found - ignoring` from line 78 in reconciler.go.

In this case, Headroom Pods won't be deleted, right?

unexge: Yeah that's true. We also have the same issue with Mountpoint Pods I think, we need to have a periodic cleanup for missed events.

yerzhan7: For stale Mountpoint Pods we have `stale_attachment_cleaner.go`, so maybe we should adjust logic in there to also cleanup Headroom Pods?

yerzhan7 (approved):

PR #548, opened 2025-07-24
Fixes on upgrade test GitHub workflow

unexge:
This is a follow-up PR for https://github.com/awslabs/mountpoint-s3-csi-driver/pull/532 to fix problems with GitHub workflow. Triggered a successful upgrade test from 1.15.0 to 2.0.0-source on https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/16503618110.

---

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -13,6 +13,14 @@ on:
         required: false
         type: boolean
         description: "If true, only upgrade tests will run."
+      upgrade_tests_old_version:
+        required: false
+        type: string
+        description: "Only used if `run_upgrade_tests` is true, specifies the old version to test upgrade from."
+      upgrade_tests_new_version:
+        required: false
+        type: string
+        description: "Only used if `run_upgrade_tests` is true, specifies the new version to test upgrade to."
 
 concurrency:
   group: e2e-cluster-${{ inputs.environment }}
@@ -81,6 +89,8 @@ jobs:
       TAG: "untested_${{ inputs.ref }}"
       K8S_VERSION: "${{ matrix.kubernetes-version }}"
       SELINUX_MODE: "${{ matrix.selinux-mode }}"
+      MOUNTPOINT_CSI_DRIVER_PREVIOUS_VERSION: "${{ inputs.upgrade_tests_old_version }}"
+      MOUNTPOINT_CSI_DRIVER_NEW_VERSION: "${{ inputs.upgrade_tests_new_version }}"
     steps:
       - name: Checkout
         uses: actions/checkout@v4
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -109,7 +109,7 @@ jobs:
         with:
           role-to-assume: ${{ vars.IAM_ROLE }}
           aws-region: ${{ vars.AWS_REGION }}
-          role-duration-seconds: ${{ inputs.run_upgrade_tests && 7200 || 14400 }}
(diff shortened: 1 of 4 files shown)


.github/workflows/e2e-tests.yaml
```diff
@@ -99,7 +109,7 @@ jobs:
         with:
           role-to-assume: ${{ vars.IAM_ROLE }}
           aws-region: ${{ vars.AWS_REGION }}
-          role-duration-seconds: 7200
+          role-duration-seconds: ${{ inputs.run_upgrade_tests && 14400 || 7200 }}
```
muddyfish, line 115: Why do we need double time here?

tests/e2e-kubernetes/scripts/run.sh
```diff
@@ -181,7 +182,7 @@ elif [[ "${ACTION}" == "run_tests" ]]; then
 elif [[ "${ACTION}" == "run_upgrade_tests" ]]; then
   set +e
   pushd tests/e2e-kubernetes
-  KUBECONFIG=${KUBECONFIG} ginkgo -p -vv -timeout 10h -- --bucket-region=${REGION} --commit-id=${TAG} --bucket-prefix=${CLUSTER_NAME} --imds-available=true --cluster-name=${CLUSTER_NAME} --run-upgrade-tests
+  KUBECONFIG=${KUBECONFIG} ginkgo -vv -timeout 10h -- --bucket-region=${REGION} --commit-id=${TAG} --bucket-prefix=${CLUSTER_NAME} --imds-available=true --cluster-name=${CLUSTER_NAME} --run-upgrade-tests
```
muddyfish, line 185: Was the `-p` running in parallel? If so, why can't we anymore?

.github/workflows/e2e-tests.yaml
unexge: Because upgrade tests [run for 2.5+ hours](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/tests/e2e-kubernetes/testsuites/upgrade.go#L26-L34).

tests/e2e-kubernetes/scripts/run.sh
unexge: Yes `-p` is for running tests in parallel, but we only have one test-case with upgrade tests. The thing is Ginkgo tries to buffer the outputs in the parallel mode and won't print output until the test-case is finished, that makes things harder to track, and we don't get any benefit from it.

.github/workflows/e2e-tests.yaml
```diff
@@ -121,6 +131,12 @@ jobs:
         run: |
           tests/e2e-kubernetes/scripts/run.sh
       - name: Run E2E Tests
+        # The CSI Driver v1 doesn't support running on SELinux enabled hosts,
```
unexge, line 137: We can remove this after 2.0.0 release as we wouldn't run an upgrade test from 1.x after that point.

.github/workflows/e2e-tests.yaml
renanmagagnin: could we add a comment with the explanation?

unexge: Added with https://github.com/awslabs/mountpoint-s3-csi-driver/pull/548/commits/e52fa5f1221e34ea4c602972989a593db50f4e15

.github/workflows/e2e-tests.yaml
unexge: Added this as a TODO comment so we don't forget

PR #553, opened 2025-07-30
Simplify GitHub workflow matrix & update some docs

unexge:
Simplifying the GitHub workflow after 2.0.0 release to:
- Add `Ubuntu2204` by default and only exclude on `1.28.13` (as it's unsupported)
- Add `Ubuntu2004` only for `1.28.13` and `1.29.8` (as it's only supported in these versions)
- Enable SELinux in all AL2023 variants.

Also, addressed a TODO in upgrade tests regarding skipping tests with SELinux-enabled variants, and updated some documents regarding feedback from previous PRs.

---

### DIFF
--- a/.github/matrix.yaml
+++ b/.github/matrix.yaml
@@ -1,11 +1,11 @@
 matrix:
   cluster-type: ["eksctl"]
   arch: ["x86", "arm"]
-  family: ["AmazonLinux2", "AmazonLinux2023", "Bottlerocket"]
+  family: ["AmazonLinux2", "AmazonLinux2023", "Bottlerocket", "Ubuntu2204"]
   kubernetes-version:
     ["1.28.13", "1.29.8", "1.30.4", "1.31.0", "1.32.1", "1.33.2"]
   include:
-    # Ubuntu2004 supported for EKS <= 1.29 and Ubuntu2204 supported for EKS >= 1.29.
+    # Ubuntu2004 supported for EKS <= 1.29
     # See https://eksctl.io/usage/custom-ami-support/?h=ubuntu#setting-the-node-ami-family.
     - cluster-type: "eksctl"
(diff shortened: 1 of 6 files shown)


tests/e2e-kubernetes/scripts/run.sh
```diff
@@ -64,7 +64,7 @@ CLUSTER_FILE=${TEST_DIR}/${CLUSTER_NAME}.${CLUSTER_TYPE}.yaml
 SSH_KEY=${SSH_KEY:-""}
 HELM_RELEASE_NAME=mountpoint-s3-csi-driver
 
-EKSCTL_VERSION=${EKSCTL_VERSION:-0.210.0}
+EKSCTL_VERSION=${EKSCTL_VERSION:-0.212.0}
```
yerzhan7: np: Wondering whether we should just always use latest version to avoid these frequent manual updates?

E.g. `EKSCTL_VERSION=${EKSCTL_VERSION:-$(curl -s https://api.github.com/repos/eksctl-io/eksctl/releases/latest | grep '"tag_name":' | cut -d'"' -f4)}`

yerzhan7 (dismissed):

muddyfish (dismissed):

.github/matrix.yaml
```diff
@@ -1,11 +1,18 @@
 matrix:
   cluster-type: ["eksctl"]
   arch: ["x86", "arm"]
-  family: ["AmazonLinux2", "AmazonLinux2023", "Bottlerocket", "Ubuntu2204"]
+  family:
+    [
+      "AmazonLinux2",
```
yerzhan7, line 6: np: I think in YAML we can just do list of strings like this?

```
family:
  - AmazonLinux2
  - AmazonLinux2023
```

PR #554, opened 2025-07-31
Gate `selectableFields` in CRD definition for Kubernetes <1.32

unexge:
[Selectable fields for custom resources](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#crd-selectable-fields) GA'd on Kubernetes 1.32. We [check Kubernetes version at the runtime](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/d9cc082b6c27922f4efb54b742b47841bd9fb590/pkg/cluster/cluster.go#L58-L70) before using this feature, but our Helm template contains this field regardless of the Kubernetes version. 

This could either cause a validation error or a warning ([as in our CI](https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/16598473580/job/46952783952#step:9:151)) like:
```
I0729 14:24:39.116674    4077 warnings.go:110] "Warning: unknown field \"spec.versions[0].selectableFields\""
```

This PR also adds a new validation step to ensure we validate the CSI Driver's Kubernetes manifests on the server-side using strict validation rules. Hoping that this will help us to catch any incompatibilities on the manifests like this in different Kubernetes versions in the future. Also [verified that the new validation step can catch this issue](https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/16650915211/job/47125853890#step:9:220) before applying the fix.

--- 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -195,6 +195,7 @@ TMP_POD_ATTACHMENT_CRD_FILE ?= "./hack/s3.csi.aws.com_mountpoints3podattachments
 generate: controller-gen
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./pkg/api/..."
 	$(CONTROLLER_GEN) crd paths="./pkg/api/..." output:crd:dir=./hack/
+	patch $(TMP_POD_ATTACHMENT_CRD_FILE) < ./hack/patches/selectablefields-k8s-version.patch
 	echo '# Auto-generated file via `make generate`. Do not edit.' > $(HELM_POD_ATTACHMENT_CRD_FILE)
 	cat $(TMP_POD_ATTACHMENT_CRD_FILE) >> $(HELM_POD_ATTACHMENT_CRD_FILE)
 	rm $(TMP_POD_ATTACHMENT_CRD_FILE)
--- a/charts/aws-mountpoint-s3-csi-driver/templates/mountpoints3podattachments-crd.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/mountpoints3podattachments-crd.yaml
@@ -124,8 +124,10 @@ spec:
             - workloadFSGroup
             type: object
         type: object
+    {{- if semverCompare ">=1.32.0-0" .Capabilities.KubeVersion.Version }}
     selectableFields:
     - jsonPath: .spec.nodeName
+    {{- end }}
     served: true
     storage: true
     subresources:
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -195,6 +195,7 @@ TMP_POD_ATTACHMENT_CRD_FILE ?= "./hack/s3.csi.aws.com_mountpoints3podattachments
 generate: controller-gen
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./pkg/api/..."
 	$(CONTROLLER_GEN) crd paths="./pkg/api/..." output:crd:dir=./hack/
+	patch $(TMP_POD_ATTACHMENT_CRD_FILE) < ./hack/patches/selectablefields-k8s-version.patch
 	echo '# Auto-generated file via `make generate`. Do not edit.' > $(HELM_POD_ATTACHMENT_CRD_FILE)
 	cat $(TMP_POD_ATTACHMENT_CRD_FILE) >> $(HELM_POD_ATTACHMENT_CRD_FILE)
```
yerzhan7, line 200: Offtopic: Just noticed that we are missing copying CRD file also to Tests folder: `./tests/crd/mountpoints3podattachments-crd.yaml`. It was probably removed accidentally in previous commits.

charts/aws-mountpoint-s3-csi-driver/templates/mountpoints3podattachments-crd.yaml
```diff
@@ -124,8 +124,10 @@ spec:
             - workloadFSGroup
             type: object
         type: object
+    {{- if semverCompare ">=1.32.0-0" .Capabilities.KubeVersion.Version }}
     selectableFields:
     - jsonPath: .spec.nodeName
+    {{- end }}
```
yerzhan7, line 130: For customers that would upgrade K8s from e.g. 1.31 to 1.33, would they be able to utilize `selectableFields` if they would update CSI Driver afterwards? I.e. does K8s allow adding `selectableFields` to existing CRD version `v2`?

Or would they have to re-install the driver fully. 

I am worried that when customers would update CSI Driver, our code in `IsSelectableFieldsSupported()` would start returning `true` and this will cause error in our Node during listing of CRDs using `nodeName` selector. (Because CRD does not mention that `nodeName` is selectable field)

unexge: Couldn't find any reference but checked the code[1] and also tested in my cluster, and it seems like `selectableFields` is mutable. So, this shouldn't cause any problem.


[1]: There are a few immutable fields on `CustomResourceDefinitionSpec` but `selectableFields` is not one of them: https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/validation/validation.go#L621-L631

Makefile
unexge: Good catch, will address in a follow-up PR. Ideally we shouldn't add this `if` block there as the tests probably expects a plain YAML manifest.

PR #557, opened 2025-08-01
Query CRD definition to decide whether to use `spec.nodeName` filter on the cache

unexge:
A follow-up PR for https://github.com/awslabs/mountpoint-s3-csi-driver/pull/554 to improve our check on deciding whether to utilize `selectableFields` while caching our CRDs. This PR now queries CRD definition at runtime and checks if we have `spec.nodeName` in the `selectableFields` of the current CRD version.

---


pkg/driver/driver.go
```diff
@@ -253,3 +261,30 @@ func setupS3PodAttachmentCache(config *rest.Config, stopCh <-chan struct{}, node
 
 	return s3paCache
 }
+
+// checkIfMountpointS3PodAttachmentHasNodeNameSelectableFieldInCurrentVersion returns whether
+// MountpointS3PodAttachment CRD definition contains `spec.nodeName` as a `selectableField` in its current version.
+func checkIfMountpointS3PodAttachmentHasNodeNameSelectableFieldInCurrentVersion(ctx context.Context, config *rest.Config) (bool, error) {
```
yerzhan7, line 267: How confident we are that this will fix the issue during K8s cluster upgrades? Can we do manual one-off validation?

Also, if customer installs our CRD with `selectableFields` in K8s v1.31 (e.g. ignoring Helm warning) I think K8s will store raw YAML definition "as is", even though it's not aware how to use this `selectableFields` field yet. So, this might cause false positive result in this function which might cause runtime errors for K8s v1.31? (I might be wrong here, but that's why I am saying we need to test/validate this)

Wondering whether fallback mechanism  is better approach here - check first with node name field selector, and then fallback without it if there is error during cache setup?

Or maybe we should also keep the original K8s version check (together with this new check)?

unexge: Verified in 1.28 (by modifying the Helm chart to apply this field regardless of the version and triggering an "unknown field" warning) and 1.33 clusters and got expected logs for both:
For 1.28:
```
I0801 10:46:00.173979       1 driver.go:231] Cluster doesn't support selectable fields, falling back to client-side filtering
```

For 1.33:
```
I0801 13:23:13.314616       1 driver.go:224] Using `spec.nodeName` filter for caching MountpointS3PodAttachment as the cluster supports it
```

I haven't tested the upgrade, will do that separately. But from my understanding, this will prevent the risk of code and Helm chart going out of sync. Now we'll always query source-of-truth to decide whether to use this feature. 

The Kubernetes reference also confirms that it drops unrecognized fields:
> By default, the API server drops fields that it does not recognize from an input that it receives (for example, the JSON body of a PUT request).
https://kubernetes.io/docs/reference/using-api/api-concepts/#field-validation

Which matches with my testing outcome. 


---

I didn't want to add fallback mechanism as it would require wasted requests and would be harder to debug if we silently fail because of another reason. I believe this approach would provide the same guarantees without the downsides.


---

I don't think having version check would provide any other benefits here.

PR #567, opened 2025-08-21
Initial development docs for the CSI Driver

unexge:
- [dev/README.md Rendered](https://github.com/unexge/mountpoint-s3-csi-driver/blob/4bd1c6e95e37331695c67587b2986a2e14e25938/dev/README.md)
- [docs/ARCHITECTURE.md Rendered](https://github.com/unexge/mountpoint-s3-csi-driver/blob/4bd1c6e95e37331695c67587b2986a2e14e25938/docs/ARCHITECTURE.md)

---

### DIFF
--- a/dev/README.md
+++ b/dev/README.md
@@ -219,7 +219,7 @@ See [the architecture document](/docs/ARCHITECTURE.md) of Mountpoint for Amazon 
 
 ## Development
 
-First step would be creating a real Kubernetes cluster, we recommend using an Amazon EKS cluster to test any changes to the CSI Driver. [eksctl](https://eksctl.io/) is the recommended tool to manage EKS clusters.
+First step would be creating a real Kubernetes cluster, we recommend using an Amazon EKS cluster to test any changes to the CSI Driver. [eksctl](https://eksctl.io/) is the recommended tool to manage EKS clusters. This section explains creating a development environment from scratch step-by-step, and it is recommended to follow each step for the first time. For subsequent cluster creations, and day-to-day development, you can use the [helper script](#helper-script).
 
 ### Create a cluster
 
@@ -245,6 +245,8 @@ $ kubectl get nodes
 NAME                                           STATUS   ROLES    AGE   VERSION
 ip-192-168-26-202.eu-north-1.compute.internal   Ready    <none>   13m   v1.33.3-eks-3abbec1
 ```
+
+You can use any region for your development stack, but it is recommended to use a region where [S3 Express One Zone is available](https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-express-Endpoints.html). Here we choose `eu-north-1`.
 
 Now you should also get [Container Insights](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/ContainerInsights.html) enabled for your cluster for metrics and logs.
 
@@ -289,6 +291,8 @@ You can uninstall the Helm chart by running:
 ```bash
 $ helm uninstall -n kube-system aws-mountpoint-s3-csi-driver
 ```
+
+Note that this is only needed if you want to fully uninstall the CSI Driver, if you want to upgrade the existing installation, you don't need to run `helm uninstall`. `helm upgrade` would automatically replace existing version in that case.
 
 ### Building the CSI Driver's container image from source
 
--- a/dev/mp-dev.sh
+++ b/dev/mp-dev.sh
@@ -13,6 +13,7 @@
 # with permissions to create ECR repositories and EKS clusters.
 #
 # This script also uses the following binaries: "aws", "eksctl", "kubectl", "helm" and "docker".
+
 set -euo pipefail
 # Uncomment for debugging
 # set -x


dev/README.md
```diff
@@ -0,0 +1,559 @@
...
+
+Even though the CSI is a generic interface, Mountpoint for Amazon S3 CSI Driver actively only test against [Kubernetes](https://kubernetes.io/).
+All end-to-end tests of the CSI Driver creates [Amazon Elastic Kubernetes Service (EKS)](https://aws.amazon.com/eks/) clusters,
+installs the CSI Driver, and runs the test against the real EKS cluster.
+
+### Kubernetes primer
+
+Kubernetes is a container orchestration system with the following architecture:
+
+![Kubernetes Architecture](https://kubernetes.io/images/docs/components-of-kubernetes.svg)
```
yerzhan7: np: This and next diagram looks bad in dark mode 😅

Screenshot 2025-08-22 at 13 22 15

unexge: Should be better now

dev/README.md
```diff
@@ -0,0 +1,666 @@
...
+2025-08-20 16:40:22 [ℹ]  setting availability zones to [eu-north-1b eu-north-1c eu-north-1a]
+...
+2025-08-20 16:57:04 [ℹ]  kubectl command should work with "/Users/burakvar/.kube/config", try 'kubectl get nodes'
+2025-08-20 16:57:04 [✔]  EKS cluster "mp-dev-cluster" in "eu-north-1" region is ready
+```
+
+It might take around 15-20 minutes to create the cluster, at the end it should configure your kubectl to access to your cluster.
+You can verify this by running:
+```bash
+$ kubectl get no
```
renanmagagnin: np: maybe we could change to `kubectl get nodes` for consistency and clarity since we haven't mentioned the shorthands up to this point.

dev/README.md
```diff
@@ -0,0 +1,666 @@
...
+
+The [binding process](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#binding) of a PVC to a PV depends on the [provisioning mode](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#provisioning), which can be static or dynamic. With static provisoninig, the cluster admins would create PVs representing available storage to the cluster, and PVCs would only be able to claim existing PVs. With dynamic provisioning, the cluster admins would instead create [StorageClasses](https://kubernetes.io/docs/concepts/storage/storage-classes/) which basically act as templates for PVs, and the cluster would automatically provision PVs based on StorageClass definition and PVC requests. The CSI Driver only supports [static provisinonig](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/docs/CONFIGURATION.md#static-provisioning) as of `v2.0.0`.
+
+For example, this is how you'd use the CSI Driver to mount `amzn-s3-demo-bucket` bucket at `/data`:
```
renanmagagnin, line 124: We mention `/data` here but the manifest mounts at `/cache`

dev/mp-dev.sh
```diff
@@ -0,0 +1,266 @@
+#!/usr/bin/env bash
```
renanmagagnin, line 1: I think we could add a short explanation at the top of the file with an overview of the script. What it can be used for and an summary of the features.

Makefile
```diff
@@ -57,7 +57,7 @@ E2E_COMMIT_ID?=local
 E2E_KUBECONFIG?=""
 
 # Kubernetes version to use in envtest for controller tests.
-ENVTEST_K8S_VERSION ?= 1.30.x
```
renanmagagnin, line 60: Out of curiosity since it seems unrelated, why did we make this change?

dev/README.md
```diff
@@ -0,0 +1,666 @@
...
+$ kubectl get pods -lapp.kubernetes.io/name=aws-mountpoint-s3-csi-driver -A
+NAMESPACE     NAME                                 READY   STATUS    RESTARTS   AGE
+kube-system   s3-csi-controller-6b4cbf9ddf-nm74l   1/1     Running   0          54s
+kube-system   s3-csi-node-dfmhd                    3/3     Running   0          54s
+```
+
+You can uninstall the Helm chart by running:
+```bash
+$ helm uninstall -n kube-system aws-mountpoint-s3-csi-driver
+```
```
renanmagagnin, line 293: Maybe we could add a note here saying that this is unnecessary when using upgrade (since it overwrites the current).

dev/README.md
```diff
@@ -0,0 +1,666 @@
...
+~/mountpoint-s3-csi-driver
+
+$ make test
+...
+Ran 55 of 55 Specs in 159.773 seconds
+SUCCESS! -- 55 Passed | 0 Failed | 0 Pending | 0 Skipped
+--- PASS: TestControllers (159.77s)
+PASS
+ok      github.com/awslabs/mountpoint-s3-csi-driver/tests/controller    160.335s
+```
```
renanmagagnin: I remember you saying these might be incorrect so leaving this comment as a reminder

dev/mp-dev.sh
```diff
@@ -0,0 +1,266 @@
...
+: "${MOUNTPOINT_CSI_DEV_ECR_REPO_NAME:=mp-dev}"
+: "${MOUNTPOINT_CSI_DEV_CLUSTER_NAME:=mp-dev-cluster}"
+
+usage() {
+    echo "usage: $0 <command>"
+    echo ""
+    echo "available commands:"
+    echo "  setup                    create ECR repository, EKS cluster, and deploy everything"
+    echo "  teardown                 delete ECR repository and EKS cluster"
+    echo "  deploy-helm-chart        deploy the Helm chart from source"
```
renanmagagnin: how about we make it explicit that we deploy from source using "ECR" or the "private ECR repo"? I think it would make it easier to see the connection with deploy-containers. We could make sure the order in which they are mentioned here is the order in which we would use them in the most common use case.

dev/mp-dev.sh
```diff
@@ -0,0 +1,266 @@
...
+    echo "  deploy-containers        build the containers from source, push to ECR, and restart the Csi Driver pods"
```
renanmagagnin:
```suggestion
    echo "  deploy-containers        build the containers from source, push to ECR, and restart the CSI Driver pods"
```

renanmagagnin: nit: I wonder if we should call them container images instead of just containers to avoid confusion

dev/mp-dev.sh
```diff
@@ -0,0 +1,266 @@
...
+            echo "  $var"
+        done
+        echo ""
+        exit 1
+    fi
+}
+
+get_aws_account_id() {
+    aws sts get-caller-identity --query Account --output text
+}
```
renanmagagnin: nit: remove the function and replace the only occurrence with the function body

dev/mp-dev.sh
```diff
@@ -0,0 +1,266 @@
...
+
+get_ecr_repository_url() {
+    local repo_name="${MOUNTPOINT_CSI_DEV_ECR_REPO_NAME}"
+    local region="${MOUNTPOINT_CSI_DEV_REGION}"
```
renanmagagnin: nit: can we avoid the redeclaration of these two in all ecr repository related functions?

dev/mp-dev.sh
renanmagagnin, line 1: Also, what are the pre-requisites to use the script? Valid AWS credentials?

dev/mp-dev.sh
```diff
@@ -0,0 +1,266 @@
...
+            --region "${region}" \
+            --force
+        echo "ECR repository '${repo_name}' deleted successfully"
+    else
+        echo "ECR repository '${repo_name}' does not exist"
+    fi
+}
+
+create_eks_cluster() {
+    local cluster_name="${MOUNTPOINT_CSI_DEV_CLUSTER_NAME}"
```
renanmagagnin: nit: can we avoid the redeclaration of these?

dev/README.md
```diff
@@ -0,0 +1,666 @@
...
+This document and the `dev/` folder contains references, documentation, and various things for developing the CSI Driver. Note that this document is written with v2 in mind, most things should also apply to v1 but there might be slight differences.
+
+## Background
+
+The CSI Driver implements [Container Storage Interface (CSI)](https://github.com/container-storage-interface/spec/blob/v1.9.0/spec.md) to provide Amazon S3 Buckets as volumes for containers using [Mountpoint for Amazon S3](https://github.com/awslabs/mountpoint-s3) under the hood.
+Mountpoint is based on Filesystem in Userspace (FUSE) and allows you to mount your buckets as local file systems.
+
+Even though the CSI is a generic interface, Mountpoint for Amazon S3 CSI Driver actively only test against [Kubernetes](https://kubernetes.io/).
+All end-to-end tests of the CSI Driver creates [Amazon Elastic Kubernetes Service (EKS)](https://aws.amazon.com/eks/) clusters,
+installs the CSI Driver, and runs the test against the real EKS cluster.
```
renanmagagnin:
```suggestion
All end-to-end tests of the CSI Driver create [Amazon Elastic Kubernetes Service (EKS)](https://aws.amazon.com/eks/) clusters,
install the CSI Driver, and run the test against the real EKS cluster.
```

dev/README.md
```diff
@@ -0,0 +1,666 @@
...
+  - [etcd](https://kubernetes.io/docs/concepts/architecture/#etcd) – the key value store used as Kubernetes' backing store for all cluster data
+  - [kube-scheduler](https://kubernetes.io/docs/concepts/architecture/#kube-scheduler) – the default scheduler that assigns pending pods to suitable nodes
+  - [kube-controller-manager](https://kubernetes.io/docs/concepts/architecture/#kube-controller-manager) – the component that runs controllers
+
+The nodes contains components:
+  - [kubelet](https://kubernetes.io/docs/concepts/architecture/#kubelet) – the component that runs on each node that is responsible for running pods/containers and communicating with the control plane
+  - [Container runtime](https://kubernetes.io/docs/concepts/architecture/#container-runtime) – the runtime responsible for running the containers
+
+Cluster admins would usually interact with the cluster using tools like [kubectl](https://kubernetes.io/docs/reference/kubectl/). kubectl allows you to define your [Kubernetes objects](https://kubernetes.io/docs/concepts/overview/working-with-objects/#kubernetes-objects) as YAML files, and submit these manifests to the control plane.
```
renanmagagnin:
```suggestion
Cluster admins would usually interact with the cluster using tools like [kubectl](https://kubernetes.io/docs/reference/kubectl/), which allows you to define your [Kubernetes objects](https://kubernetes.io/docs/concepts/overview/working-with-objects/#kubernetes-objects) as YAML files, and submit these manifests to the control plane.
```

dev/README.md
```diff
@@ -0,0 +1,666 @@
...
+
+These Kubernetes objects defines the desired state, and once submitted to the control plane, multiple [Kubernetes controllers](https://kubernetes.io/docs/concepts/architecture/controller/) would try to make the current state match to the desired state.
```
renanmagagnin: ### 
```suggestion
These Kubernetes objects define the desired state, and once submitted to the control plane, multiple [Kubernetes controllers](https://kubernetes.io/docs/concepts/architecture/controller/) would try to make the current state match to the desired state.
```

Makefile
unexge: I just realized we're using an old version of Kubernetes control plane and just updated while writing the docs. This is not used anywhere except dev environments. In the CI, we're using our full matrix to test against all Kubernetes version we support.

dev/README.md
unexge: This was for showing how it can be uninstalled rather than uninstalling it for the next installation.

dev/mp-dev.sh
```diff
@@ -0,0 +1,262 @@
+#!/usr/bin/env bash
+#
+# mp-dev.sh provides utilities for day-to-day development of the CSI Driver.
+#
+# It allows developers to setup a development environment (an ECR repository and an EKS cluster) with "setup" command,
+# and also allows termination of the deveopment environment with "teradown" command.
```
renanmagagnin:
```suggestion
# and also allows termination of the development environment with "teardown" command.
```

dev/mp-dev.sh
```diff
@@ -0,0 +1,262 @@
...
+#
+# It also allows developers to deploy Helm chart and containers from source
+# with "deploy-helm-chart" and "deploy-containers" commands respectively.
+#
+# == Requirements ==
+# This script requires a functioning AWS CLI that can access to an AWS account
+# with permissions to create ECR repositories and EKS clusters.
+#
+# This script also uses the following binaries: "aws", "eksctl", "kubectl", "helm" and "docker".
+set -euo pipefail
```
renanmagagnin: nit: add space between these two

dev/README.md
```diff
@@ -0,0 +1,666 @@
...
+    kubelet->>apiSrv: 8. look for newly assigned pod(s)
+    apiSrv->>kubelet: 9. bind pod to node
+    kubelet->>container: 10. start container
+    kubelet->>apiSrv: 11. update pod status
+    apiSrv-->>etcd: 12. save new state
+```
+
+#### Kubernetes Storage
+
+Kubernetes provide both persistent and temporary storage to the Pods via [volumes](https://kubernetes.io/docs/concepts/storage/volumes/). Pods can use any number of persistent or temporary volumes simultaneously. Temporary or [ephemeral volumes](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/) usually have a lifecycle tied to the Pods, and gets deleted alongise the Pod. [Persistent volumes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/) on the other hand can outlive Pods, and can be shared among multiple Pods if the [access mode](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes) of the volume allows. Kubernetes won't destroy persistent volumes at Pod termination, it is up to cluster admins to manually delete them or use external tools to manage their lifecycles.
```
renanmagagnin:
```suggestion
Kubernetes provide both persistent and temporary storage to the Pods via [volumes](https://kubernetes.io/docs/concepts/storage/volumes/). Pods can use any number of persistent or temporary volumes simultaneously. Temporary or [ephemeral volumes](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/) usually have a lifecycle tied to the Pods, and get deleted alongside the Pod. [Persistent volumes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/) on the other hand can outlive Pods, and can be shared among multiple Pods if the [access mode](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes) of the volume allows. Kubernetes won't destroy persistent volumes at Pod termination, it is up to cluster admins to manually delete them or use external tools to manage their lifecycles.
```

dev/README.md
```diff
@@ -0,0 +1,666 @@
...
+
+The volumes basically represented as directories within Pods, you specify volumes to use in the Pod definition, and then you can mount these volumes within containers at the desired path. As a result, the containers would have a filesystem with the desired volumes mounted.
```
renanmagagnin:
```suggestion
The volumes are basically represented as directories within Pods, you specify volumes to use in the Pod definition, and then you can mount these volumes within containers at the desired path. As a result, the containers would have a filesystem with the desired volumes mounted.
```

dev/README.md
```diff
@@ -0,0 +1,666 @@
...
+
+For example, this is how you'd use a [emptyDir](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir) – an ephemeral volume created from the kubelet's backing storage medium:
```
renanmagagnin:
```suggestion
For example, this is how you'd use an [emptyDir](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir) – an ephemeral volume created from the kubelet's backing storage medium:
```

dev/README.md
```diff
@@ -0,0 +1,666 @@
...
+    name: busybox
+    # Mounting the volume `cache-volume` in the container at `/cache`
+    volumeMounts:
+    - mountPath: /cache
+      name: cache-volume
+```
+
+Kubernetes would automatically create the volume with 500MiB limit, and would also automatically terminate alongside the Pod.
+
+Kubernetes provides many kind of volume types out of the box, but it also provides support for [out-of-tree volume plugins](https://kubernetes.io/docs/concepts/storage/volumes/#out-of-tree-volume-plugins) – which most importantly allows CSI Drivers to provide new volume types without needing to change Kubernetes source. A CSI Driver volume can be used through three different way:
```
renanmagagnin:
```suggestion
Kubernetes provides many kinds of volume types out of the box, but it also provides support for [out-of-tree volume plugins](https://kubernetes.io/docs/concepts/storage/volumes/#out-of-tree-volume-plugins) – which most importantly allows CSI Drivers to provide new volume types without needing to change Kubernetes source. A CSI Driver volume can be used through three different ways:
```

dev/README.md
```diff
@@ -0,0 +1,666 @@
...
+  - Through a [PersistentVolume](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistent-volumes) referenced by a [`persistentVolumeClaim`](https://kubernetes.io/docs/concepts/storage/volumes/#persistentvolumeclaim)
+  - Through a [generic ephemeral volumes](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes)
+  - Through a [CSI ephemeral volume](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#csi-ephemeral-volumes)
+
+Last two as you can guess from their names aims to provide ephemeral volumes where the lifecycle of the volume would be tied to a Pod. The first one aims to provide persistent volumes which is more suitable for Amazon S3 and it is the only volume type the CSI Driver supports as of `v2.0.0`.
```
renanmagagnin:
```suggestion
The last two, as you can guess from the names, aim to provide ephemeral volumes where the lifecycle of the volume would be tied to a Pod. The first one aims to provide persistent volumes which is more suitable for Amazon S3 and it is the only volume type the CSI Driver supports as of `v2.0.0`.
```

dev/README.md
```diff
@@ -0,0 +1,666 @@
...
+
+The [binding process](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#binding) of a PVC to a PV depends on the [provisioning mode](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#provisioning), which can be static or dynamic. With static provisoninig, the cluster admins would create PVs representing available storage to the cluster, and PVCs would only be able to claim existing PVs. With dynamic provisioning, the cluster admins would instead create [StorageClasses](https://kubernetes.io/docs/concepts/storage/storage-classes/) which basically act as templates for PVs, and the cluster would automatically provision PVs based on StorageClass definition and PVC requests. The CSI Driver only supports [static provisinonig](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/docs/CONFIGURATION.md#static-provisioning) as of `v2.0.0`.
+
+For example, this is how you'd use the CSI Driver to mount `amzn-s3-demo-bucket` bucket at `/cache`:
```
renanmagagnin:
```suggestion
For example, this is how you'd use the CSI Driver to mount `amzn-s3-demo-bucket` bucket at `/data`:
```

dev/README.md
```diff
@@ -0,0 +1,666 @@
...
+spec:
+  volumes:
+  - name: s3-volume
+    persistentVolumeClaim:
+      claimName: s3-pvc
+  containers:
+  - image: busybox
+    name: busybox
+    volumeMounts:
+    - mountPath: /cache
```
renanmagagnin:
```suggestion
    - mountPath: /data
```

dev/README.md
```diff
@@ -0,0 +1,666 @@
...
+  - [Identity Service](https://github.com/container-storage-interface/spec/blob/v1.9.0/spec.md#identity-service-rpc) – For providing healthcheck, plugin version and set of features provided by the CSI Driver
+  - [Node Service](https://github.com/container-storage-interface/spec/blob/v1.9.0/spec.md#node-service-rpc) – For doing mounting in the node by communicating with kubelet
+  - [Controller Service](https://github.com/container-storage-interface/spec/blob/v1.9.0/spec.md#controller-service-rpc) – For doing dynamic provisioning, volume publishing etc. by communicating with the control plane
+
+As of `v2.0.0`, the CSI Driver only implements [node](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/v2.0.0/pkg/driver/node/node.go) and [identity](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/v2.0.0/pkg/driver/identity.go) services.
+
+For node service, the most important RPC calls the CSI Driver implement are `NodePublishVolume` and `NodeUnpublishVolume`. They are called by kubelet whenever a Pod uses a volume provided by the CSI Driver, and the calls are basically performs mounting and unmounting an S3 volume to a pod respectively.
```
renanmagagnin:
```suggestion
For node service, the most important RPC calls the CSI Driver implements are `NodePublishVolume` and `NodeUnpublishVolume`. They are called by kubelet whenever a Pod uses a volume provided by the CSI Driver, and the calls basically perform mounting and unmounting of an S3 volume to a pod, respectively.
```

dev/README.md
```diff
@@ -0,0 +1,666 @@
...
+
+#### `NodePublishVolume`
+
+In this RPC call, the CSI Driver receives details of the volume and a target path to mount the filesystem.
+
+Most importantly, the CSI Driver receives `mountOptions` and `volumeAttributes` of the PV. These are free-form fields that the CSI Driver free to interpret or design. The CSI Driver passes `mountOptions` to Mountpoint process as-is mostly, and uses information from `volumeAttributes` to decide on credentials, cache configuration etc., and at the ends converts them to more Mountpoint options. See [`volumecontext` Go package](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/v2.0.0/pkg/driver/node/volumecontext/volume_context.go) for set of configuration the CSI Driver accepts in `volumeAttributes`.
```
renanmagagnin:
```suggestion
Most importantly, the CSI Driver receives `mountOptions` and `volumeAttributes` of the PV. These are free-form fields that the CSI Driver is free to interpret or design. The CSI Driver passes `mountOptions` to Mountpoint process as-is mostly, and uses information from `volumeAttributes` to decide on credentials, cache configuration etc., and at the ends converts them to more Mountpoint options. See [`volumecontext` Go package](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/v2.0.0/pkg/driver/node/volumecontext/volume_context.go) for set of configuration the CSI Driver accepts in `volumeAttributes`.
```

dev/mp-dev-cluster.yaml
```diff
@@ -0,0 +1,69 @@
+apiVersion: eksctl.io/v1alpha5
+kind: ClusterConfig
+
+metadata:
+  name: mp-dev-cluster
+  region: eu-north-1
```
renanmagagnin, line 6: nit: maybe we could add a comment that this should be a region that supports express

dev/README.md
```diff
@@ -0,0 +1,666 @@
...
+
+In this RPC call, the CSI Driver basically reverts the operations performed in the `NodePublishVolume` call. Namely, the CSI Driver unmounts the filesystem provided for the Pod, and then cleans up any written credentials.
+
+### Architecture of Mountpoint for Amazon S3 CSI Driver
+
+See [the architecture document](/docs/ARCHITECTURE.md) of Mountpoint for Amazon S3 CSI Driver for more details on the design of the CSI Driver.
+
+## Development
+
+First step would be creating a real Kubernetes cluster, we recommend using an Amazon EKS cluster to test any changes to the CSI Driver. [eksctl](https://eksctl.io/) is the recommended tool to manage EKS clusters.
```
renanmagagnin: Not sure but maybe we should mention the helper script here. I guess the goal could be to explain that these are detailed explanation but, as long as the steps are understood, the helper script can be used for convenience. This should be enough to avoid having first time readers follow the steps before learning about the script.

dev/README.md
renanmagagnin: I know, I just meant we could add the note/tip to avoid the misunderstanding that it might be necessary every time we want to reinstall.

dev/README.md
```diff
@@ -0,0 +1,666 @@
...
+            "scanOnPush": false
+        },
+        "encryptionConfiguration": {
+            "encryptionType": "AES256"
+        }
+    }
+}
+```
+
+Now that we have a registry, we can now build our container image and push there.
```
renanmagagnin:
```suggestion
Now that we have a registry, we can now build our container image and push it there.
```

dev/README.md
```diff
@@ -0,0 +1,666 @@
...
+# "image.pullPolicy=Always" disables this caching behaviour and ensures Kubernetes would always pull the latest image each time it starts the CSI Driver's containers.
+$ helm upgrade --install aws-mountpoint-s3-csi-driver \
+    --namespace kube-system \
+    --set image.repository="111122223333.dkr.ecr.eu-north-1.amazonaws.com/mp-dev" \
+    --set image.pullPolicy=Always \
+    --set image.tag=latest \
+   ./charts/aws-mountpoint-s3-csi-driver
+```
+
+Each time after you push a new version of the container image, you'd need to restart the Pods for changes to take effect. You can verify your Pods has the up-to-date container image by checking the hash:
```
renanmagagnin:
```suggestion
Each time after you push a new version of the container image, you'd need to restart the CSI Driver Pods for changes to take effect. You can verify your Pods has the up-to-date container image by checking the hash:
```

dev/README.md
```diff
@@ -0,0 +1,666 @@
...
+  help                     show this help message
+
+required environment variables:
+  MOUNTPOINT_CSI_DEV_ECR_REPOSITORY  ECR repository URL (e.g., 111122223333.dkr.ecr.eu-north-1.amazonaws.com/mp-dev)
+  MOUNTPOINT_CSI_DEV_REGION          AWS region for the dev stack (e.g., eu-north-1)
+```
+
+### Running tests
+
+Ideally you would want to run tests before or after making any change. The CSI Driver has three different test suites:
```
renanmagagnin:
```suggestion
Ideally you would want to run tests before and after making any change. The CSI Driver has three different test suites:
```

dev/README.md
```diff
@@ -0,0 +1,666 @@
...
+```
+
+These tests by default downloads and runs against 1.33 version of Kubernetes control plane, and in [the CI they run against our whole version matrix](/.github/workflows/controller-tests.yaml).
+
+See [Writing controller tests](https://book.kubebuilder.io/cronjob-tutorial/writing-tests) and
+[Configuring envtest for integration tests](https://book.kubebuilder.io/reference/envtest) for more details on controller tests.
+
+#### End-to-end tests
+
+The CSI Driver has a end-to-end test suite as well. These tests run against real Kubernetes clusters with [a various Kubernetes version and host matrix in the CI](/.github/matrix.yaml) for each PR.
```
renanmagagnin:
```suggestion
The CSI Driver has an end-to-end test suite as well. These tests run against real Kubernetes clusters with [a matrix in the CI of various Kubernetes versions and hosts](/.github/matrix.yaml) for each PR.
```

dev/mp-dev-cluster.yaml
unexge: This is not a requirement unless you want to do something with S3 Express, but I choose a region with S3 Express support. But I can add a comment to README to suggest picking a region with S3 Express support

Issue #543, opened 2025-07-18
Scheduler problems with v2


unexge:
## Problem

The CSI Driver v2 spawns Mountpoint Pods into the same nodes as the workloads to provide volumes. The CSI Driver spawns these Mountpoint Pods _after_ the workloads are assigned to nodes. This approach is taken because we might be able to [share an existing Mountpoint Pod](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/docs/MOUNTPOINT_POD_SHARING.md) on a specific node, eliminating the need to spawn a new one. Therefore, the CSI Driver needs to wait until a workload is scheduled to a specific node before deciding whether to share an existing Mountpoint Pod or spawn a new one.

Due to this dynamic spawning of Mountpoint Pods on specific nodes, a problem emerges with schedulers like [Kubernetes Scheduler](https://kubernetes.io/docs/concepts/scheduling-eviction/kube-scheduler/) or [Karpenter Scheduler](https://karpenter.sh/docs/concepts/scheduling/). Since schedulers aren't aware of the Mountpoint Pod requirements while making decisions for the workloads, they might make suboptimal scheduling decisions. As a result, there might not be enough space for a Mountpoint Pod on a specific node where the workload is deployed. In such cases, the Mountpoint Pod, and consequently the workload pod, would remain in a `Pending` state. Eventually, the workload pod would transition into the `Failed` state if the Mountpoint Pod can't be scheduled to the node.

### Reserving headroom for Mountpoint Pods

We're working on a feature to automatically reserve some headroom for Mountpoint Pods to help with this issue for node autoscalers like Karpenter. The CSI Driver spawns Headroom Pods with inter-pod affinity rules to their corresponding Workload Pods. These Headroom Pods are deployed using a pause container with a negative priority – making them the first target in case of an eviction. The CSI Driver also assigns a preempting priority class to Mountpoint Pods – the default priority class is a non-preempting one if this feature is not enabled – with the goal of Mountpoint Pods preempting Headroom Pods if needed. 

See https://github.com/awslabs/mountpoint-s3-csi-driver/pull/536 fore more details on limitations of this workaround, and how to enable this feature if that's suitable.

### Using a high-priority preempting priority class for Mountpoint Pods

[...]

PR #573, opened 2025-09-02
feat: support deploying the node DaemonSet with hostNetwork enabled

rodrigorfk:
*Issue #, if available:*

*Description of changes:*
Currently, the helm chart does not allow changing the the `hostNetwork` configuration for the node DaemonSet and for certain use cases, it will be interesting to do so as many users running EKS clusters with standard amazon-vpc-cni setups are keen to reduce the amount of IPs needed by DaemonSets, a common practices in this case is to enable `hostNetwork` mode for trustworthy daemons. As a side requirement of that, users might have to change the port where the healthz `liveness-probe` container is running, so it will avoid conflicts with other daemons listening on the same port, for example the https://github.com/kubernetes-sigs/aws-ebs-csi-driver also listens on the default `9808` port. 

### DIFF
--- a/charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
@@ -29,6 +29,7 @@ spec:
         {{- end }}
       serviceAccountName: {{ .Values.node.serviceAccount.name }}
       priorityClassName: system-node-critical
+      hostNetwork: {{ .Values.node.hostNetwork }}
       {{- with .Values.node.affinity }}
       affinity: {{- toYaml . | nindent 8 }}
       {{- end }}
@@ -130,7 +131,7 @@ spec:
               mountPropagation: Bidirectional
           ports:
             - name: healthz
-              containerPort: 9808
+              containerPort: {{ .Values.node.healthzPort }}
               protocol: TCP
           livenessProbe:
             httpGet:
@@ -188,6 +189,7 @@ spec:
             allowPrivilegeEscalation: false
           args:
             - --csi-address=/csi/csi.sock
+            - --health-port={{ .Values.node.healthzPort }}
           volumeMounts:
             - name: plugin-dir
               mountPath: /csi
--- a/charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
@@ -39,6 +39,7 @@ spec:
         {{- end }}
       serviceAccountName: {{ .Values.node.serviceAccount.name }}
       priorityClassName: system-node-critical
+      hostNetwork: {{ .Values.node.hostNetwork }}
       {{- with .Values.node.affinity }}
       affinity: {{- toYaml . | nindent 8 }}
       {{- end }}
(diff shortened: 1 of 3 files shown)


tadiwa-aizen (changes requested):
Thanks for the contribution, @rodrigorfk! This is a clean, well-scoped change.

One small ask before we merge: could you add documentation for the two new Helm values (`node.hostNetwork` and `node.healthzPort`) in `docs/CONFIGURATION.md`? A short section noting what they do and the IP-savings / port-conflict use case would help users discover the feature. Thanks again!

rodrigorfk:
Thank you @tadiwa-aizen, the documentation was updated.

tadiwa-aizen (changes requested):
Hi @rodrigorfk — thanks for this! I ran it by our EKS team. No objection to the feature, but a few asks before we merge:

**1) Set a `dnsPolicy` for the hostNetwork case.** The template sets `hostNetwork` but no `dnsPolicy`. EKS flagged that hostNetwork pods silently fall back to the node's `/etc/resolv.conf` unless `ClusterFirstWithHostNet` is set — which breaks resolving in-cluster names. Please auto-set `dnsPolicy: ClusterFirstWithHostNet` when `hostNetwork` is `true` (or expose a `node.dnsPolicy` value), and document it. [Docs..](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy) 

**2) Document the IMDS security trade-off.** EKS noted that `hostNetwork: true` bypasses the IMDS hop-limit=1 protection — the pod can always reach IMDS and assume the node's IAM role. Please add a doc warning so users opt in knowingly.

rodrigorfk:
Done @tadiwa-aizen, your request was addressed.

tadiwa-aizen (changes requested):
Thanks for this  @rodrigorfk — we've just reviewed it as a team and we're good to move forward. Two changes before merge:

1) Document the port failure mode: if `healthzPort` conflicts with a port already in use on the node, the liveness probe can't bind and the node pod restart-loops — surfaces at rollout, so confirm the port is free.

2) Remove the `dnsPolicy` stuff (the block in `node.yaml`, the value in `values.yaml`, and the DNS section in `CONFIGURATION.md`). We checked the node's code and it doesn't resolve any in-cluster DNS, so it's not needed here. Apologies for the churn.

Everything else (hostNetwork default false, healthzPort, IMDS warning) is good as-is. 
We've also decided to handle testing.

rodrigorfk:
Hi @tadiwa-aizen, no problem at all, the `dnsPolicy` work was removed and the por failure mode documented as request. Please take another look.

tadiwa-aizen (dismissed):
LGTM

tadiwa-aizen:
Tested on my personal EKS cluster and everything works as expected. The pod IP matches the node's internal IP and, as expected, the health check on 9808 binds to the host. I also ran the static provisioning YAML and the Mountpoint pods spun up and worked.

tadiwa-aizen:
Hi @rodrigorfk — thank you for this, and for your patience through the review.

We're actively working on a new daemonset architecture that reduces the number of pods — [a single shared mounter DaemonSet instead of per-mount Mountpoint Pods](https://github.com/awslabs/mountpoint-s3-csi-driver/issues/543#issuecomment-4305014764) — which we believe addresses this. So we won't be merging this one, but we'll re-evaluate after that architecture launches.

Thanks again for the work.

Issue #579, opened 2025-09-11
mount-s3-namespace.yaml under helm chart conditional


harshashenoy98:
/feature

**Is your feature request related to a problem? Please describe.**
Make mount-s3-namespace.yaml under helm chart conditional.
Currently, this yaml file gets deployed without condition. We want it to have a condition that by default it can deploy or create the namespace, but if we choose to opt out of it, then don't deploy.

**Describe the solution you'd like in detail**
Currently, this yaml file gets deployed without condition. We want it to have a condition that by default it can deploy or create the namespace, but if we choose to opt out of it, then don't deploy.

As we use ArgoCD to deploy it, the argo application autocreates the namespace along with certain annotations that we pass from ArgoCD. But this yaml file overrides these annotations causing the s3-csi-controller deployment to have issues.

**Describe alternatives you've considered**
PR - https://github.com/awslabs/mountpoint-s3-csi-driver/pull/580

**Additional context**

PR #580, opened 2025-09-11
Add `mountpointPod.createNamespace` flag to Helm chart

harshashenoy98:
*Issue #, if available:*
https://github.com/awslabs/mountpoint-s3-csi-driver/issues/579

*Description of changes:*
Currently, this yaml file gets deployed without condition. We want it to have a condition that by default it can deploy or create the namespace, but if we choose to opt out of it, then don't deploy.

As we use ArgoCD to deploy it, the argo application autocreates the namespace along with certain annotations that we pass from ArgoCD. But this yaml file overrides these annotations causing the s3-csi-controller deployment to have issues


charts/aws-mountpoint-s3-csi-driver/values.yaml
```diff
@@ -88,6 +88,7 @@ controller:
 
 mountpointPod:
   namespace: mount-s3
+  namespaceEnabled: true
```
muddyfish: I'd change the name to `createNamespace` - the namespace is still required to use the driver, but the use case for this flag is assuming it's pre-configured

dannycjones:
Hey @harshashenoy98. We'd like to integrate this change. Would you like to continue owning it until completion, addressing the feedback in the PR?

Alternatively, we can take ownership of getting this into `main`.

charts/aws-mountpoint-s3-csi-driver/values.yaml
unexge: It's also worth noting that `mount-s3` namespace created by the CSI Driver enforces [“restricted” pod security standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted):

https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/charts/aws-mountpoint-s3-csi-driver/templates/mount-s3-namespace.yaml#L7 

The CSI Driver's service account gets permissions to create pods in this namespace, and without this restriction it would get higher privileges. Not necessarily a problem but it's worth noting that and suggesting having same standards on the namespaces created outside.

dannycjones: I've updated the parameter name and added a comment on the Helm chart value.

https://github.com/awslabs/mountpoint-s3-csi-driver/pull/580/files#diff-4805cfed36c125bd1835f4357120cbc78fd3f6f539439f61ab103fde9023f440R91-R94

muddyfish (approved):

Issue #584, opened 2025-09-22
Failed to create mountpoint pod if volumeHandle is `s3://amzn-s3-demo-bucket`


yerzhan7:
/kind bug
> **_NOTE:_**  If this is a filesystem related bug, please take a look at the [Mountpoint repo](https://github.com/awslabs/mountpoint-s3) to submit a [bug report](https://github.com/awslabs/mountpoint-s3/issues/new?assignees=&labels=bug&projects=&template=bug-report.yml)

**What happened?** Failed to create MP Pod with volumehandle `s3://amzn-s3-demo-bucket`

```
“error”:“Pod \“mp-XXXX\” is invalid: metadata.labels: Invalid value: \“s3://amzn-s3-demo-bucket\“: a valid label must be an empty string or consist of alphanumeric characters, ‘-’, ‘_’ or ‘.’, and must start and end with an alphanumeric character (e.g. ‘MyValue’,  or ‘my_value’,  or ‘12345’, regex used for validation is ‘(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?‘)“
```

**What you expected to happen?** Create MP Pod

**How to reproduce it (as minimally and precisely as possible)?** Use volumehandle `s3://amzn-s3-demo-bucket`

**Anything else we need to know?**: Affects only v2.0.0

**Environment**
- Kubernetes version (use `kubectl version`): N/A
- Driver version: v2.0.0

PR #585, opened 2025-09-22
Move MP Pod VolumeID and PV name to Annotations instead of Labels

yerzhan7:
*Issue #, if available:* https://github.com/awslabs/mountpoint-s3-csi-driver/issues/584

*Description of changes:*

If customers use very long `volumeHandle` or `name` in the PV (or use invalid chars like `s3://amzn-s3-demo-bucket`) then MP Pod creation fails due to error.

Labels value have the following limits: 63 char max, Must start with letter/number, can contain -, _, ., letters and numbers

Whereas Annotations have no such limits (max size is 256KB). 

Solution: Move from Labels to Annotations.

### DIFF
--- a/pkg/driver/node/mounter/pod_unmounter.go
+++ b/pkg/driver/node/mounter/pod_unmounter.go
@@ -227,8 +227,14 @@ func (u *PodUnmounter) writeExitFile(podPath string) error {
 
 // cleanupCredentials removes credentials associated with the Mountpoint Pod
 func (u *PodUnmounter) cleanupCredentials(mpPod *corev1.Pod) error {
+	volumeID := mpPod.Annotations[mppod.AnnotationVolumeId]
+	if volumeID == "" {
+		// Fallback to legacy label for backward compatibility with old MP pods (v2.0.0)
+		volumeID = mpPod.Labels[mppod.DeprecatedLabelVolumeId]
+	}
+
 	return u.credProvider.Cleanup(credentialprovider.CleanupContext{
-		VolumeID:  mpPod.Labels[mppod.LabelVolumeId],
+		VolumeID:  volumeID,
 		PodID:     string(mpPod.UID),
 		WritePath: mppod.PathOnHost(u.podPath(string(mpPod.UID)), mppod.KnownPathCredentials),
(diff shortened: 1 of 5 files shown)


pkg/driver/node/mounter/pod_unmounter.go
```diff
@@ -227,8 +227,14 @@ func (u *PodUnmounter) writeExitFile(podPath string) error {
 
 // cleanupCredentials removes credentials associated with the Mountpoint Pod
 func (u *PodUnmounter) cleanupCredentials(mpPod *corev1.Pod) error {
+	volumeID := mpPod.Annotations[mppod.AnnotationVolumeId]
```
vladem: I'd suggest using:

```go
volumeID, exists := mpPod.Annotations[mppod.AnnotationVolumeId]
```

Also, what happens if neither the annotation nor the label for `VolumeID` exists? will we fail to clean up creds?

tests/controller/controller_test.go
```diff
@@ -984,6 +984,28 @@ var _ = Describe("Mountpoint Controller", func() {
 			waitForObjectToDisappear(mountpointPod.Pod)
 		})
 
+		It("should set AnnotationVolumeId with long volumeHandle in mountpoint pod", func() {
```
vladem, line 987: this test actually creates a pod with an annotation, right? it doesn't use a mock or something?

yerzhan7: These are controller tests - they test against local Kubernetes API server. So, they create MP Pods in there.

pkg/driver/node/mounter/pod_unmounter.go
yerzhan7: > I'd suggest using:

Fixed. Thanks for reviewing!

> Also, what happens if neither the annotation nor the label for VolumeID exists? 

In theory, this should never happen as we always assign either label or annotation to MP Pod via our controller.

> will we fail to clean up creds?

The cleanup code will just silently succeed as the token file with empty string volumeID does not exist. Which is still okay because the volume itself will be removed anyway by K8s after MP Pod is deleted.

vladem (approved):

Issue #566, opened 2025-08-19
MountPoint NodeAffinity conflicts with Karpenter spotToSpotConsolidation


korjwl1:
/kind bug
> **_NOTE:_**  If this is a filesystem related bug, please take a look at the [Mountpoint repo](https://github.com/awslabs/mountpoint-s3) to submit a [bug report](https://github.com/awslabs/mountpoint-s3/issues/new?assignees=&labels=bug&projects=&template=bug-report.yml)

**What happened?**
Updated mountpoint to v2, and now pods are created in the 'mount-s3' namespace.
The problem is, the pods are created with nodeAffinity - matchFields that targets specific node.
Karpenter does not work with the matchFields when consolidating, thus keep output such logs
`{"level":"DEBUG","time":"2025-08-19T01:14:25.163Z","logger":"controller","caller":"lo@v1.47.0/slice.go:501","message":"ignoring pod, node selector term with matchFields is not supported","commit":"058c665","controller":"provisioner","namespace":"","name":"","reconcileID":"759e9bc2-c1fb-48b2-bd7b-2c4e4d7ea14e","Pod":{"name":"mp-jtg8k","namespace":"mount-s3"}}`
Until the specific pod is removed, Karpenter stops consolidating nodes.

**What you expected to happen?**
Ideally, Karpenter should be able to move the pod to a new node for the successful consolidation.

**How to reproduce it (as minimally and precisely as possible)?**
I turned on the spotToSpotConsolidation feature on the Karpenter version 1.2.1.
The problem occurred when Karpenter tried to change the node to a cheaper one. (m5 -> t3)

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`): v1.31.4-eks-0f56d01
- Driver version: 2.0.0 (helm chart version)

PR #587, opened 2025-09-23
Assigns pods to nodes directly instead of using affinity

Amitabitbul:
Switches from node affinity with label matching to explicit node assignment for pod scheduling. This simplifies pod placement and avoids reliance on node labels.

This pull request updates the scheduling logic for Mountpoint Pods in the `pkg/podmounter/mppod/creator.go` file. Instead of using node affinity with labels, it now directly assigns the pod to the target node using the `NodeName` field. This change simplifies pod scheduling and ensures more reliable placement.

Pod scheduling improvements:

* Replaced node affinity configuration with direct assignment via the `NodeName` field in the Mountpoint Pod spec, ensuring pods are scheduled on the intended node without relying on node labels.

#566


muddyfish:
Closing as stale

PR #605, opened 2025-10-17
Add Update Sidecar Images periodic job

yerzhan7:
*Issue #, if available:* Need to automate updating default values of sidecar images 

*Description of changes:* Add periodic GitHub action to check latest versions and update if needed, and create PR.

- Run script locally - seems to work. I will test in GitHub actions via manual trigger.


dannycjones (commented):
Does Dependabot not work if we configure it? https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#package-ecosystem-

I do note that it says only Helm V3 is supported. However, maybe that means we should upgrade?

yerzhan7:
Dependabot only updates Helm dependencies (`helm dependency update`). We do not have any helm dependencies.

PR #617, opened 2025-10-30
Add env var for EKS Pod Identity token audience and service principal 

dannycjones:
*Issue #, if available:* N/A

*Description of changes:*

For testing in internal EKS environments, we may wish to modify the token audience and service principal used. This change allows it to be overridden in the node service and end-to-end tests.

The default value is unchanged, and remains `"pods.eks.amazonaws.com"`.

We may consider making changes to the Helm chart, however I'm leaving it out for this PR.

### DIFF
--- a/tests/e2e-kubernetes/testsuites/credentials.go
+++ b/tests/e2e-kubernetes/testsuites/credentials.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	goerrors "errors"
 	"fmt"
+	"os"
 	"slices"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


tests/e2e-kubernetes/testsuites/credentials.go
```diff
@@ -86,6 +87,16 @@ type s3CSICredentialsTestSuite struct {
 	tsInfo storageframework.TestSuiteInfo
 }
 
+func determineServiceAccountTokenAudienceEKS() string {
```
muddyfish, line 90: Lets follow up on removing the duplication here

dannycjones: Agree, I will follow up next week.

PR #623, opened 2025-11-03
Update eksctl_delete_cluster_cf_stack to disable termination protection

muddyfish:
*Issue #, if available:* N/A

*Description of changes:*
Update the eksctl delete cluster script to support deleting stacks with termination protection enabled.

### DIFF
--- a/tests/e2e-kubernetes/scripts/eksctl.sh
+++ b/tests/e2e-kubernetes/scripts/eksctl.sh
@@ -124,9 +124,11 @@ function eksctl_delete_cluster() {
   REGION=${3}
   FORCE=${4:-false}
 
+  STACK_NAME="eksctl-${CLUSTER_NAME}-cluster"
   if ! eksctl_cluster_exists "${BIN}" "${CLUSTER_NAME}"; then
     # Try to delete CloudFormation stack even if the cluster does not exists just in case
     # if the stack is stuck in `ROLLBACK_COMPLETE` status.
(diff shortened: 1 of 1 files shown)


tests/e2e-kubernetes/scripts/eksctl.sh
```diff
@@ -124,9 +124,11 @@ function eksctl_delete_cluster() {
   REGION=${3}
   FORCE=${4:-false}
 
+  STACK_NAME="eksctl-${CLUSTER_NAME}-cluster"
   if ! eksctl_cluster_exists "${BIN}" "${CLUSTER_NAME}"; then
     # Try to delete CloudFormation stack even if the cluster does not exists just in case
     # if the stack is stuck in `ROLLBACK_COMPLETE` status.
+    aws cloudformation update-termination-protection --no-enable-termination-protection --region "${REGION}" --stack-name "${STACK_NAME}"
```
dannycjones, line 131: Why do we run this twice in both branches? Why not just once at the start of this function?

muddyfish: There's the alternative case on [line 135](https://github.com/awslabs/mountpoint-s3-csi-driver/pull/623/files#diff-a5fec6510aa2af6d4e903d20fbfbbb6d85aee2494408cfcc25debcd34d7820e3R137) where we skip cluster deletion. I don't think it matters too much if we remove termination protection there, but I don't think it's worth changing this PR for

dannycjones (approved):

PR #652, opened 2025-12-01
Clear pod attachment expectation during handling of inactive pod

yerzhan7:
*Issue #, if available:*

*Description of changes:* When several workloads sharing the same Mountpoint Pod are scheduled concurrently,
and the first workload terminates before the pod attachment is created, the other workloads
may wait indefinitely for a creation that will never complete with error messages:

```
Failed to find corresponding MountpointS3PodAttachment custom resource
MountpointS3PodAttachment creation is pending, requeuing
```

*Solution*: Clear pod attachment expectation during handling of inactive pod (i.e. during deletion of MountpointS3PodAttachment in reconciler)

*Manual Testing:* Reproduced the issue using new e2e test case (need to run around 10-15 times on average to hit the race condition). And verified that the fix fixed the issue:

```
...
Spawning Mountpoint Pod <----- event for "first workload is scheduled"
Mountpoint Pod spawned
MountpointS3PodAttachment is created
Successfully removed workload pod UID from MountpointS3PodAttachment <----- event for "first workload is no longer active"
Mountpoint pod has zero workload UIDs. Adding s3.csi.aws.com/needs-unmount annotation
Mountpoint pod has zero workload UIDs. Will remove it from MountpointS3PodAttachment
MountpointS3PodAttachment has zero Mountpoint Pods. Will delete it
Deleted MountpointS3PodAttachment as it no longer has any attached Mountpoint Pods
Clearing stale creation expectation for deleted MountpointS3PodAttachment
...
Spawning Mountpoint Pod <----- second workload successfully spawns MP Pod and creates Pod attachment
...
```

*Alternative Solutions considered:*
 - Use fixed name for Pod Attachments - this might help by avoiding having expectation logic and using resource name as de-duplication mechanism
    - Cons: big change, requires backward compatibility logic with existing pod attachments
 - Configure TTL for `s3paExpectations` in-memory map to avoid indefinite wait
    - Cons: Does not fix the underlying race condition and increases pod creation time

### DIFF
--- a/cmd/aws-s3-csi-controller/csicontroller/reconciler.go
+++ b/cmd/aws-s3-csi-controller/csicontroller/reconciler.go
@@ -43,7 +43,24 @@ const (
 type Reconciler struct {
 	mountpointPodConfig  mppod.Config
 	mountpointPodCreator *mppod.Creator
-	s3paExpectations     *expectations
+
+	// s3paExpectations tracks MountpointS3PodAttachments created by this reconciler to handle eventual consistency.
+	//
+	// Problem: After creating a MountpointS3PodAttachment (S3PA), it may not immediately appear in the reconciler's
+	// informer cache because the cache relies on watch events from the API server (eventual consistency).
+	// This can cause the reconciler to miss newly created attachments, resulting in lost pod sharing opportunities.
+	//
+	// Solution: s3paExpectations maintains an in-memory map of pod attachments created by this process, allowing
+	// the reconciler to know about them before they appear in the cache.
+	//
+	// Management: We carefully manage expectations to avoid indefinite waiting:
+	// - setPending() is called in handleNewS3PodAttachment() immediately after successfully creating a new S3PA
+	// - clear() is called in two scenarios:
+	//   1. In handleExistingS3PodAttachment() when a pending S3PA is found (it appeared in the cache)
+	//   2. In removeWorkloadFromS3PodAttachment() when deleting an S3PA with a stale pending expectation
+	//
+	// Note: Reconcile() processes events sequentially, eliminating concurrency concerns.
+	s3paExpectations *expectations
 	client.Client
 }
 
@@ -299,7 +316,7 @@ func (r *Reconciler) spawnOrDeleteMountpointPodIfNeeded(
 	log := r.setupLogger(ctx, workloadPod, pvc, workloadUID, fieldFilters, s3pa)
 
 	if !isPodActive(workloadPod) {
-		return r.handleInactivePod(ctx, s3pa, workloadUID, log)
+		return r.handleInactivePod(ctx, s3pa, workloadUID, fieldFilters, log)
 	}
 
 	if s3pa != nil {
@@ -406,13 +423,13 @@ func (r *Reconciler) getExistingS3PodAttachment(ctx context.Context, fieldFilter
 }
 
 // handleInactivePod handles inactive workload pod.
-func (r *Reconciler) handleInactivePod(ctx context.Context, s3pa *crdv2.MountpointS3PodAttachment, workloadUID string, log logr.Logger) (bool, error) {
+func (r *Reconciler) handleInactivePod(ctx context.Context, s3pa *crdv2.MountpointS3PodAttachment, workloadUID string, fieldFilters client.MatchingFields, log logr.Logger) (bool, error) {
 	if s3pa == nil {
 		log.Info("Workload pod is not active. Did not find any MountpointS3PodAttachments.")
 		return DontRequeue, nil
 	}
 
-	return r.removeWorkloadFromS3PodAttachment(ctx, s3pa, workloadUID, log)
+	return r.removeWorkloadFromS3PodAttachment(ctx, s3pa, workloadUID, fieldFilters, log)
 }
 
 // handleExistingS3PodAttachment handles existing S3 Pod Attachment.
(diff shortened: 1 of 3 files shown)


cmd/aws-s3-csi-controller/csicontroller/reconciler.go
vladem, line 445: usage of `r.s3paExpectations` really confused me at first, could you add an explanation of why it is needed please (as comment to `handleExistingS3PodAttachment` method or to the `Reconciler::s3paExpectations` field)? something like this would've helped me in understanding:

> Resources created via k8s API follow the eventual consistency model, meaning that after creation of `MountpointS3PodAttachment` it may not immediately show up in k8s API list responses. Missing an existing pod attachment may result in a missed opportunity for the pod sharing. To mitigate that we use `s3paExpectations` which keeps track of pod attachments created by this process.
>
> We manage `s3paExpectations` carefully to avoid indefinitely waiting for a pod attachment. The key invariant is that once a pod attachment is deleted, it is also removed from expectations. As of now, we don't account for cases when pod attachments are deleted outside of our code and thus this should be avoided.
>
> Note that the reconciler processes one event at a time, so no concurrent calls to this method.

vladem: Also probably I would add a TODO to note the idea about the usage of resource name as a deduplication mechanism, it actually feels more robust (e.g. it accounts for deletions outside of our code).

cmd/aws-s3-csi-controller/csicontroller/reconciler.go
```diff
@@ -618,6 +618,16 @@ func (r *Reconciler) removeWorkloadFromS3PodAttachment(ctx context.Context, s3pa
 			log.Error(err, "Failed to delete MountpointS3PodAttachment")
 			return Requeue, err
 		}
+		log.Info("Deleted MountpointS3PodAttachment as it no longer has any attached Mountpoint Pods")
+		if r.s3paExpectations.isPending(fieldFilters) {
+			// Clear stale creation expectations to prevent indefinite waiting.
+			// Race condition: When several workloads sharing the same Mountpoint Pod are scheduled concurrently,
```
vladem: This reads like there is still a race condition, but I think this is not the case. May be we can reduce it to the first sentence of this comment? In the test we could add details about this issue. 

Also any other places in code where we delete a pod attachment? Do they also require clearing of `s3paExpectations`?

tests/e2e-kubernetes/testsuites/pod_sharing.go
```diff
@@ -357,6 +357,51 @@ func (t *s3CSIPodSharingTestSuite) DefineTests(driver storageframework.TestDrive
...
+			// Create second pod on the same node
+			ginkgo.By(fmt.Sprintf("Creating second pod on node %s", targetNode))
+			pod2 := e2epod.MakePod(f.Namespace.Name, nil, []*v1.PersistentVolumeClaim{resource.Pvc}, admissionapi.LevelBaseline, "")
+			pod2.Spec.NodeName = targetNode
+			pod2, err = createPodWithoutWaiting(ctx, f.ClientSet, f.Namespace.Name, pod2)
+			framework.ExpectNoError(err)
+
+			// Wait for second pod to reach Running status
+			ginkgo.By("Waiting for second pod to reach Running status")
+			err = e2epod.WaitForPodNameRunningInNamespace(ctx, f.ClientSet, pod2.Name, f.Namespace.Name)
```
vladem, line 408: This is where the test was failing, right? You had to run the test many times in a loop to reproduce?

Also just for clarity: you mention concurrently created pods as an issue, but this test created pods sequentially. I think the actual trigger is the immediate deletion of the pods just after creation.

vladem (commented):
A couple of suggestions for comments, otherwise LGTM.

(I would also look into using pod attachment name as a deduplication mechanism in future)

tests/e2e-kubernetes/testsuites/pod_sharing.go
dannycjones: I had some questions here actually. Could we just do a loop of create and deletes to make this test more likely to hit the case that's currently failing?

cmd/aws-s3-csi-controller/csicontroller/reconciler.go
dannycjones: Is it actually feasible to de-dup using the resource name? I'm worried about K8s resource name length limits.

yerzhan7: Fixed resource name was originally considered (with name as hash of the fields), but was discarded. I do not recall exact reasons why, but it does have it's own edge cases. Overall, expectations is a common pattern in K8s controller due to this Informer cache consistency issue - see https://ahmet.im/blog/controller-pitfalls/#expectations-pattern.

Will update the comments, thanks for reviewing!

tests/e2e-kubernetes/testsuites/pod_sharing.go
yerzhan7: > I think the actual trigger is the immediate deletion of the pods just after creation.

Yes - that is true.

> Could we just do a loop of create and deletes to make this test more likely to hit the case that's currently failing?

I would not do it because if we run this in a loop many times (took 10 times in my cluster on average to hit the edge case) the test duration will increase by at least 20min because sometimes background stale MP pod cleanup takes another 2-3 min. If we would skip pod cleanup waiting, then we would quickly hit max-pods-per-node limit.

cmd/aws-s3-csi-controller/csicontroller/reconciler.go
yerzhan7: We also delete pod attachment in stale pod attachment cleaner but that one does not require clearing of expectation because it only deletes pod attachment when stale workloads are removed.

Issue #577, opened 2025-09-07
Failing to mount in microk8s path 108 characters limit v2


oriolcx:
/kind bug

NOTE: Not a filesystem bug in Mountpoint itself; this is about the CSI node plugin’s kubelet path handling.

**What happened?**
On Ubuntu's MicroK8s (snap install), kubelet’s root directory is long (/var/snap/microk8s/common/var/lib/kubelet). The node-driver-registrar places the registration socket that exceeds the Linux AF_UNIX pathname limit (107 bytes), so the registrar fails to bind the socket and the driver never registers with kubelet.

**What you expected to happen?**
The CSI node plugin should register successfully with kubelet on MicroK8s without manual reconfiguration of kubelet’s --root-dir.

**How to reproduce it (as minimally and precisely as possible)?**
Ubuntu 22.04.4 LTS
```
snap install microk8s --classic --channel=1.32
sudo usermod -a -G microk8s $USER
su - $USER
microk8s config > ~/.kube/config
helm repo add aws-mountpoint-s3-csi-driver https://awslabs.github.io/mountpoint-s3-csi-driver
helm repo update
helm upgrade --install aws-mountpoint-s3-csi-driver \
    --namespace kube-system \
    --version 2.0.0 \
    --set node.kubeletPath=/var/snap/microk8s/common/var/lib/kubelet aws-mountpoint-s3-csi-driver/aws-mountpoint-s3-csi-driver

Then, just mount a path in kubernetes. It worked without issues with v1.5, but it seems to fail with v2.0
```

**Anything else we need to know?**:
This appears to be a path-length interaction between MicroK8s’ snap root dir and the CSI registration socket naming scheme.

**Environment**
- Kubernetes version:
Client Version: v1.33.4
Kustomize Version: v5.6.0
Server Version: v1.32.3
- Driver version:
-2.0.0

muddyfish:
Hi - thanks for spending the time to make bug report.

Whilst we don't officially support microk8s, I see this can impact other k8s implementations. 
I'll keep this open for tracking purposes

tadiwa-aizen:
We've addressed this issue here. https://github.com/awslabs/mountpoint-s3-csi-driver/pull/656

PR #656, opened 2025-12-04
Use hard-coded kubelet path for s3-plugin container

tadiwa-aizen:
*Issue #577 *

**Description**

On MicroK8s (snap install), kubelet uses `/var/snap/microk8s/common/var/lib/kubelet` as its root directory. In CSI Driver v2, Unix socket paths built from this base exceed Linux's 108-byte `UNIX_PATH_MAX` limit, causing bind() / connect() to fail with invalid argument. The driver either fails to register with kubelet or fails to communicate with Mountpoint pods, making S3 mounts impossible. This affects any Kubernetes distribution with a long kubelet path.

**Fix**
Hard-code the container-side mountPath to `/var/lib/kubelet `regardless of the host path. The host path remains 
configurable via `node.kubeletPath` (used as the volume source), but inside the container all paths are short.

Replace the single `KUBELET_PATH` env var with:
- `CONTAINER_KUBELET_PATH`=`/var/lib/kubelet `— path as seen inside the container
- `HOST_KUBELET_PATH` — path as seen on the host (from node.kubeletPath)

Add `KubeletHostPathToContainerPath()` to translate host paths (received from kubelet RPCs) to container paths in both `NodePublishVolume` and `NodeUnpublishVolume`.

**Verification**
Tested on Ubuntu 22.04 with MicroK8s (snap, channel 1.32) and node.kubeletPath=/var/snap/microk8s/common/var/lib/kubelet.

Without the fix, the driver fails to register as reported in Issue #577 — the registration socket path exceeds Linux's 108-byte limit.

With the fix — driver registers successfully and all 3 node containers are running:
```
$ microk8s kubectl get pods -n kube-system | grep s3-csi
s3-csi-controller-558c76774b-4mc8f         1/1     Running   0          5m
s3-csi-node-spmkn                          3/3     Running   0          5m
```

```
$ microk8s kubectl logs -n kube-system -l app=s3-csi-node -c node-driver-registrar
I0322 14:02:26.423604       1 main.go:145] "Version" version="v2.16.0-eksbuild.1"
I0322 14:02:26.428943       1 node_register.go:56] "Starting Registration Server" socketPath="/registration/s3.csi.aws.com-reg.sock"
I0322 14:02:26.429167       1 node_register.go:66] "Registration Server started" socketPath="/registration/s3.csi.aws.com-reg.sock"
I0322 14:02:26.965157       1 main.go:102] "Received NotifyRegistrationStatus call" status="plugin_registered:true"
```

S3 mount works — workload pod writes to the bucket:
```
$ microk8s kubectl exec s3-app -- ls -la /data/
$ aws s3 ls tadiwaom-s3-test/
...
2026-03-22 21:30:34         26 Sun Mar 22 21:30:32 UTC 2026.txt
```

**Testing**
Manual Testing with E2E tests passing: (rosa was not set up in my CI) 
https://github.com/tadiwa-aizen/mountpoint-s3-csi-driver/actions/runs/23405969976


charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
```diff
@@ -72,15 +72,13 @@ spec:
...
-            - name: KUBELET_PATH
-              value: {{ .Values.node.kubeletPath }}
+              value: unix:/var/lib/kubelet/plugins/s3.csi.aws.com/csi.sock
             - name: CSI_NODE_NAME
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
             - name: HOST_PLUGIN_DIR
-              value: {{ trimSuffix "/" .Values.node.kubeletPath }}/plugins/s3.csi.aws.com/
+              value: /var/lib/kubelet/plugins/s3.csi.aws.com/
```
dannycjones, line 91: Does it seem reasonable to just remove this environment variable at this point?

charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
```diff
@@ -72,15 +72,13 @@ spec:
             - --v={{ .Values.node.logLevel }}
           env:
             - name: CSI_ENDPOINT
-              value: unix:{{ trimSuffix "/" .Values.node.kubeletPath }}/plugins/s3.csi.aws.com/csi.sock
```
dannycjones, line 85: I was going to suggest removing `CSI_ENDPOINT` variable, but I think it's actually important to keep to stay in compliance with the spec. https://github.com/container-storage-interface/spec/blob/master/spec.md?plain=1#L3449

charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
```diff
@@ -140,7 +138,7 @@ spec:
             - name: ADDRESS
               value: /csi/csi.sock
             - name: DRIVER_REG_SOCK_PATH
-              value: {{ trimSuffix "/" .Values.node.kubeletPath }}/plugins/s3.csi.aws.com/csi.sock
```
dannycjones, line 155: Why does this work?

dannycjones (commented):
I'll keep digging a little into the comments added, nice to see a small fix plus validation!

charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
dannycjones: As far as I understand it, this is the path that should be given to `kubelet` so that it knows how to communicate with the CSI driver. This happens outside the containers, and so won't be impacted by our rebinding.

https://github.com/kubernetes-csi/node-driver-registrar/blob/a2dc54ee9b035e1abe906b146e0c1cd0138f1c57/README.md?plain=1#L48-L51

I note as well that this particular path won't be available to this container, so that is a data point backing this up.

charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
dannycjones: I am interested in removing it just because it's ANOTHER place to configure the kubelet path.

dannycjones: Given it could be part of our public interface (unlikely), I think we need to leave the Go code that references it in place. However, I do think it would be worth dropping the environment variable specified here.

charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
dannycjones: @tadiwa-aizen and I were looking at the microk8s installation provided by snap, it creates a symlink from `/var/lib/kubelet` to `var/snap/microk8s/common/var/lib/kubelet` which explains why this would work at the moment.

I still think we should revert the change to this environment variable in particular, since the argument represents the path of kubelet on the host.

charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
```diff
@@ -109,7 +107,7 @@ spec:
             {{- end }}
           volumeMounts:
             - name: kubelet-dir
-              mountPath: {{ .Values.node.kubeletPath }}
+              mountPath: /var/lib/kubelet
```
dannycjones, line 126: One doubt I have is that `NodePublishVolume` will include a target path in its request body. This path will be on the host, if I understand correctly.

Why then would this targetPath work if we don't mount the actual kubelet path in the same location?

dannycjones: This doesn't work. The mount silently fails (mounts in some random location inside the mp pod) and the S3 bucket doesn't show up inside the container. :(

dannycjones: I think we need to update the code to fail mounting here, it is NOT okay that the code today just logs as a failure but still proceeds.

```
E1204 20:21:05.166731   39937 node.go:102] NodePublishVolume: target path "/var/snap/microk8s/common/var/lib/kubelet/pods/43e60fc7-0771-4380-8ffb-b344c42019cd/volumes/kubernetes.io~csi/s3-pv/mount" is not in kubelet path "/var/lib/kubelet". This might cause mounting issues, please ensure you have correct kubelet path configured.
I1204 20:21:05.166756   39937 node.go:148] NodePublishVolume: mounting danny-bucket at /var/snap/microk8s/common/var/lib/kubelet/pods/43e60fc7-0771-4380-8ffb-b344c42019cd/volumes/kubernetes.io~csi/s3-pv/mount with options [--allow-delete --allow-root --prefix=some-s3-prefix/ --region=eu-west-1]
```

https://github.com/awslabs/mountpoint-s3-csi-driver/blob/796ba8b9ea90bba2119b36072b8a94dafde6a86b/pkg/driver/node/node.go#L101-L103

charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
```diff
@@ -79,15 +79,17 @@ spec:
...
-              value: {{ .Values.node.kubeletPath }}
+              value: unix:/var/lib/kubelet/plugins/s3.csi.aws.com/csi.sock
             - name: CSI_NODE_NAME
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
             - name: HOST_PLUGIN_DIR
-              value: {{ trimSuffix "/" .Values.node.kubeletPath }}/plugins/s3.csi.aws.com/
+              value: /var/lib/kubelet/plugins/s3.csi.aws.com/
+            - name: KUBELET_PATH
```
muddyfish: Should this be renamed? It's not obvious how this differs from the other KUBELET_PATH

pkg/driver/node/node.go
```diff
@@ -98,8 +98,11 @@ func (ns *S3NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePubl
 		return nil, status.Error(codes.InvalidArgument, "Target path not provided")
 	}
 
+	// Translate target path from host format to container format if needed
+	target = util.TranslateKubeletPath(target)
```
muddyfish: Can we use good variable names for this - it'll be confusing in the future

pkg/driver/node/node_test.go
```diff
@@ -46,7 +46,7 @@ func TestNodePublishVolume(t *testing.T) {
 				Mode: csi.VolumeCapability_AccessMode_MULTI_NODE_MULTI_WRITER,
 			},
 		}
-		targetPath = "/target/path"
+		targetPath = "/var/lib/kubelet/target/path"
```
muddyfish, line 51: Why does this need changing?

pkg/util/kubelet.go
```diff
@@ -13,3 +16,24 @@ func KubeletPath() string {
...
+	hostPath := os.Getenv("KUBELET_PATH_HOST")
+	containerPath := KubeletPath()
+
+	// If no host path is configured, or paths are the same, no translation needed
+	if hostPath == "" || hostPath == containerPath {
+		return path
+	}
+
+	// If the path starts with the host kubelet path, replace it with the container path
+	if strings.HasPrefix(path, hostPath) {
```
muddyfish, line 48: And if it doesn't? Is this problematic?

charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
tadiwa-aizen: How about?..
```
- name: KUBELET_PATH
  value: /var/lib/kubelet  # ← Container path

- name: HOST_KUBELET_PATH
  value: {{ .Values.node.kubeletPath }} # ← Host path
```

pkg/driver/node/node.go
tadiwa-aizen: Noted, how is:
```
translatedTarget := util.TranslateKubeletPath(target) ...
```
?

pkg/driver/node/node_test.go
tadiwa-aizen: The PR itself adds validation that target paths must start with the kubelet path (/var/lib/kubelet). The old test path /target/path would fail this validation, so it needs to be updated to /var/lib/kubelet/target/path to match how kubelet actually passes paths.

pkg/util/kubelet.go
tadiwa-aizen: If it doesn't have the prefix, we return the path unchanged (line 38). This is not problematic because it means the path is already in the correct format for the container.

charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
muddyfish: Why not `CONTAINER_KUBELET_PATH`?

pkg/driver/node/node.go
muddyfish: Why not `targetContainer` and `targetHost`?

pkg/util/kubelet.go
```diff
@@ -13,3 +16,24 @@ func KubeletPath() string {
 	}
 	return kubeletPath
 }
+
+// TranslateKubeletPath translates a path from host kubelet path to container kubelet path.
+// This is useful when the kubelet path on the host is different from the kubelet path in the container.
+// For example, in MicroK8s, the host path is /var/snap/microk8s/common/var/lib/kubelet
+// but the container sees it as /var/lib/kubelet.
+func TranslateKubeletPath(path string) string {
```
muddyfish: What about one of these `KubeletHostPathToContainerPath`/`KubeletPathFromHostToContainer`?

pkg/util/kubelet.go
muddyfish: And what if it's not in the correct format? What validation is done?

tadiwa-aizen: If it's not in the correct format, it gets returned unchanged to the caller function `NodePublishVolume` in `pkg/driver/node/node.go`, which then immediately  validates it on line 101-103.

For example...
- Invalid path `/tmp/random/path `→ doesn't match host prefix → returned as `/tmp/random/path`
- Caller receives `/tmp/random/path` and validates: does it start with /`var/lib/kubelet?` → No
- Returns error: "Target path "`/tmp/random/path`" is not in kubelet path "`/var/lib/kubelet`"..."

charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
```diff
@@ -147,7 +149,7 @@ spec:
             - name: ADDRESS
               value: /csi/csi.sock
             - name: DRIVER_REG_SOCK_PATH
-              value: {{ trimSuffix "/" .Values.node.kubeletPath }}/plugins/s3.csi.aws.com/csi.sock
+              value: /var/lib/kubelet/plugins/s3.csi.aws.com/csi.sock
```
muddyfish: Is this correct? Is Danny's comment still relevant?

pkg/util/kubelet.go
```diff
@@ -1,15 +1,39 @@
...
 
-import "os"
+import (
+	"os"
+	"strings"
+)
 
 const defaultKubeletPath = "/var/lib/kubelet"
 
 // KubeletPath returns path of the kubelet.
```
muddyfish: Should this be renamed to `ContainerKubeletPath`?

pkg/driver/node/node.go
```diff
@@ -93,13 +93,16 @@ func (ns *S3NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePubl
...
 		return nil, status.Error(codes.InvalidArgument, "Target path not provided")
 	}
 
-	if !strings.HasPrefix(target, kubeletPath) {
-		klog.Errorf("NodePublishVolume: target path %q is not in kubelet path %q. This might cause mounting issues, please ensure you have correct kubelet path configured.", target, kubeletPath)
+	// Translate target path from host format to container format if needed
+	targetContainer := util.KubeletHostPathToContainerPath(targetHost)
+
+	if !strings.HasPrefix(targetContainer, kubeletPath) {
+		return nil, status.Errorf(codes.InvalidArgument, "Target path %q is not in kubelet path %q. Please ensure you have correct kubelet path configured.", targetContainer, kubeletPath)
```
muddyfish: This error message may now be confusing for users. Can we improve it?

pkg/util/kubelet.go
```diff
@@ -1,15 +1,39 @@
...
 	}
 	return kubeletPath
 }
+
+// KubeletHostPathToContainerPath translates a path from host kubelet path to container kubelet path.
+// This is useful when the kubelet path on the host is different from the kubelet path in the container.
+// For example, in MicroK8s, the host path is /var/snap/microk8s/common/var/lib/kubelet
+// but the container sees it as /var/lib/kubelet.
+func KubeletHostPathToContainerPath(path string) string {
+	hostPath := os.Getenv("HOST_KUBELET_PATH")
```
muddyfish: This seems problematic - we've effectively renamed `KubeletPath` to `ContainerKubeletPath`, but the function was meant to make sure that it works even if `KUBELET_PATH` was undefined. This used to map to the host's kubelet path, so `KubeletPath` should probably have actually become `HostKubeletPath`. 

Do we have tests for not supplying the environment variables in the helm chart rendering?

pkg/util/kubelet.go
muddyfish: Should this function's contract validate that the original path was in the host's kubelet?

pkg/driver/node/node.go
```diff
@@ -36,7 +36,7 @@ import (
 	"github.com/awslabs/mountpoint-s3-csi-driver/pkg/util"
 )
 
-var kubeletPath = util.KubeletPath()
+var kubeletPath = util.ContainerKubeletPath()
```
muddyfish: Fix variable name

pkg/driver/node/node.go
```diff
@@ -93,13 +93,20 @@ func (ns *S3NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePubl
...
 		return nil, status.Error(codes.InvalidArgument, "Target path not provided")
 	}
 
-	if !strings.HasPrefix(target, kubeletPath) {
-		klog.Errorf("NodePublishVolume: target path %q is not in kubelet path %q. This might cause mounting issues, please ensure you have correct kubelet path configured.", target, kubeletPath)
+	// Translate target path from host format to container format if needed
+	targetContainer := util.KubeletHostPathToContainerPath(targetHost)
+
+	if !strings.HasPrefix(targetContainer, kubeletPath) {
+		hostKubeletPath := os.Getenv("HOST_KUBELET_PATH")
```
muddyfish: The logic here seems subtly wrong

pkg/driver/node/node.go
```diff
@@ -93,13 +93,20 @@ func (ns *S3NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePubl
...
+		if hostKubeletPath == "" {
+			hostKubeletPath = kubeletPath
+		}
+		return nil, status.Errorf(codes.InvalidArgument, "Target path %q does not match expected kubelet path. Host kubelet path is configured as %q, but received target path %q. Please verify your kubelet configuration.", targetContainer, hostKubeletPath, targetHost)
```
muddyfish: Could you include an example error message this prints?

pkg/driver/node/node.go
```diff
@@ -145,15 +152,15 @@ func (ns *S3NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePubl
...
-	klog.V(4).Infof("NodePublishVolume: mounting %s at %s with options %v", bucket, target, args.SortedList())
+	klog.V(4).Infof("NodePublishVolume: mounting %s at %s with options %v", bucket, targetContainer, args.SortedList())
 
 	credentialCtx := credentialProvideContextFromPublishRequest(req, args)
 
-	if err := ns.Mounter.Mount(ctx, bucket, target, credentialCtx, args, fsGroup); err != nil {
-		os.Remove(target)
-		return nil, status.Errorf(codes.Internal, "Could not mount %q at %q: %v", bucket, target, err)
+	if err := ns.Mounter.Mount(ctx, bucket, targetContainer, credentialCtx, args, fsGroup); err != nil {
+		os.Remove(targetContainer)
```
muddyfish, line 199: Not related to your changes, but we should probably catch if this returns an error - maybe in a new PR?

pkg/driver/node/node.go
tadiwa-aizen: Agreed — the inline validation and fallback logic here was misleading. I've refactored this: the validation now lives inside  `KubeletHostPathToContainerPath`, which uses HostKubeletPath() (with a proper default) and returns an error directly.

pkg/util/kubelet.go
tadiwa-aizen: Noted, thanks. For the rename concern — we've added a `HostKubeletPath()` function that reads `HOST_KUBELET_PATH `(defaulting to `/var/lib/kubelet`), mirroring `ContainerKubeletPath()`. KubeletHostPathToContainerPath now uses both functions instead of a raw os.Getenv. So the 
semantics are clear: `HostKubeletPath()` = where kubelet is on the host, `ContainerKubeletPath()` = where it's mounted inside the container.

For the **Helm rendering tests** — from what i can tell ..the project doesn't currently have any Helm chart rendering tests for any environment variables ... CSI_ENDPOINT, HOST_PLUGIN_DIR, MOUNTPOINT_NAMESPACE, etc. all have the same gap. Adding Helm template tests would be a broader effort beyond the scope of this fix, but I'm willing to do it if you see it important for this change

pkg/driver/node/node.go
tadiwa-aizen: Yes. Good catch — I'll create a task to "Add error handling for the `os.Remove` cleanup call"  and add it to the task backlog.

Issue #650, opened 2025-11-28
V2 does not properly handle '=' in argument values


ksquaredkey:
/kind bug
> **_NOTE:_**  If this is a filesystem related bug, please take a look at the [Mountpoint repo](https://github.com/awslabs/mountpoint-s3) to submit a [bug report](https://github.com/awslabs/mountpoint-s3/issues/new?assignees=&labels=bug&projects=&template=bug-report.yml)

**What happened?**
Upgrading from v1 to v2, my existing PV with mountOptions was seeing the "prefix" mount option mis-parsed by the CSI pod.

```
Events:
  Type     Reason       Age                    From               Message
  ----     ------       ----                   ----               -------
  Normal   Scheduled    5m59s                  default-scheduler  Successfully assigned vault/vault-snapshot-manual-hwb-8h9b7 to ip-10-121-239-20.us-west-2.compute.internal
  Warning  FailedMount  2m45s (x6 over 5m57s)  kubelet            MountVolume.SetUp failed for volume "vault-snapshot-s3-pv" : 
rpc error: code = Internal desc = Could not mount "common-di-telem-backups" at "/var/lib/kubelet/pods/2c2b7a64-8483-4df6-8966-
20708dd6a5b0/volumes/kubernetes.io~csi/vault-snapshot-s3-pv/mount": Failed to mount at source 
"/var/lib/kubelet/plugins/s3.csi.aws.com/mnt/mp-xltz7": Failed to wait for Mountpoint Pod mp-xltz7 to be ready for 
/var/lib/kubelet/plugins/s3.csi.aws.com/mnt/mp-xltz7: Mountpoint Pod mp-xltz7 failed: error: unexpected argument 
'--prefix kwk3-di/sub' found

  tip: a similar argument exists: '--prefix'

Usage: mount-s3 --allow-delete --allow-root --foreground --prefix <PREFIX> <BUCKET_NAME> <DIRECTORY>

For more information, try '--help'.
. You can see Mountpoint logs by running: `kubectl logs -n mount-s3 mp-xltz7`. If the Mountpoint Pod already restarted, you can also pass `--previous` to get logs from the previous run.. You can see Mountpoint logs by running: `kubectl logs -n mount-s3 mp-xltz7`. If the Mountpoint Pod already restarted, you can also pass `--previous` to get logs from the previous run.
[...]

PR #658, opened 2025-12-05
Fix prefix parsing

rajdchak:
*Issue #, if available:* https://github.com/awslabs/mountpoint-s3-csi-driver/issues/650

*Description of changes:*
Fix the bug in prefix parsing

### DIFF
--- a/pkg/mountpoint/args_test.go
+++ b/pkg/mountpoint/args_test.go
@@ -622,7 +622,14 @@ func TestParseArgsProperties(t *testing.T) {
 			normalizedKey == "--help" || normalizedKey == "-h" ||
 			normalizedKey == "--version" || normalizedKey == "-v"
 
-		if len(result) == 0 && !disallowed {
+		if disallowed {
+			if len(result) != 0 {
+				t.Fatalf("Expected empty result for disallowed key, got %v for input: %q", result, input)
+			}
+			return
+		}
+
+		if len(result) == 0 {
 			t.Fatalf("ParseArgs produced empty result for input: %q", input)
 		}
 
@@ -632,9 +639,9 @@ func TestParseArgsProperties(t *testing.T) {
 			}
 		}
 
-		parsedValue, exists := args.Value(argPair.key)
-		if exists && parsedValue != strings.Trim(argPair.value, " ") {
-			t.Fatalf("Value not preserved: expected %q, got %q for input: %q", strings.TrimSpace(argPair.value), parsedValue, input)
+		parsedValue, _ := args.Value(argPair.key)
+		if parsedValue != strings.Trim(argPair.value, " ") {
+			t.Fatalf("Value not preserved: expected %q, got %q for input: %q", strings.Trim(argPair.value, " "), parsedValue, input)
 		}
 
 		reparsed := mountpoint.ParseArgs(result)


go.mod
```diff
@@ -19,6 +19,7 @@ require (
 	k8s.io/klog/v2 v2.130.1
 	k8s.io/kubectl v0.31.3
 	k8s.io/mount-utils v0.29.4
+	pgregory.net/rapid v1.2.0
```
dannycjones, line 22: We need to check licensing on this

pkg/mountpoint/args_test.go
```diff
@@ -152,6 +155,97 @@ func TestParsingMountpointArgs(t *testing.T) {
...
+				"allow-delete",
+				"region us-west-2",
+				"prefix=kwk3-di/sub vault/",
+			},
+			want: []string{
+				"--allow-delete",
+				"--prefix=kwk3-di/sub vault/",
+				"--region=us-west-2",
+			},
+		},
```
dannycjones, line 184: not sure if this case would even work, although I think in this case we'd just recommend using `--prefix="kwk3-di/sub vault/"`

I'm wondering how that would get invoked by Mountpoint when passed in this way via the driver?

pkg/mountpoint/args_test.go
```diff
@@ -152,6 +155,97 @@ func TestParsingMountpointArgs(t *testing.T) {
...
+				"allow-delete",
+				"region us-west-2",
+				"prefix kwk3-di/sub vault/",
+			},
+			want: []string{
+				"--allow-delete",
+				"--prefix=kwk3-di/sub vault/",
+				"--region=us-west-2",
+			},
+		},
```
dannycjones, line 210: This one again doesn't seem super meaningful. What should customer provide in this case, and does it work? Should we even reject this? (I consider rejecting out of scope for now though.)

I'd recommend this to customer:

```
prefix "kwk3-di/sub vault/"
```

OR

```
prefix="kwk3-di/sub vault/
```

Do these work as expected?

pkg/mountpoint/args_test.go
```diff
@@ -152,6 +155,97 @@ func TestParsingMountpointArgs(t *testing.T) {
...
+				"allow-delete",
+				"region us-west-2",
+				"prefix backup files = production/",
+			},
+			want: []string{
+				"--allow-delete",
+				"--prefix=backup files = production/",
+				"--region=us-west-2",
+			},
+		},
```
dannycjones, line 249: I'd drop this unless we implement rejection. This does not seem useful for us to test.

dannycjones (commented):
nice, put some initial comments

I haven't reviewed the proptests themselves yet

go.mod
dannycjones: I think we're good to include this.

pkg/mountpoint/args_test.go
dannycjones: After discussing together, this is not going to work due to CSI's processing of the arguments. Documented it in #660, so I think this test does make sense now because it is actually showing how customers have to format arguments and it would be a breaking change to make an improvement here.

pkg/mountpoint/args_test.go
dannycjones: Similar to https://github.com/awslabs/mountpoint-s3-csi-driver/pull/658/files#r2602121834, this is actually one of the only ways to pass prefix. `prefix "backup files = production/"` doesn't work.

So keep!

pkg/mountpoint/args_test.go
```diff
@@ -494,3 +588,52 @@ func TestCreatingMountpointArgsFromAlreadyParsedArgs(t *testing.T) {
 	parsedArgs := mountpoint.ParseArgs(args.SortedList())
 	assert.Equals(t, want, parsedArgs.SortedList())
 }
+
+func TestParseArgsProperties(t *testing.T) {
+	rapid.Check(t, func(t *rapid.T) {
+		key := rapid.StringMatching(`-{0,2}[a-z][a-z0-9-]*`).Draw(t, "key")
```
dannycjones, line 601: This matches long arguments, but there can also be short arguments.

pkg/mountpoint/args_test.go
rajdchak: as we checked together, passing prefix of these types via the driver works fine. The representation of this type in cli would be --prefix="kwk3-di/sub vault/"

pkg/mountpoint/args_test.go
```diff
@@ -494,3 +588,52 @@ func TestCreatingMountpointArgsFromAlreadyParsedArgs(t *testing.T) {
...
+		value := rapid.StringOf(rapid.OneOf(
+			rapid.Rune().Filter(func(r rune) bool { return r > 32 && r < 127 }),
```
dannycjones: What's the point of this?

Would `rapid.String()` work too?

pkg/mountpoint/args_test.go
dannycjones: It may be worthwhile to create a function which will return a generator for the arguments.

It would have the signature similar to this:

```go
type argPair struct {
    key: string,
    value: string,
}

func argumentGenerator() *rapid.Generator[argPair] {
    //
}
```

This generator should be able to generate values like this:

- `--debug`
- `region`, `us-east-1`
- `l`, `log-directory/`

The `ParseArgs` function does do extra things like filter out some unsupported arguments, so we may need to update the test or generator to accomodate things like `--help` or `-f` being missing.

pkg/mountpoint/args_test.go
dannycjones: Can you add a comment above this test input to clarify what you just confirmed? i.e. that it represents how customers can pass something like `--prefix "kwk3-di/sub vault/".

dannycjones (changes requested):
Feedback around the new property-based tests.

pkg/mountpoint/args_test.go
dannycjones: The other bit implied but I forgot to mention is to test arguments that don't have a value.

rajdchak: As discussed, this expression can accommodate short args as well.

pkg/mountpoint/args_test.go
```diff
@@ -494,3 +589,72 @@ func TestCreatingMountpointArgsFromAlreadyParsedArgs(t *testing.T) {
...
+		for _, arg := range result {
+			if !strings.HasPrefix(arg, "--") && !strings.HasPrefix(arg, "-") {
+				t.Fatalf("Key not normalized: %q from input: %q", arg, input)
+			}
+		}
+
+		parsedValue, exists := args.Value(argPair.key)
+		if exists && parsedValue != strings.Trim(argPair.value, " ") {
+			t.Fatalf("Value not preserved: expected %q, got %q for input: %q", strings.TrimSpace(argPair.value), parsedValue, input)
+		}
```
dannycjones: Does this miss if we drop the value unexpectedly?

Can we instead check if the `argPair.value` is non-empty, and if so check that it matches? The `args` is the system under test, we shouldn't trust it to tell us if the value should exist.

rajdchak: What do you mean by we drop the value? 

We are already doing a check in the beginning
if len(result) == 0 && !disallowed {
t.Fatalf("ParseArgs produced empty result for input: %q", input)
} 
if the key is not present in the args which is under test, it should catch here and fail.

rajdchak: The check for this is happening here 
if len(result) == 0 {
t.Fatalf("ParseArgs produced empty result for input: %q", input)
}

dannycjones (approved):
LGTM, thanks for adding the property-based testing, it is a big improvement!

pkg/mountpoint/args_test.go
rajdchak: If that passes it implies value will exist and then we check whether it matches the expected value. So I removed the exists check from the condition here.

PR #660, opened 2025-12-08
Add Mountpoint-specific options section to CONFIGURATION.md

dannycjones:
*Issue #, if available:* #658

*Description of changes:*

Update CONFIGURATION.md to clarify how arguments are parsed by the CSI driver and passed to Mountpoint.
Importantly, it calls out that things like quotes will be interpreted differently, impacting how you might use the prefix argument when it contains spaces.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -17,17 +17,17 @@ Below shows an example `mountOptions` value for these options
 
 ```yaml
 mountOptions:
-  - prefix hello world/
+  - prefix some-prefix/
   - maximum-throughput-gbps 25
   - write-part-size 16777216
 ```
 
 > [!IMPORTANT]
 > Mount options are passed directly to Mountpoint without shell interpretation.
-> Unlike in a shell, quoted strings like `"my value"` are treated as literal quotes,
-> not as a way to group words with spaces.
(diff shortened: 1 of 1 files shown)


docs/CONFIGURATION.md
```diff
@@ -2,9 +2,35 @@
...
+
+> [!IMPORTANT]
+> Mount options are passed directly to Mountpoint without shell interpretation.
+> Unlike in a shell, quoted strings like `"my value"` are treated as literal quotes,
+> not as a way to group words with spaces.
+> The `prefix` argument provided above highlights this,
+> it will be interpreted by Mountpoint similar to `--prefix="hello world/"`.
+
+See [the Mountpoint documentation](https://github.com/awslabs/mountpoint-s3/blob/main/doc/CONFIGURATION.md) for all available configuration options.
+
```
rajdchak, line 33: As part of this note, shall we also include how we accept prefix with spaces in mountOptions vs the cli?

dannycjones: I'm not sure what you mean, can you elaborate? I was thinking the example given demonstrates that.

muddyfish: Might also be worth including an option like `prefix=foo/` to show you can do it

Issue #535, opened 2025-07-14
Ability to Configure Mountpoint Pod


kevarr:
/feature

**Is your feature request related to a problem? Please describe.**

Hello, AWS Mountpoint S3 CSI Driver's mountpoint Pod mechanism provides value by enabling the [Mountpoint Pod Sharing](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/v2/docs/MOUNTPOINT_POD_SHARING.md) feature&mdash;Allowing multiple Pods to share a single S3 mountpoint and cache. 

Currently, the primary way to customize these mountpoint Pods is by setting resource requests and limits via `volumeAttributes` on each PersistentVolume.

There are two main challenges with this approach:

1. The curent mountpoint Pod spec is not sufficient for environments with strict security requirements. We often need to configure Pod-level and container-level `securityContext` settings (e.g., `readOnlyRootFilesystem`, `seccompProfile`), which is not possible today.

2. Configuration is applied on a per-PV basis. This requires teams to repeat the same configuration across every PV, which is prone to human error. A mechanism to define a baseline configuration for all mountpoint Pods could greatly improve manageability and governance.

**Describe the proposed solution**

I would like to propose a solution where the mountpoint Pods are fully configurable. This could be managed by a `CustomResourceDefinition` or `ConfigMap` that defines reusable templates, which users can then select via an annotation on their `PersistentVolumeClaim`. A similar approach is used by Istio[^1]

A conceptual implementation could look like:

```yaml
# A special template named default could serve as a global fallback for any PVC that doesn't request a specific template
# Merged with the internal default
apiVersion: s3.csi.aws.com/v1alpha1
kind: MountpointPodTemplate
metadata:
  name: default
spec:
  template:
    spec:
      securityContext:
        seccompProfile:
          type: RuntimeDefault
      containers:
      - name: mountpoint # This should match the default container name to be merged
        securityContext:
          readOnlyRootFilesystem: true
          capabilities:
            drop: ["ALL"]
---
apiVersion: s3.csi.aws.com/v1alpha1
kind: MountpointPodTemplate
metadata:
  name: high-throughput # Merged with the global default template above
spec:
  template:
    spec:
      containers:
      - name: mountpoint
        resources:
          requests:
            cpu: "1"
            memory: "1Gi"
          limits:
            cpu: "2"
            memory: "1Gi"
```

```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
[...]

Issue #533, opened 2025-07-11
Support injecting environment variables to Mountpoint Pod


unexge:
The CSI Driver manages [environment variables passed to Mountpoint](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/72dd1bd3c28bb4610b1cc5b6d94d737102293ce8/pkg/driver/node/envprovider/provider.go#L11-L28), and doesn't provide a way to inject environment variables explicitly as of today. 

These environment variables mostly related to AWS credentials, but there are also some environment variables as well (e.g., `AWS_MAX_ATTEMPTS` and `UNSTABLE_MOUNTPOINT_MAX_PREFETCH_WINDOW_SIZE`).

For Amazon EKS customers, the CSI Driver manages all environment variables for AWS credentials automatically and there isn't any need for tweaking those variables. There is a need for customers outside Amazon EKS to tweak some of them, like, `AWS_ENDPOINT_URL_STS`.

The CSI Driver should allow injecting some environment variables to Mountpoint Pod in a controlled way, as using some environment variables might result unexpected results. The following list of environment variables looks like good candidates for customization:
- `AWS_ENDPOINT_URL_STS` - to use IRSA outside Amazon EKS
- `AWS_MAX_ATTEMPTS` - to specify maximum retries on Mountpoint/CRT.
   -  We currently [support configuring this value via `mountOptions`](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/examples/kubernetes/static_provisioning/aws_max_attempts.yaml#L17) - which we should probably deprecate in favor of this mechanism if/when we implement it
- `UNSTABLE_MOUNTPOINT_MAX_PREFETCH_WINDOW_SIZE` - to configure [maximum prefetch window size on Mountpoint](https://github.com/awslabs/mountpoint-s3/blob/main/doc/CONFIGURATION.md#maximum-prefetch-window-size) - though this is an unstable environment variable and might be removed in future releases

---

We'd be happy to hear if there are any other environment variables people would like to customize/specify.

jmccl:
As well as UNSTABLE_MOUNTPOINT_MAX_PREFETCH_WINDOW_SIZE, I inject AWS_SECRET_ACCESS_KEY and AWS_ACCESS_KEY_ID via this script: https://github.com/jmccl/RustGallery/blob/main/mountpoint-s3/mounts3.template

yerzhan7:
[...]

PR #663, opened 2025-12-13
Add proxy support (HTTPS_PROXY, NO_PROXY) via volume attributes

phuhung273:
*Issue #, if available:*
Relates #533, #535

*Description of changes:*
Support injecting NO_PROXY, HTTPS_PROXY to mount-s3 process

### DIFF
--- a/.github/workflows/e2e-test-untrusted.yaml
+++ b/.github/workflows/e2e-test-untrusted.yaml
@@ -30,15 +30,11 @@ jobs:
       ref: ${{ github.event.pull_request.head.sha }}
     secrets: inherit
   e2e-rosa:
-    name: "E2E ROSA Tests (${{ matrix.region }})"
+    name: E2E ROSA Tests
     uses: ./.github/workflows/e2e-rosa-tests.yaml
-    strategy:
-      matrix:
-        region:
-          - 'us-east-1'
     needs: approval
     with:
-      environment: "${{ matrix.region }}-rosa-untrusted"
+      environment: "rosa-untrusted"
       ref: ${{ github.event.pull_request.head.sha }}
     secrets: inherit
   outcome:
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,7 +44,7 @@ RUN MP_ARCH=`echo ${TARGETARCH} | sed s/amd64/x86_64/` && \
     patchelf --set-rpath '$ORIGIN' /mountpoint-s3/bin/mount-s3
 
 # Build driver. Use BUILDPLATFORM not TARGETPLATFORM for cross compilation
-FROM --platform=$BUILDPLATFORM public.ecr.aws/eks-distro-build-tooling/golang:1.26.1 as builder
+FROM --platform=$BUILDPLATFORM public.ecr.aws/eks-distro-build-tooling/golang:1.26.2 as builder
 ARG TARGETARCH
 
 WORKDIR /go/src/github.com/awslabs/mountpoint-s3-csi-driver
--- a/Dockerfile.local
+++ b/Dockerfile.local
@@ -46,7 +46,7 @@ RUN cd mountpoint-s3 && \
 #
 
 # Use BUILDPLATFORM not TARGETPLATFORM for cross compilation
-FROM --platform=$BUILDPLATFORM public.ecr.aws/eks-distro-build-tooling/golang:1.26.1 as builder
+FROM --platform=$BUILDPLATFORM public.ecr.aws/eks-distro-build-tooling/golang:1.26.2 as builder
 ARG TARGETARCH
 
 WORKDIR /go/src/github.com/awslabs/mountpoint-s3-csi-driver
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/awslabs/mountpoint-s3-csi-driver
 
-go 1.26.1
+go 1.26.2
 
 require (
 	github.com/aws/aws-sdk-go-v2 v1.39.2
--- a/tests/e2e-kubernetes/go.mod
+++ b/tests/e2e-kubernetes/go.mod
@@ -1,6 +1,6 @@
 module github.com/awslabs/mountpoint-s3-csi-driver/tests/e2e-kubernetes
 
-go 1.26.1
+go 1.26.2
 
 require (
 	github.com/aws/aws-sdk-go-v2 v1.39.2
--- a/tests/e2e-kubernetes/scripts/eksctl.sh
+++ b/tests/e2e-kubernetes/scripts/eksctl.sh
@@ -209,6 +209,21 @@ function eksctl_delete_cluster_cf_stack() {
         return 0
     fi
 
+    # Delete any subnets orphaned by a CloudFormation SDK retry during stack creation.
+    # CreateSubnet is not idempotent: if the first API call succeeds but the response times out
+    # (a transient issue), the SDK retries and gets a CIDR conflict error. CFN sees only the
+    # failure, records no PhysicalResourceId, and cannot clean up the subnet during rollback.
+    # These orphaned subnets are still tagged with the stack name but are unknown to CFN,
+    # and will block VPC deletion.
+    SUBNETS=$(aws ec2 describe-subnets --region ${REGION} \
+        --filters "Name=tag:aws:cloudformation:stack-name,Values=${STACK_NAME}" \
+        --query 'Subnets[*].SubnetId' --output text)
+    if [ -n "$SUBNETS" ]; then
+        for SUBNET_ID in $SUBNETS; do
+            aws ec2 delete-subnet --region ${REGION} --subnet-id ${SUBNET_ID} || true
+        done
+    fi
+
     aws cloudformation delete-stack --region ${REGION} --stack-name ${STACK_NAME}
 
     # GuardDury creates resources (namely an endpoint and a security group), which are not handled by eks cfn stack and prevents it from being deleted


tests/controller/controller_test.go
```diff
@@ -984,6 +984,37 @@ var _ = Describe("Mountpoint Controller", func() {
...
+			mpContainer := &mountpointPod.Spec.Containers[0]
+
+			noProxyUpperCase := ""
+			noProxyLowerCase := ""
+
+			for _, env := range mpContainer.Env {
+				switch env.Name {
+				case "NO_PROXY":
+					noProxyUpperCase = env.Value
+				case "no_proxy":
```
yerzhan7: Why we need to support lower case env variables?

docs/CONFIGURATION.md
```diff
@@ -61,6 +61,9 @@ spec:
       mountpointContainerResourcesRequestsMemory: 1Gi
       mountpointContainerResourcesLimitsCpu: 500m
       mountpointContainerResourcesLimitsMemory: 1Gi
+      # Optional: Configure Mountpoint container proxy
+      mountpointContainerNoProxy: ""
+      mountpointContainerHttpsProxy: ""
```
yerzhan7: As suggested in the previous comment I would prefer for this to be in the following YAML format instead:

```yaml
volumeAttributes:
  mountpointContainerEnv: |
    - name: UNSTABLE_MOUNTPOINT_MAX_PREFETCH_WINDOW_SIZE
      value: "502000"
    - name: HTTPS_PROXY
      value: https://example.com
```

pkg/podmounter/mppod/creator.go
```diff
@@ -393,6 +394,28 @@ func (c *Creator) configureResourceLimits(mpContainer *corev1.Container, volumeA
 	return nil
 }
 
+// configureProxy configures proxy environment variables of the container if its specified in the volume attributes.
+func (c *Creator) configureProxy(mpContainer *corev1.Container, volumeAttributes map[string]string) {
```
yerzhan7: As suggested in the previous comment, it's better to pass environmental variables in the node driver component during `NodePublishVolume` requests rather than in controller. 

Passing in controller wouldn't work because it's the node driver that constructs and feeds environmental variables into `mount-s3` process.

yerzhan7:
> Just 1 issue: adding https_proxy test in controller_test.go requires setting up the proxy, so I only add no_proxy for simplicity. Please let me know if this is acceptable.

K8s controller test framework do not actually run containers. You need to create end-2-end tests instead.

I agree that setting up proxy might be a bit complicated for the test. Ideally, we want that in the test, but for now it's fine to do negative test - i.e. set HTTPS_PROXY to a non-existent proxy and verify that mounting fails.

yerzhan7 (changes requested):

tests/controller/controller_test.go
phuhung273: From my experience, some tools use uppercase, some use lowercase, some work with both:
- `curl` uses `http_proxy, HTTPS_PROXY, NO_PROXY`, ref: https://curl.se/docs/manpage.html
- `wget` uses `http_proxy, https_proxy, no_proxy`, ref: https://www.gnu.org/software/wget/manual/html_node/Proxies.html

So I always set both to ensure everything works. Have you ever seen such issue ?

docs/CONFIGURATION.md
phuhung273: Sorry I missed this part, I've updated to use YAML format.

pkg/podmounter/mppod/creator.go
phuhung273: Thank you for the instruction, I've updated to `NodePublishVolume` approach. Im not sure if `credentialprovider` is the correct place to add, so just give a minimal code change (not enough test) to show the idea. Could you help take a look ?

tests/controller/controller_test.go
yerzhan7: In our case, we control the UX via `mountpointContainerEnv`, so there's no reason to support both upper and lower cases — it would confuse users. Let's  support only uppercase to be consistent with the env variables we plan to support in the future (like `AWS_ENDPOINT_URL_STS`). 

If `mount-s3` internally needs the lowercase variant, we can handle that mapping inside the driver. But at the moment it looks like it's not necessary.

pkg/podmounter/mppod/creator.go
yerzhan7: `credentialProvider` is incorrect place as it only relates to credential stuff.

We should parse new volume attribute in `node.go` along with parsing of ther volume attributes (`bucketName`, etc.).

Then probably you would need to adjust `Mounter.Mount()` interface and add another `envprovider.Environment` parameter. Then pass it into `mountS3AtSource()` method where you would merge it with `env` variable.

tests/controller/controller_test.go
phuhung273: Sure, I've removed the lowercase vars.

pkg/podmounter/mppod/creator.go
phuhung273: I've updated to use your suggestion. Hope I get your point correct. Could you please have a look, thanks.

pkg/driver/node/envprovider/provider.go
```diff
@@ -25,6 +25,8 @@ const (
 	EnvSecretAccessKey                 = "AWS_SECRET_ACCESS_KEY"
 	EnvSessionToken                    = "AWS_SESSION_TOKEN"
 	EnvMountpointCacheKey              = "UNSTABLE_MOUNTPOINT_CACHE_KEY"
+	EnvHttpsProxy                      = "HTTPS_PROXY"
```
yerzhan7: In Go, the correct way is `EnvHTTPSProxy`.

See https://google.github.io/styleguide/go/decisions.html#initialisms

pkg/driver/node/mounter/mounter.go
```diff
@@ -6,12 +6,13 @@ import (
...
 
 	"github.com/awslabs/mountpoint-s3-csi-driver/pkg/driver/node/credentialprovider"
+	"github.com/awslabs/mountpoint-s3-csi-driver/pkg/driver/node/envprovider"
 	"github.com/awslabs/mountpoint-s3-csi-driver/pkg/mountpoint"
 )
 
 // Mounter is an interface for mount operations
 type Mounter interface {
-	Mount(ctx context.Context, bucketName string, target string, credentialCtx credentialprovider.ProvideContext, args mountpoint.Args, fsGroup string) error
+	Mount(ctx context.Context, bucketName string, target string, credentialCtx credentialprovider.ProvideContext, args mountpoint.Args, fsGroup string, customEnv envprovider.Environment) error
```
yerzhan7: I would prefer name `userEnv`. 

`customEnv` is a bit vague. `userEnv` indicates better than it's environment from user.

Same in `mountS3AtSource()` method.

pkg/driver/node/mounter/pod_mounter.go
```diff
@@ -204,10 +205,11 @@ func (pm *PodMounter) Mount(ctx context.Context, bucketName string, target strin
 //
 // If any step fails, it ensures cleanup by unmounting the source path.
 func (pm *PodMounter) mountS3AtSource(ctx context.Context, source string, mpPod *corev1.Pod, podPath string,
-	bucketName string, credEnv envprovider.Environment, authenticationSource credentialprovider.AuthenticationSource,
+	bucketName string, credEnv envprovider.Environment, customEnv envprovider.Environment, authenticationSource credentialprovider.AuthenticationSource,
 	args mountpoint.Args) error {
 	env := envprovider.Default()
 	env.Merge(credEnv)
+	env.Merge(customEnv)
```
yerzhan7: The merge order in here should be changed so that credential env vars always take the highest priority and user provided env takes the lowest priority:
```go
  env := envprovider.Environment{}
  env.Merge(userEnv)
  env.Merge(envprovider.Default())
  env.Merge(credEnv)
```

Currently `customEnv` is merged last, which means user-provided values could theoretically override credential env vars. While the allowlist prevents this today, it's safer to ensure credentials can never be overridden regardless of future allowlist changes.

pkg/driver/node/node.go
```diff
@@ -319,3 +326,28 @@ func logSafeNodePublishVolumeRequest(req *csi.NodePublishVolumeRequest) *csi.Nod
 		VolumeContext:     safeVolumeContext,
 	}
 }
+
+// customEnvFromRequest validate and return environment variables provided by user
+func customEnvFromRequest(req *csi.NodePublishVolumeRequest) (envprovider.Environment, error) {
```
yerzhan7: `customEnvFromRequest` -> `userEnvFromRequest`

pkg/driver/node/node.go
```diff
@@ -319,3 +326,28 @@ func logSafeNodePublishVolumeRequest(req *csi.NodePublishVolumeRequest) *csi.Nod
...
+			return nil, status.Errorf(codes.InvalidArgument, "Invalid format: %v", err)
+		}
+		for _, ctxEnv := range ctxEnvs {
+			switch ctxEnv.Name {
+			case envprovider.EnvHttpsProxy:
+				env[envprovider.EnvHttpsProxy] = ctxEnv.Value
+			case envprovider.EnvNoProxy:
+				env[envprovider.EnvNoProxy] = ctxEnv.Value
+			default:
+				return nil, status.Errorf(codes.InvalidArgument, "Environment variable not allowed: %s", ctxEnv.Name)
```
yerzhan7: We should return `fmt.Error` in this method like  `fmt.Errorf("environment variable not allowed: %s", ctxEnv.Name)`, and let the caller to wrap it into `status.Errorf`. 

At the moment, there is double wrapping of gRPC status error.

pkg/driver/node/envprovider/provider.go
yerzhan7, line 46: Can we create similar allowlist for user provided env? E.g.
```go
 var UserEnvAllowlist = []Key{
        EnvHTTPSProxy,
        EnvNoProxy,
  }
```

And then use it your customEnvFromRequest method?

(we can also do a map instead of list for faster lookups)

docs/CONFIGURATION.md
```diff
@@ -61,6 +61,8 @@ spec:
       mountpointContainerResourcesRequestsMemory: 1Gi
       mountpointContainerResourcesLimitsCpu: 500m
       mountpointContainerResourcesLimitsMemory: 1Gi
+      # Optional: Configure Mountpoint container environment variables
+      mountpointContainerEnv: ""
```
yerzhan7: May I suggest the following:

```

# ----- ENVIRONMENT VARIABLE CONFIGURATION -----
# Optional: Pass extra environment variables to the mount-s3 process. Default: none.
# Supported variables: HTTPS_PROXY, NO_PROXY
mountpointEnv.HTTPS_PROXY: "https://proxy.example.com"
mountpointEnv.NO_PROXY: "169.254.169.254"
```

nit: Also, it would be nice if you add a full working example under `examples/kubernetes/static_provisioning/`?

pkg/driver/node/envprovider/provider.go
phuhung273: Thank you for showing this convention

pkg/driver/node/node.go
phuhung273: Updated to `userEnv`

pkg/driver/node/node.go
phuhung273: Sure, I've updated to `fmt.Error`

pkg/driver/node/envprovider/provider.go
phuhung273: Sure, I've updated to use a UserEnvAllowlist. Later env can be added to this list seamlessly.

Given the allow list is still small and can be optimized to a map easily, I chose to use slice for readability. Please let me know what you think. Thanks for the feedback.

pkg/driver/node/mounter/pod_mounter.go
phuhung273: Thank you for the explanation, I've updated to use your snippet to lower user env priority.

pkg/driver/node/mounter/mounter.go
phuhung273: Yeah I shared the same thought. Updated to `userEnv`

docs/CONFIGURATION.md
phuhung273: Yes, that makes lots of sense to have a working example. As i havent run this change in my corp cluster, please allow me some time for e2e.

examples/kubernetes/static_provisioning/environment_variables.yaml
```diff
@@ -0,0 +1,55 @@
...
+  mountOptions:
+    - allow-delete
+    - region us-east-1
+  csi:
+    driver: s3.csi.aws.com # Required
+    volumeHandle: s3-csi-driver-volume # Must be unique
+    volumeAttributes:
+      bucketName: s3-csi-driver
+      # Supported variables: HTTPS_PROXY, NO_PROXY
+      mountpointEnv.HTTPS_PROXY: "https://proxy.example.com"
```
yerzhan7: I was thinking having a simple/minimal example with a working proxy.

e.g. could you please add this?

```
apiVersion: v1
kind: Pod
metadata:
  name: proxy
  labels:
    app: proxy
spec:
  containers:
    - name: proxy
      image: ubuntu/squid:latest
      ports:
        - containerPort: 3128
---
apiVersion: v1
kind: Service
metadata:
  name: proxy
spec:
  selector:
    app: proxy
  ports:
    - port: 3128
```

And then do:

```
mountpointEnv.HTTPS_PROXY: "http://proxy.default.svc.cluster.local:3128"
```

pkg/driver/node/envprovider/provider.go
```diff
@@ -66,6 +75,24 @@ func Default() Environment {
 	return environment
 }
 
+func ParseUserEnvFromVolumeContext(volumeCtx map[string]string) (Environment, error) {
+	env := Environment{}
+	for key, value := range volumeCtx {
+		if !strings.Contains(key, "mountpointEnv.") {
```
yerzhan7: `Contains()` would also match `adsfmountpointEnv.` which is not what we want. Let's use `CutPrefix`?

Also, let's extract constant to the top of the file (`const mountpointEnvPrefix = "mountpointEnv."`) and add allowed list of env variables to the error message (for better error message).

E.g.
```go
envName, ok := strings.CutPrefix(key, mountpointEnvPrefix)
if !ok {
    continue
}
if !slices.Contains(userEnvAllowlist, envName) {
    return nil, fmt.Errorf("environment variable not allowed: %s (allowed: %v)", key, userEnvAllowlist)
}
env.Set(envName, value)
```

docs/CONFIGURATION.md
```diff
@@ -62,6 +62,12 @@ spec:
       mountpointContainerResourcesLimitsCpu: 500m
       mountpointContainerResourcesLimitsMemory: 1Gi
 
+      # ----- ENVIRONMENT VARIABLE CONFIGURATION -----
+      # Optional: Pass extra environment variables to the mount-s3 process. Default: none.
+      # Supported variables: HTTPS_PROXY, NO_PROXY
+      mountpointEnv.HTTPS_PROXY: ""
+      mountpointEnv.NO_PROXY: ""
```
yerzhan7: Could we actually set example values like this?

```
mountpointEnv.HTTPS_PROXY: "http://proxy.example.com:3128"
mountpointEnv.NO_PROXY: "sts.amazonaws.com"
```

examples/kubernetes/static_provisioning/proxy.yaml
```diff
@@ -0,0 +1,55 @@
+apiVersion: v1
```
yerzhan7, line 1: Could we rename file to be `proxy.yaml`?

also, we might need to add new entry in https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/examples/kubernetes/static_provisioning/README.md

pkg/driver/node/mounter/pod_mounter.go
```diff
@@ -208,9 +209,11 @@ func (pm *PodMounter) Mount(ctx context.Context, bucketName string, target strin
 //
 // If any step fails, it ensures cleanup by unmounting the source path.
 func (pm *PodMounter) mountS3AtSource(ctx context.Context, source string, mpPod *corev1.Pod, podPath string,
-	bucketName string, credEnv envprovider.Environment, authenticationSource credentialprovider.AuthenticationSource,
+	bucketName string, credEnv envprovider.Environment, userEnv envprovider.Environment, authenticationSource credentialprovider.AuthenticationSource,
 	args mountpoint.Args) error {
-	env := envprovider.Default()
+	env := envprovider.Environment{}
```
yerzhan7, line 216: Could you please add small code comment like this?
```
// Build environment with precedence (highest wins): credEnv > Default() > userEnv
```

yerzhan7:
Thanks for the update @phuhung273, added some comments and did some manual testing with proxy (it works).

We'd also need e2e tests for this feature. Ideally at least these:

1. Positive proxy test — deploy a Squid proxy pod + service in the test namespace, then mount a volume with `mountpointEnv.HTTPS_PROXY` pointing to the proxy service. Verify the mount succeeds and check the Mountpoint pod logs for a line like "through a tunnel via proxy" to confirm traffic actually went through the proxy (would need `--debug-crt` mount flag).
2. Negative proxy test — set `mountpointEnv.HTTPS_PROXY` to a non-existent proxy and verify the mount fails.
3. Unallowed env var test — set an env var not in the allowlist (e.g. `mountpointEnv.FOO`) and verify it's rejected.

For reference:
- `accessmode.go` file is the simplest test suite to use as a template
- `credentials.go` shows how to pass custom volume attributes via contextWithVolumeAttributes / VolumeAttributesFromContext

Would you be up for writing these? I understand it might be tricky — if you'd prefer, I'm happy to write the e2e tests on my side.

phuhung273:
Thanks @yerzhan7 for the help. If this isn't urgent, I would like to try writing a new test suite based on your guidance, so that i can get better understanding of the driver. Please allow me a few days for this e2e.

pkg/driver/node/mounter/pod_mounter.go
phuhung273: Sure I've added this comment

examples/kubernetes/static_provisioning/proxy.yaml
phuhung273: Sure I've renamed and added an entry to static_provisioning/README

docs/CONFIGURATION.md
phuhung273: Sure, example values added

pkg/driver/node/envprovider/provider.go
phuhung273: Thank you for catching this. I've updated to use `CutPrefix`

examples/kubernetes/static_provisioning/environment_variables.yaml
phuhung273: I've added Squid e2e test. Do you think we still need an example in static_provisioning ?

tests/e2e-kubernetes/testsuites/proxy.go
```diff
@@ -0,0 +1,215 @@
...
+					{
+						Name:  "proxy",
+						Image: "ubuntu/squid:latest",
+						Ports: []v1.ContainerPort{{ContainerPort: proxyPort}},
+					},
+				},
+			},
+		}
+
+		proxyPod, err := client.Create(ctx, proxyPod, metav1.CreateOptions{})
```
yerzhan7: We should create pods using `createPod()` helper method  that also waits for it to be in Running state.

tests/e2e-kubernetes/testsuites/proxy.go
```diff
@@ -0,0 +1,215 @@
...
+			fmt.Sprintf("uid=%d", defaultNonRootUser),
+			fmt.Sprintf("gid=%d", defaultNonRootGroup),
+			"allow-other",
+			"debug",
+			"debug-crt",
+		})
+		l.resources = append(l.resources, resource)
+
+		ginkgo.By("Creating proxy pod")
+		client := f.ClientSet.CoreV1().Pods(v1.NamespaceDefault)
```
yerzhan7: We should use test namespace for test isolation, instead of default namespace: `f.Namespace.Name`

tests/e2e-kubernetes/testsuites/proxy.go
```diff
@@ -0,0 +1,215 @@
...
+			Spec: v1.ServiceSpec{
+				Selector: proxyLabels,
+				Ports:    []v1.ServicePort{{Port: proxyPort}},
+			},
+		}
+
+		proxyService, err = serviceClient.Create(ctx, proxyService, metav1.CreateOptions{})
+		framework.ExpectNoError(err)
+		defer func() {
+			e2eservice.WaitForServiceDeletedWithFinalizer(ctx, f.ClientSet, v1.NamespaceDefault, proxyService.Name)
```
yerzhan7: This defer cleanup method does not actually delete the service? (it just waits for deletion)

tests/e2e-kubernetes/testsuites/proxy.go
```diff
@@ -0,0 +1,215 @@
...
+		proxyUrl := "proxy.default.svc.cluster.local"
+		var proxyPort int32 = 3128
+		proxy := contextWithVolumeAttributes(ctx, map[string]string{
+			"mountpointEnv.HTTPS_PROXY": fmt.Sprintf("http://%s:%d", proxyUrl, proxyPort),
+		})
+
+		resource := createVolumeResource(proxy, l.config, pattern, v1.ReadWriteMany, []string{
+			fmt.Sprintf("uid=%d", defaultNonRootUser),
+			fmt.Sprintf("gid=%d", defaultNonRootGroup),
+			"allow-other",
```
yerzhan7: We do not need these extra flags for simple proxy test.

tests/e2e-kubernetes/testsuites/proxy.go
```diff
@@ -0,0 +1,215 @@
...
+		checkWriteToPath(ctx, f, pod, fileInVol, toWrite, seed)
+		ginkgo.By("Checking read from a volume")
+		checkReadFromPath(ctx, f, pod, fileInVol, toWrite, seed)
+		ginkgo.By("Checking file group owner")
+		e2epod.VerifyExecInPodSucceed(ctx, f, pod, fmt.Sprintf("stat -L -c '%%a %%g %%u' %s | grep '644 %d %d'", fileInVol, defaultNonRootGroup, defaultNonRootUser))
+		ginkgo.By("Checking dir group owner")
+		e2epod.VerifyExecInPodSucceed(ctx, f, pod, fmt.Sprintf("stat -L -c '%%a %%g %%u' %s | grep '755 %d %d'", volPath, defaultNonRootGroup, defaultNonRootUser))
+		ginkgo.By("Checking pod identity")
+		e2epod.VerifyExecInPodSucceed(ctx, f, pod, fmt.Sprintf("id | grep 'uid=%d gid=%d groups=%d'", defaultNonRootUser, defaultNonRootGroup, defaultNonRootGroup))
+
```
yerzhan7, line 183: These checks are all unnecessary for proxy test - let's just do simple `write` check only.

tests/e2e-kubernetes/testsuites/proxy.go
```diff
@@ -0,0 +1,215 @@
...
+	})
+
+	ginkgo.It("should fail when mountpointEnv.HTTPS_PROXY set to a non-existent proxy", func(ctx context.Context) {
+		expectFailToMount(contextWithVolumeAttributes(ctx, map[string]string{
+			"mountpointEnv.HTTPS_PROXY": "nonexistentproxy",
+		}))
+	})
+
+	ginkgo.It("should reject unallowed env var", func(ctx context.Context) {
+		expectFailToMount(contextWithVolumeAttributes(ctx, map[string]string{
```
yerzhan7, line 201: Could we add another parameter `expectedErrMsg str` to `expectFailToMount()` method and then make error message assertions like these?

```go
		err = e2eevents.WaitTimeoutForEvent(ctx, f.ClientSet, f.Namespace.Name, eventSelector, expectedErrMsg, 5*time.Minute)
		framework.ExpectNoError(err, "Expected FailedMount event containing %q", expectedErrMsg)
		framework.Logf("Got FailedMount event containing %q", expectedErrMsg)
```
```go
...
		expectFailToMount(contextWithVolumeAttributes(ctx, map[string]string{
			"mountpointEnv.FOO": "BAR",
		}), "environment variable not allowed")
```

And for the `"should fail when mountpointEnv.HTTPS_PROXY set to a non-existent proxy"` test case above the error should be `AWS_IO_DNS_INVALID_NAME` I think.

tests/e2e-kubernetes/testsuites/proxy.go
```diff
@@ -0,0 +1,215 @@
...
+		gomega.Expect(err).NotTo(gomega.HaveOccurred(), "Failed to find Mountpoint pods")
+
+		logs, err := e2epod.GetPodLogs(ctx, f.ClientSet, mpPods[0].Namespace, mpPods[0].Name, mpPods[0].Spec.Containers[0].Name)
+		framework.ExpectNoError(err)
+		gomega.Expect(logs).To(gomega.ContainSubstring(fmt.Sprintf("through a tunnel via proxy \"%s\"", proxyUrl)))
+	})
+
+	ginkgo.It("should fail when mountpointEnv.HTTPS_PROXY set to a non-existent proxy", func(ctx context.Context) {
+		expectFailToMount(contextWithVolumeAttributes(ctx, map[string]string{
+			"mountpointEnv.HTTPS_PROXY": "nonexistentproxy",
```
yerzhan7: `nonexistentproxy` -> `http://proxy.invalid:3128` to be safe

tests/e2e-kubernetes/testsuites/proxy.go
```diff
@@ -0,0 +1,215 @@
...
+			framework.Logf("Didn't get FailedMount event: %s", eventSelector)
+		}
+
+		pod, err = client.Get(ctx, pod.Name, metav1.GetOptions{})
+		framework.ExpectNoError(err)
+		gomega.Expect(pod.Status.Phase).To(gomega.Equal(v1.PodPending))
+	}
+
+	ginkgo.It("should be able to operate behind proxy", func(ctx context.Context) {
+		proxyUrl := "proxy.default.svc.cluster.local"
```
yerzhan7: Could we generate unique names (e.g., `f.UniqueName + "-proxy"` or use `GenerateName`) and deploy the proxy pod/service in `f.Namespace.Name` instead of `default`? Then construct the `HTTPS_PROXY` URL dynamically from the actual service name and namespace. This would make the test fully isolated.

And for proxy label it could be something like this:

```go
proxyName := f.UniqueName + "-proxy"
proxyLabels := map[string]string{
    "app": proxyName,
}
```

tests/e2e-kubernetes/testsuites/proxy.go
```diff
@@ -0,0 +1,215 @@
...
+func (t *s3CSIProxyTestSuite) DefineTests(driver storageframework.TestDriver, pattern storageframework.TestPattern) {
+	type local struct {
+		resources []*storageframework.VolumeResource
+		config    *storageframework.PerTestConfig
+	}
+	var (
+		l local
+	)
+
+	f := framework.NewFrameworkWithCustomTimeouts("proxy", storageframework.GetDriverTimeouts(driver))
```
yerzhan7: Could you please add `NamespacePrefix+"proxy"`? I think some other test files do not have it by mistake.

tests/e2e-kubernetes/testsuites/proxy.go
```diff
@@ -0,0 +1,215 @@
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "proxy",
+				Namespace: v1.NamespaceDefault,
+				Labels:    proxyLabels,
+			},
+			Spec: v1.PodSpec{
+				Containers: []v1.Container{
+					{
+						Name:  "proxy",
+						Image: "ubuntu/squid:latest",
```
yerzhan7: Could we please change to `public.ecr.aws/ubuntu/squid:latest` to avoid introducing new dependency on Docker Hub for e2e tests? (All other e2e tests use images from ECR Public repo)

However, in the `examples` folder you should still use `ubuntu/squid:latest` for consistency with other images in other `examples`.

examples/kubernetes/static_provisioning/environment_variables.yaml
yerzhan7: Yes, working examples are very useful for customers.

tests/e2e-kubernetes/testsuites/proxy.go
phuhung273: Yes I can confirm this method actually deletes the service https://github.com/kubernetes/kubernetes/blob/6c1cd99aef09161ddb07b8ade6c9564e9b9bfe27/test/e2e/framework/service/wait.go#L34-L38

tests/e2e-kubernetes/testsuites/proxy.go
phuhung273: Sure, e2e updated to use Public ECR while `examples` use docker hub public

tests/e2e-kubernetes/testsuites/proxy.go
phuhung273: Sure, I've updated framework name as suggested.

tests/e2e-kubernetes/testsuites/proxy.go
phuhung273: I agree with this. The reason was Squid needs more configs if we want to run in `Restricted security level`. So I decided to run it in `default` with no restriction while keeping testing namespace under `Restricted`.

Now, to keep Squid simple, I decreased namespace to `Baseline`, keeping workload pod at `Restricted`. Please see my comment and let me know if this is fair enough.

The test is fully isolated now. Appreciate the suggestion.

tests/e2e-kubernetes/testsuites/proxy.go
phuhung273: Sure, I've updated invalid proxy URL as suggested.

tests/e2e-kubernetes/testsuites/proxy.go
phuhung273: Thank you for the suggestion, I was thinking the same. Updated to also validate error message.

tests/e2e-kubernetes/testsuites/proxy.go
phuhung273: Sure, i worried it is not enough so just put everything. Removed all besides from `checkWriteToPath`

tests/e2e-kubernetes/testsuites/proxy.go
phuhung273: Yes i've removed this as we only need `write` validation

tests/e2e-kubernetes/testsuites/proxy.go
phuhung273: Sure, updated to use `createPod`

examples/kubernetes/static_provisioning/proxy.yaml
```diff
@@ -20,8 +20,8 @@ spec:
     volumeAttributes:
       bucketName: s3-csi-driver
       # Supported variables: HTTPS_PROXY, NO_PROXY
-      mountpointEnv.HTTPS_PROXY: "http://proxy.example.com:3128"
-      mountpointEnv.NO_PROXY: "sts.amazonaws.com"
+      mountpointEnv.HTTPS_PROXY: "http://proxy.default.svc.cluster.local:3128"
+      mountpointEnv.NO_PROXY: "169.254.169.254"
```
yerzhan7: `169.254.169.254` is an IMDS endpoint which is HTTP (not HTTPS) I believe. So, the NO_PROXY to IMDS is redundant here as it will always bypass proxy.

Let's do instead `sts.us-east-1.amazonaws.com` as a legitimate example where customers might want to avoid proxy.

tests/e2e-kubernetes/testsuites/proxy.go
```diff
@@ -90,71 +96,64 @@ func (t *s3CSIProxyTestSuite) DefineTests(driver storageframework.TestDriver, pa
...
 	}
 
 	ginkgo.It("should be able to operate behind proxy", func(ctx context.Context) {
-		proxyUrl := "proxy.default.svc.cluster.local"
+		proxyName := fmt.Sprintf("%s-proxy", f.UniqueName)
+		proxyUrl := fmt.Sprintf("%s.%s.svc.cluster.local", proxyName, f.Namespace.Name)
 		var proxyPort int32 = 3128
 		proxy := contextWithVolumeAttributes(ctx, map[string]string{
 			"mountpointEnv.HTTPS_PROXY": fmt.Sprintf("http://%s:%d", proxyUrl, proxyPort),
+			"mountpointEnv.NO_PROXY":    "169.254.169.254",
```
yerzhan7: np: We do not actually test this here - it can be removed.

docs/CONFIGURATION.md
```diff
@@ -62,6 +62,12 @@ spec:
       mountpointContainerResourcesLimitsCpu: 500m
       mountpointContainerResourcesLimitsMemory: 1Gi
 
+      # ----- ENVIRONMENT VARIABLE CONFIGURATION -----
+      # Optional: Pass extra environment variables to the mount-s3 process. Default: none.
+      # Supported variables: HTTPS_PROXY, NO_PROXY
+      mountpointEnv.HTTPS_PROXY: "http://proxy.example.com:3128"
+      mountpointEnv.NO_PROXY: "sts.amazonaws.com"
```
yerzhan7: Sorry, noticed just now. Could we change NO_PROXY example to `sts.us-east-1.amazonaws.com` as Mountpoint-S3 does not support global STS endpoint.

tests/e2e-kubernetes/testsuites/proxy.go
phuhung273: Sure this NO_PROXY is removed

examples/kubernetes/static_provisioning/proxy.yaml
phuhung273: Sure, NO_PROXY changed to `sts.us-east-1.amazonaws.com`

docs/CONFIGURATION.md
phuhung273: Sure, NO_PROXY changed to `sts.us-east-1.amazonaws.com`

PR #687, opened 2026-02-05
Remove exclusion of Bottlerocket on ARM tests

dannycjones:
*Issue #, if available:*

*Description of changes:*

The end-to-end test workflows currently exclude ARM-based Bottlerocket runs. No reason is given, so removing this exclusion.

The workflow is tested here by pushing to the workflow testing branch prefix: https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/21716747095

### DIFF
--- a/.github/matrix.yaml
+++ b/.github/matrix.yaml
@@ -23,9 +23,6 @@ matrix:
     - family: "AmazonLinux2023"
       selinux-mode: "enforcing"
   exclude:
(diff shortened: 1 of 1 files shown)


dannycjones:
The tests are passing but we are hitting service quota limits on number of VPCs, etc.. Leaving in draft until that's resolved.

PR #753, opened 2026-03-19
Update container base image from AL2 to AL2023

tadiwa-aizen:
*Issue #, if available:*

*Description of changes:*
Update container base image from Amazon Linux 2 to AL2023.

Manual testing:
Manual E2E tests pass: https://github.com/tadiwa-aizen/mountpoint-s3-csi-driver/actions/runs/23293621981/job/67735199646
 (Disregarding rosa as that isn't set up on my ci)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,15 +15,12 @@
 ARG MOUNTPOINT_VERSION=1.22.2
 
 # Download the mountpoint tarball and produce an installable directory
-# Building on Amazon Linux 2 because it has an old libc version. libfuse from the os
-# is being packaged up in the container and a newer version linking to a too new glibc
-# can cause portability issues
-FROM --platform=$TARGETPLATFORM public.ecr.aws/amazonlinux/amazonlinux:2 as mp_builder
+FROM --platform=$TARGETPLATFORM public.ecr.aws/amazonlinux/amazonlinux:2023 as mp_builder
 ARG MOUNTPOINT_VERSION
 ARG TARGETARCH
 ARG TARGETPLATFORM
-# We need the full version of GnuPG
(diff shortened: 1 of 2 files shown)


Dockerfile
yerzhan7: Should we update to AL2023 here as well?

yerzhan7: AL2023 uses DNF instead of `yum`

yerzhan7: Is this AL2 or AL2023? Should we update this one as well?

yerzhan7 (dismissed):

Dockerfile
tadiwa-aizen: There is an AL2023 variant available (`golang:1.26-yum-al23`), but since this project builds with `CGO_ENABLED=0`, the binaries are fully statically linked with no `glibc`  dependency — so the build OS doesn't affect the output. Happy to update it to `golang:1.26-yum-al23` for consistency if you'd prefer.

yerzhan7: Let's update this one to AL2023 in this PR for consistency? 

Also, I do not think we need the `-yum` variant - the base one should suffice: `golang:1.26.1-al23`

yerzhan7: Fixed in https://github.com/awslabs/mountpoint-s3-csi-driver/commit/a1b2889cc8addf0394d258559b01482b66c4f484

jet-tong (approved):
LGTM - I also ran [tests on my fork](https://github.com/jet-tong/mountpoint-s3-csi-driver/actions/runs/25912412172) but since there's no workflow changes (and only Dockerfile changes) it seems the fork PR untrusted workflow test does test the AL2023 image upgrade already so it should be working on ROSA as well.

PR #761, opened 2026-03-30
Allow e2e untrusted outcome to skip if PR is from same repo

jet-tong:
*Issue #, if available:*

*Description of changes:*

Small fix for PR #702 where the outcome job (All E2E Tests Passed) fails for PRs from the same repo and not from forks (e.g. dependabot PRs). Non-blocking since "E2E Tests (Untrusted) / All E2E Tests Passed (pull_request_target)" is not required in same repo PRs (see #748).

*Testing:*

Before: https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/23645789481
image

After: https://github.com/jet-tong/mountpoint-s3-csi-driver/actions/runs/23745772025
image


.github/workflows/e2e-test-untrusted.yaml
```diff
@@ -43,7 +43,7 @@ jobs:
     secrets: inherit
   outcome:
     name: All E2E Tests Passed
-    if: always()
+    if: ${{ github.event.pull_request.head.repo.id != github.event.pull_request.base.repo.id }}
```
yerzhan7: Shouldn't we keep `always()`?

```
${{ always() && github.event.pull_request.head.repo.id != github.event.pull_request.base.repo.id }}
```

So that fork PRs with failing tests explicitly fails?

jet-tong: Good call! I was thinking as long as the if statement evaluates to true it'll always run. If I understand this right if tests fail and there's no always(), this job would be skipped since `needs` jobs have failed?

Will fix this now.

yerzhan7 (approved):

PR #773, opened 2026-04-14
Group dependabot PRs by package-ecosystem

jet-tong:
*Issue #, if available:*

*Description of changes:*

Group dependabot PRs to reduce PR noise. This also could mitigate concurrency group being 1 problem, when 3+ dependabot PRs rebase together there'll be cancelled runs. The groupings are:

- `github-actions` - all action bumps in one PR
- `gomod` (`/tests/e2e-kubernetes`) - all e2e test Go dep bumps in one PR
- `gomod` (root `/`) - all production Go dep bumps in one PR
  - I rarely see these PRs (see [v2.5.0 go.mod blame](https://github.com/awslabs/mountpoint-s3-csi-driver/blame/release-2.5/go.mod)) except for security issues (e.g. https://github.com/awslabs/mountpoint-s3-csi-driver/pull/751), so this is a new group that'll create new PRs - we may or may not want this. Also consider reducing frequency to monthly for this one. 

Update 2026-05-01: 

Following this PR https://github.com/awslabs/mountpoint-s3-csi-driver/pull/786 we won't absolutely require aggressive grouping like I proposed before, but it would be nice to keep them for maintainability which might result in e.g. 1 gh-actions dependabot PR, 2 gomod PRs (root / e2e) per week. 

Plan: keep github actions major changes since they're quite frequent, but separate major version changes for gomod groups. (**X**.y.z)
```
groups:
  minor-and-patch:
    update-types:
      - "minor"
      - "patch"
```

Existing open PRs would have to be closed for dependabot to re-open next Monday ofc after merge. 

Reference: https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#groups--

*Testing:* 

See https://github.com/jet-tong/mountpoint-s3-csi-driver/pulls for the PRs it generated.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,23 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    groups:
+      all-actions:
+        patterns:
+          - "*"
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      all-root-deps:
+        patterns:
+          - "*"
+  - package-ecosystem: "gomod"
+    directory: "/tests/e2e-kubernetes"
+    schedule:
+      interval: "weekly"
+    groups:
+      all-e2e-deps:
+        patterns:
+          - "*"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,7 @@ updates:
     schedule:
       interval: "weekly"
     groups:
-      all-actions:
+      gh-actions:
         patterns:
           - "*"
   - package-ecosystem: "gomod"
@@ -14,14 +14,16 @@ updates:
     schedule:
       interval: "weekly"
     groups:
-      all-root-deps:
-        patterns:
-          - "*"
+      gomod-root-minor-and-patch:
+        update-types:
+          - "minor"
+          - "patch"
   - package-ecosystem: "gomod"
     directory: "/tests/e2e-kubernetes"
     schedule:
       interval: "weekly"
     groups:
-      all-e2e-deps:
-        patterns:
-          - "*"
+      gomod-e2e-minor-and-patch:
+        update-types:
+          - "minor"
+          - "patch"


.github/dependabot.yml
```diff
@@ -5,3 +5,23 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    groups:
+      all-actions:
+        patterns:
+          - "*"
+  - package-ecosystem: "gomod"
```
yerzhan7, line 12: I recall there was some issue with `gomod` thingy. See comment in  https://github.com/awslabs/mountpoint-s3-csi-driver/pull/606#issuecomment-3433183401

Is it resolved?

jet-tong: Thanks good call - let me check/test.

jet-tong: Test runs are here: https://github.com/jet-tong/mountpoint-s3-csi-driver/pulls - there is a pretty big list of updates that was built up. 

I see now about the comment - in my current version it groups all dependabot PRs into 1 for those 3 groups, it doesn't separate major version / minor versions so we've still yet to discuss what to keep separate so we notice when some of them are breaking changes to our repo.

If we keep X.0.0 dependabot (major version) as separate PRs, and 0.X.0/X.0.0 golang package updates (minor+major) as separate PRs, it would actually basically mean all of those dependabot updates would be separate in https://github.com/jet-tong/mountpoint-s3-csi-driver/pull/1 and https://github.com/jet-tong/mountpoint-s3-csi-driver/pull/2 / https://github.com/jet-tong/mountpoint-s3-csi-driver/pull/3 - so it would help the CI situation here (but may be good for long-term).

Question: How often do go.mod / dependabot version bumps are breaking changes where we need to manually update code? Perhaps we can adjust separation based on that, or even not separate them at all? (i.e. keep this PR as is, v.s. adding `minor-and-patch:` flag)

jet-tong: Chatted in DM - will just add this patch to both gomod groups, and still group major gh-actions changes and minor gomod changes for easier maintainability. 
```
groups:
  minor-and-patch:
    update-types:
      - "minor"
      - "patch"
```

yerzhan7 (approved):

yerzhan7 (approved):

PR #786, opened 2026-05-01
Queue concurrency group runs not cancel with `queue:max`

jet-tong:
*Issue #, if available:*

*Description of changes:* 

Github Team [shipped `queue: max`](https://github.com/orgs/community/discussions/12835#discussioncomment-16602100) around 2 weeks ago which would solve our concurrency group problems. 

Add `queue: max` to concurrency groups - pending runs should now queue in FIFO order instead of cancelling each other. 

Ref: https://github.com/orgs/community/discussions/12835

*Testing:*

I [tested this in my repo](https://github.com/jet-tong/mountpoint-s3-csi-driver/actions/runs/25215331763) in trusted environment 
image

*Points that suggest risk:*

- They shipped it without public announcement
- I don't see it in [documentation](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency)
- Some problems raised in discussion (but I don't think either of them is blocking for us)
- Recent Github problems suggest we should be cautious about applying this

### DIFF
--- a/.github/workflows/delete-cluster.yaml
+++ b/.github/workflows/delete-cluster.yaml
@@ -8,6 +8,7 @@ on:
 
 concurrency:
   group: e2e-cluster-${{ inputs.environment }}
+  queue: max
 
 env:
   IMAGE_NAME: "s3-csi-driver"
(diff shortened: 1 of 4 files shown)


yerzhan7 (approved):
🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉

muddyfish (approved):

PR #790, opened 2026-05-08
Add daemonset mounter binary

vladem:
*Description of changes:*

Introduces `aws-s3-csi-daemonset-mounter`, a long-lived process that manages multiple Mountpoint instances within a single pod. The driver sends mount options and a FUSE fd over a Unix domain socket; the mounter spawns one Mountpoint child per connection.

Key changes:

1. New binary under cmd/aws-s3-csi-daemonset-mounter with accept loop, connection handler, and process manager
2. RecvOnConn now accepts a deadline to prevent stuck clients from blocking the accept loop
3. Tests for fd-leak absence on recv errors (bad JSON, extra fds, timeout) and across the full handleConnection lifecycle
4. Dockerfile and Makefile wired up for the new binary

---


cmd/aws-s3-csi-daemonset-mounter/process_manager.go
```diff
@@ -0,0 +1,212 @@
...
+		}
+		pw.w.Write([]byte(pw.prefix))
+		pw.w.Write(line)
+		pw.w.Write([]byte("\n"))
+	}
+	return len(p), nil
+}
+
+// LogStatusPeriodically logs the number of tracked and actual child processes at the given interval.
+func (pm *ProcessManager) LogStatusPeriodically(interval time.Duration) {
```
jet-tong, line 159: It's great to log them during development but maybe log them less frequently in the future - perhaps we maybe want to integrate this better with metrics to track Mountpoint processes later on?

cmd/aws-s3-csi-daemonset-mounter/process_runner.go
```diff
@@ -0,0 +1,64 @@
...
+// ProcessRunner starts a command and returns a handle to wait on it.
+type ProcessRunner interface {
+	Start(cmd *exec.Cmd) (ProcessHandle, error)
+}
+
+// defaultProcessRunner is the real implementation that starts OS processes.
+type defaultProcessRunner struct{}
+
+func (r *defaultProcessRunner) Start(cmd *exec.Cmd) (ProcessHandle, error) {
+	var stderrBuf bytes.Buffer
```
jet-tong: (AI flagged): This buffer grows for the lifetime of each mount-s3 process, consider capping at last N bytes (e.g. 64KB tail buffer). 

I don't expect it to be as big of a problem though since it's like that in the pod mounter binary too but with --debug + multiple mountpoint instances running in one pod this may be worth flagging.

cmd/aws-s3-csi-daemonset-mounter/process_runner.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (h *realProcessHandle) Pid() int { return h.cmd.Process.Pid }
+
+func (h *realProcessHandle) Wait() (int, []byte) {
+	err := h.cmd.Wait()
+	exitCode := 0
+	if err != nil {
+		if exitErr, ok := err.(*exec.ExitError); ok {
+			exitCode = exitErr.ExitCode()
+		}
```
jet-tong, line 65: (AI Flagged): if cmd.Wait() returns a non-ExitError (rare, e.g. pipe I/O failure), exitCode stays 0 and no .error file is written. The primary side would time out with a generic message instead of the real error. Consider treating any non-nil error as failure (return 1, stderrBuf.Bytes()).

cmd/aws-s3-csi-daemonset-mounter/process_runner.go
```diff
@@ -0,0 +1,64 @@
...
+	} else {
+		cmd.Stderr = &stderrBuf
+	}
+	if err := cmd.Start(); err != nil {
+		return nil, err
+	}
+	return &realProcessHandle{cmd: cmd, stderrBuf: &stderrBuf}, nil
+}
+
+type realProcessHandle struct {
```
jet-tong: np: When reading `realProcessHandle` I was thinking there's a non-real one somewhere but it's only in the tests - perhaps defaultProcessHandle (like defaultProcessRunner) or just processHandle?

cmd/aws-s3-csi-daemonset-mounter/main.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// Periodic observability: log number of tracked and actual child processes
+	go pm.LogStatusPeriodically(30 * time.Second)
+
+	// Accept loop — sequential, kernel backlog queues concurrent requests
+	for {
+		conn, err := listener.Accept()
+		if err != nil {
+			// Check if listener was closed (shutdown)
+			if opErr, ok := err.(*net.OpError); ok && opErr.Err.Error() == "use of closed network connection" {
```
jet-tong: np: I checked where the error string came from, and it seems it could use net package and replace with `errors.Is(err, net.ErrClosed)` which might be slightly cleaner? ([source](https://github.com/golang/go/issues/4373))

jet-tong (commented):
Tested with rough driver / primary daemonset code, multi mount is working.

pkg/mountpoint/mountoptions/mount_options.go
```diff
@@ -137,38 +152,52 @@ func Recv(ctx context.Context, sockPath string) (Options, error) {
...
 
-		messageN, unixRightsN, _, _, err := unixConn.ReadMsgUnix(message, unixRights)
+		messageN, unixRightsN, _, _, err := conn.ReadMsgUnix(message, unixRights)
 		if err != nil {
 			if errors.Is(err, io.EOF) {
 				break
 			}
-
-			return Options{}, fmt.Errorf("failed to read message from unix socket %s: %w", sockPath, err)
+			return Options{}, fmt.Errorf("failed to read message from connection: %w", err)
```
jet-tong, line 161: (AI Flagged) can add `closeFds(unixRightsBuf)` here to prevent fd leak.

jet-tong (dismissed):

jet-tong (approved):

PR #797, opened 2026-05-20
Add DaemonsetMounter for node-side integration with mounter binary

jet-tong:
## Description

This PR adds `DaemonsetMounter` to the 'primary' node daemonset, which along with [PR #790](https://github.com/awslabs/mountpoint-s3-csi-driver/pull/790)'s 'secondary' mounter daemonset, forms the MVP for DaemonSet-based architecture aimed to resolve scheduler problems with pod mode (v2) architecture. 

> [!WARNING]
> Note that this MVP still has multiple gaps and requires further work to make daemonset architecture a complete product.

### Summary

- `DaemonsetMounter`
  - `Mount` flow like `pod_mounter.go`: `IsMountPoint` check (for RequiresRepublish) -> `GetCommDir` (communicate with mounter pod) -> FUSE mount -> Send fd+options -> `waitForMount` (poll for readiness)
  - `commDir` lifecycle: 
    - `DiscoverCommDir` (at startup): retries `tryDiscoverCommDir` until the secondary mounter daemonset is discovered, returning the path to the emptyDir `commDir` for communication
    - `StartCommDirWatch` (background goroutine) calls `checkCommDir` every 5s to detect stale path and run `tryDiscoverCommDir` to rediscover and update `commDir`. 
    - `Mount` flow calls `GetCommDir` to get the atomic pointer `commDir`. If `mountoptions.Send` fails, it will signal `rediscoverCh` to wake the watcher to try to rediscover commDir.
- `driver.go` separates pod / daemonset mounter modes
- Helm charts
  - Mounter DaemonSet template (`mounter-daemonset.yaml`)
    - `OnDelete` strategy used, since restarting a mounter pod will disrupt ALL active mounts on that node, we want users to manually control when to update the mounter daemonset pods. 
  - RBAC Role granting the node service account permission to list pods in its own namespace

We chose to use a startup discoveries followed by background goroutine to maintain `commDir`, instead of lazy discovery via NodePublishVolume (NPV), for these reasons:
- NPV doesn't get blocked / slowed down
- Clear atomic management of commDir in the architecture (write only happens via `StartCommDirWatch`, and NPV only reads the commDir.
- Mount errors are surfaced to kubelet for visibility (e.g. mounter is restarted, commDir not found yet), as kubelet retries have better visibility than our own retries

<details>
<summary>Other Changes: Testing / Dev changes (Expand to view)</summary>

- `daemonset_mounter_test.go`
  - Mount tests (correctly passes mount options, no duplicate mounts, unmount source with send failure / fails with error file)
  - Unmount test
  - commDir lifecycle tests (rejects invalid pod states, goroutine stops on channel close, Mount fails if commDir not discovered, Mount nils commDir on staleness, cancelled context doesn't cause stale commDir). We can reduce these tests / adjust based on what e2e tests will cover.
- `assert.Contains` + `assert.ErrorIs` test helpers
- `mp-dev.sh` development helpers for daemonset mode
- Basic E2E testing
  - daemonset e2e tests with basic test suites:
    - `testsuites.InitVolumesTestSuite`
    - `custom_testsuites.InitS3AccessModeTestSuite`
    - `custom_testsuites.InitS3MountOptionsTestSuite`
    - `custom_testsuites.InitS3CSIMultiVolumeTestSuite`
  - Note we closed https://github.com/awslabs/mountpoint-s3-csi-driver/pull/821 in favor of using a feature branch for daemonset architecture development. 

I've also manually tested ROSA (since we have not implemented IRSA yet) by enabling `hostNetwork: true` (NOT recommended for users). IMDS doesn't work on ROSA for us due to hop limit 1 vs EKS's hop limit 2. See the [PR](https://github.com/awslabs/mountpoint-s3-csi-driver/pull/825) to temp skip ROSA tests for daemonset feature branch for more details. 
```sh
Ran 11 of 42 Specs in 217.320 seconds
SUCCESS! -- 11 Passed | 0 Failed | 0 Pending | 31 Skipped
PASS
ok      github.com/awslabs/mountpoint-s3-csi-driver/tests/e2e-kubernetes        217.375s
```

</details>

Please refer to [commit messages](https://github.com/awslabs/mountpoint-s3-csi-driver/pull/797/commits) for more details on the designs.

Requires:
- https://github.com/awslabs/mountpoint-s3-csi-driver/pull/790 (Merged)
- https://github.com/awslabs/mountpoint-s3-csi-driver/pull/825
  - IMDS doesn't work on ROSA for us due to hop limit 1 and no `hostNetwork: true` - see more context in the PR
- https://github.com/awslabs/mountpoint-s3-csi-driver/pull/827 (small fix for PR#825 untrusted yaml changes)

Not (yet) included in this PR:
- Daemonset mode specific e2e testing (commDir rediscovery, mounting edge cases etc)
- Documentation updates (we can add docs as this MVP matures)

What this MVP doesn't cover:
- Concurrent accept loop in secondary mounter to lower sendOptionsTimeout
- User agent update
- Recovery from corrupted mountpoint
- Upgrade/downgrade paths
- Credentials (IRSA / EKS Pod Identity / K8s secrets, and pod/volume level credentials)
- Pod/volume sharing

### DIFF
--- a/charts/aws-mountpoint-s3-csi-driver/templates/NOTES.txt
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/NOTES.txt
@@ -5,4 +5,13 @@ Learn more about the file system operations Mountpoint supports: https://github.
 {{- if .Release.IsUpgrade }}
 
 While the CSI driver has been upgraded, existing Mountpoint Pods remain unchanged to avoid disrupting your active workloads. New Mountpoint Pods with the updated version are created only when workload pods are started or restarted. You can view which pods exist created by older V2 versions of the CSI driver using `kubectl get pods -n mount-s3 --label-columns="s3.csi.aws.com/mounted-by-csi-driver-version"`.
+{{- if eq .Values.experimental.mounterMode "daemonset"}}
+
+IMPORTANT (daemonset mounter mode):
+After helm upgrade, s3-csi-daemonset-mounter pods will continue running the previous version (OnDelete update strategy). 
+This is intentional, since restarting a mounter pod will disrupt ALL active mounts on that node. 
+
+To roll out the update to mounter pods (temporarily breaks ALL active mounts): 
+kubectl delete pod -n {{.Release.Namespace }} -l app=s3-csi-daemonset-mounter
+{{- end }}
 {{- end }}
--- a/charts/aws-mountpoint-s3-csi-driver/templates/NOTES.txt
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/NOTES.txt
@@ -2,16 +2,15 @@ Thank you for using Mountpoint for Amazon S3 CSI Driver v{{ .Chart.Version }}.
 
 Learn more about the file system operations Mountpoint supports: https://github.com/awslabs/mountpoint-s3/blob/main/doc/SEMANTICS.md
 
-{{- if .Release.IsUpgrade }}
-
-While the CSI driver has been upgraded, existing Mountpoint Pods remain unchanged to avoid disrupting your active workloads. New Mountpoint Pods with the updated version are created only when workload pods are started or restarted. You can view which pods exist created by older V2 versions of the CSI driver using `kubectl get pods -n mount-s3 --label-columns="s3.csi.aws.com/mounted-by-csi-driver-version"`.
 {{- if eq .Values.experimental.mounterMode "daemonset"}}
 
-IMPORTANT (daemonset mounter mode):
-After helm upgrade, s3-csi-daemonset-mounter pods will continue running the previous version (OnDelete update strategy). 
-This is intentional, since restarting a mounter pod will disrupt ALL active mounts on that node. 
+IMPORTANT: After helm upgrade, mounter pods continue running the previous version (OnDelete strategy) to avoid disrupting active mounts. 
+Node recycling (EKS node group update, Karpenter drift/expiration) completes the upgrade seamlessly. 
+If node recycling is not feasible, manually delete the mounter pod - this will temporarily break all active mounts on that node:
+kubectl delete pod -n {{ .Release.Namespace }} -l app=s3-csi-daemonset-mounter --field-selector spec.nodeName=<node-name>
+{{- else if eq .Values.experimental.mounterMode "pod" }}
 
-To roll out the update to mounter pods (temporarily breaks ALL active mounts): 
-kubectl delete pod -n {{.Release.Namespace }} -l app=s3-csi-daemonset-mounter
+{{- if .Release.IsUpgrade }}
+While the CSI driver has been upgraded, existing Mountpoint Pods remain unchanged to avoid disrupting your active workloads. New Mountpoint Pods with the updated version are created only when workload pods are started or restarted. You can view which pods exist created by older V2 versions of the CSI driver using `kubectl get pods -n mount-s3 --label-columns="s3.csi.aws.com/mounted-by-csi-driver-version"`.
 {{- end }}
 {{- end }}
--- a/charts/aws-mountpoint-s3-csi-driver/templates/mount-s3-namespace.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/mount-s3-namespace.yaml
@@ -1,3 +1,4 @@
+{{- if ne .Values.experimental.mounterMode "daemonset" }}
 {{- if .Values.mountpointPod.createNamespace }}
 ---
 apiVersion: v1
@@ -7,4 +8,5 @@ metadata:
   labels:
     name: {{ .Values.mountpointPod.namespace }}
     pod-security.kubernetes.io/enforce: restricted
+{{- end }}
 {{- end }}
--- a/charts/aws-mountpoint-s3-csi-driver/templates/mount-s3-namespace.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/mount-s3-namespace.yaml
@@ -1,4 +1,3 @@
-{{- if ne .Values.experimental.mounterMode "daemonset" }}
 {{- if .Values.mountpointPod.createNamespace }}
 ---
 apiVersion: v1
@@ -8,5 +7,4 @@ metadata:
   labels:
     name: {{ .Values.mountpointPod.namespace }}
     pod-security.kubernetes.io/enforce: restricted
-{{- end }}
 {{- end }}
--- a/charts/aws-mountpoint-s3-csi-driver/templates/mounter-daemonset.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/mounter-daemonset.yaml
@@ -1,4 +1,7 @@
 {{- if eq .Values.experimental.mounterMode "daemonset" }}
+{{- if gt (len .Values.daemonsetMounters) 1 }}
+{{- fail "Only one entry in daemonsetMounters is supported currently" }}
+{{- end }}
 {{- $mounter := index .Values.daemonsetMounters 0 }}
 kind: DaemonSet
 apiVersion: apps/v1
--- a/charts/aws-mountpoint-s3-csi-driver/templates/mounter-daemonset.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/mounter-daemonset.yaml
@@ -3,7 +3,7 @@
 kind: DaemonSet
 apiVersion: apps/v1
 metadata:
-  name: aws-s3-csi-daemonset-mounter
+  name: s3-csi-daemonset-mounter
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "aws-mountpoint-s3-csi-driver.labels" . | nindent 4 }}
@@ -12,12 +12,12 @@ spec:
     type: OnDelete
   selector:
     matchLabels:
-      app: aws-s3-csi-daemonset-mounter
+      app: s3-csi-daemonset-mounter
       {{- include "aws-mountpoint-s3-csi-driver.selectorLabels" . | nindent 6 }}
   template:
     metadata:
       labels:
-        app: aws-s3-csi-daemonset-mounter
+        app: s3-csi-daemonset-mounter
         {{- include "aws-mountpoint-s3-csi-driver.labels" . | nindent 8 }}
         {{- with $mounter.podLabels }}
         {{- toYaml . | nindent 8 }}
--- a/charts/aws-mountpoint-s3-csi-driver/templates/mounter-daemonset.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/mounter-daemonset.yaml
@@ -2,7 +2,7 @@
 kind: DaemonSet
 apiVersion: apps/v1
 metadata:
-  name: s3-csi-mounter
+  name: aws-s3-csi-daemonset-mounter
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "aws-mountpoint-s3-csi-driver.labels" . | nindent 4 }}
@@ -11,12 +11,12 @@ spec:
     type: OnDelete
   selector:
     matchLabels:
-      app: s3-csi-mounter
+      app: aws-s3-csi-daemonset-mounter
       {{- include "aws-mountpoint-s3-csi-driver.selectorLabels" . | nindent 6 }}
   template:
     metadata:
       labels:
-        app: s3-csi-mounter
+        app: aws-s3-csi-daemonset-mounter
         {{- include "aws-mountpoint-s3-csi-driver.labels" . | nindent 8 }}
         {{- with .Values.mounterDaemonset.podLabels }}
         {{- toYaml . | nindent 8 }}
--- a/charts/aws-mountpoint-s3-csi-driver/templates/mounter-daemonset.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/mounter-daemonset.yaml
@@ -1,0 +1,76 @@
+{{- if eq .Values.experimental.mounterMode "daemonset" }}
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: s3-csi-mounter
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "aws-mountpoint-s3-csi-driver.labels" . | nindent 4 }}
+spec:
+  updateStrategy:
+    type: OnDelete
+  selector:
+    matchLabels:
+      app: s3-csi-mounter
+      {{- include "aws-mountpoint-s3-csi-driver.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      labels:
+        app: s3-csi-mounter
+        {{- include "aws-mountpoint-s3-csi-driver.labels" . | nindent 8 }}
+        {{- with .Values.mounterDaemonset.podLabels }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+    spec:
+      nodeSelector:
+        kubernetes.io/os: linux
+        {{- with .Values.mounterDaemonset.nodeSelector }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+      priorityClassName: system-node-critical
+      {{- with .Values.mounterDaemonset.affinity }}
+      affinity: {{- toYaml . | nindent 8 }}
+      {{- end }}
+      tolerations:
+        {{- if .Values.mounterDaemonset.tolerateAllTaints }}
+        - operator: Exists
+        {{- end }}
+        {{- with .Values.mounterDaemonset.tolerations }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+      {{- with .Values.mounterDaemonset.imagePullSecrets }}
+      imagePullSecrets:
+      {{- range . }}
+        - name: {{ . }}
+      {{- end }}
+      {{- end }}
+      containers:
+      - name: mounter
+        image: {{ include "csiDriverImageName" . }}
+        imagePullPolicy: {{ .Values.image.pullPolicy }}
+        command:
+          - /bin/aws-s3-csi-daemonset-mounter
+        args:
+          - --comm-dir=/comm
+          - --mountpoint-bin-dir=/mountpoint-s3/bin
+          - --v={{ .Values.node.logLevel }}
+        securityContext:
+          runAsUser: 1000
+          runAsNonRoot: true
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
+          capabilities:
+            drop: ["ALL"]
+          seccompProfile:
+            type: RuntimeDefault
+        volumeMounts:
+          - name: comm
+            mountPath: /comm
+        {{- with .Values.mounterDaemonset.resources }}
+        resources:
+          {{- toYaml . | nindent 12 }}
+        {{- end }}
+      volumes:
+      - name: comm
+        emptyDir: {}
+{{- end }}
--- a/charts/aws-mountpoint-s3-csi-driver/templates/mounter-daemonset.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/mounter-daemonset.yaml
@@ -12,6 +12,8 @@ metadata:
     {{- include "aws-mountpoint-s3-csi-driver.labels" . | nindent 4 }}
 spec:
   updateStrategy:
+    # We use OnDelete strategy because restarting this pod disrupts ALL active mounts on the node.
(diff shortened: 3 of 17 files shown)


pkg/driver/driver.go
```diff
@@ -117,28 +118,38 @@ func NewDriver(endpoint string, mpVersion string, nodeID string) (*Driver, error
...
-	mpMounter := mpmounter.New()
-	podWatcher := watcher.New(clientset, mountpointPodNamespace, nodeID, podWatcherResyncPeriod)
-	err = podWatcher.Start(stopCh)
-	if err != nil {
-		klog.Fatalf("Failed to start Pod watcher: %v\n", err)
-	}
+	var nodeMounter mounter.Mounter
 
-	s3paCache := setupS3PodAttachmentCache(config, stopCh, nodeID, kubernetesVersion)
+	if mounterMode == "daemonset" {
```
vladem: nit: make "daemonset" a constant defined at the top of the file

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -0,0 +1,355 @@
...
+	"github.com/awslabs/mountpoint-s3-csi-driver/pkg/mountpoint/mountoptions"
+	"github.com/awslabs/mountpoint-s3-csi-driver/pkg/util"
+)
+
+const (
+	// Name of the emptyDir volume on the secondary daemonset used for communication.
+	commVolumeName = "comm"
+
+	// Label used to identify the secondary (mounter) daemonset pods.
+	mounterPodLabel = "app=s3-csi-mounter"
```
vladem: lets use a more specific label, maybe `aws-s3-csi-daemonset-mounter`?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -0,0 +1,336 @@
...
+			}
+			sockPath = filepath.Join(commDir, "mount.sock")
+			errFilePath = filepath.Join(commDir, mountId+".error")
+			err = mountoptions.Send(sendCtx, sockPath, mountoptions.Options{
+				Fd:         fd,
+				BucketName: bucketName,
+				Args:       args.SortedList(),
+				Env:        env.List(),
+				VolumeId:   mountId,
+			})
```
vladem: For the sake of simplicity I'd remove this retry, we can rely on kubelet retrying `NodePublishVolume` (having a clarifying comment is good). Just invalidating the commDir should be fine. (Alternative -- make it a loop with 2 iterations.)

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -0,0 +1,336 @@
...
+	if dm.commDir != "" {
+		sockPath := filepath.Join(dm.commDir, "mount.sock")
+		if _, err := os.Stat(sockPath); err == nil {
+			return dm.commDir, nil
+		}
+		klog.V(4).Infof("DaemonsetMounter: cached comm dir is stale, re-discovering")
+		dm.commDir = ""
+	}
+
+	commDir, err := dm.discoverCommDir(ctx)
```
vladem: so this function may retry up to 30s and we'll keep holding a lock for all that time? may be a bit hard to debug, a mechanism to consider: sync discovery on driver's startup with retries and a goroutine periodically checking that socket exists and doing rediscovery if needed, on mount path we just load `commDir atomic.Pointer[string]` (if nil return an error)

returning an error on NodePublishVolume when re-discovery is in progress is arguably better behaviour (customers see an error in `kubectl pod describe` instead of requests taking 30s to discover and timeout)

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -0,0 +1,336 @@
...
+	mountResultCh := make(chan error, 2)
+
+	// Poll for error file
+	go func() {
+		wait.PollUntilContextCancel(ctx, daemonsetMountPollInterval, true, func(ctx context.Context) (bool, error) {
+			content, err := os.ReadFile(errFilePath)
+			if err != nil {
+				return false, nil
+			}
+			os.Remove(errFilePath)
```
vladem, line 437: lets attempt errFile deletion on `Unmount` too (for the case when mount succeeded, but MP crashed afterwards)

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -0,0 +1,336 @@
...
+	env.Merge(envprovider.Default())
+
+	// Move --aws-max-attempts to env if provided
+	if maxAttempts, ok := args.Remove(mountpoint.ArgAWSMaxAttempts); ok {
+		env.Set(envprovider.EnvMaxAttempts, maxAttempts)
+	}
+
+	// Step 2: Send mount options to secondary daemonset
+	sockPath := filepath.Join(commDir, "mount.sock")
+	errFilePath := filepath.Join(commDir, mountId+".error")
```
vladem: lets make these 2 strings constants defined at the top of the file

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -0,0 +1,336 @@
...
+	// Mount identifier must be unique per mount (not per volume), since multiple pods
+	// can mount the same volume. Use workload pod UID to distinguish.
+	// TODO: change for pod/volume sharing
+	podUID := credentialCtx.WorkloadPodID
+	volumeId := credentialCtx.VolumeID
+	mountId := podUID + "-" + volumeId
+
+	// Idempotency: if target is already a healthy Mountpoint mount, return early.
+	// Kubelet may call NodePublishVolume repeatedly (requiresRepublish, retries).
+	if isMounted, err := dm.IsMountPoint(target); err == nil && isMounted {
```
vladem: additionally lets return an err if this check failed

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -0,0 +1,336 @@
...
+	// Step 2: Send mount options to secondary daemonset
+	sockPath := filepath.Join(commDir, "mount.sock")
+	errFilePath := filepath.Join(commDir, mountId+".error")
+
+	// Remove old error file if exists
+	os.Remove(errFilePath)
+
+	klog.V(4).Infof("DaemonsetMounter: sending mount options for volume %s (mount %s) to %s", volumeId, mountId, sockPath)
+
+	sendCtx, sendCancel := context.WithTimeout(ctx, daemonsetMountReadyTimeout)
```
vladem: whats the value of this timeout in pod_mounter?

charts/aws-mountpoint-s3-csi-driver/values.yaml
```diff
@@ -135,3 +135,21 @@ experimental:
   reserveHeadroomForMountpointPods: false
   headroomPodImage: public.ecr.aws/eks-distro/kubernetes/pause:3.10
   headroomPodLabels: {}
+  # Mounter mode: "pod" (V2 default) or "daemonset" (experimental two-daemonset architecture).
```
vladem, line 140: Lets clarify in the comment that implementation of the `daemonset` mode is not complete yet and exposed in the helm chart only to allow testing.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -0,0 +1,336 @@
...
+
+// discoverCommDir finds the secondary mounter pod on this node and returns the path
+// to its emptyDir comm volume as seen from the primary daemonset (via kubelet pod dir).
+// It retries until secondary daemonset mounter appears in case primary starts before secondary.
+func (dm *DaemonsetMounter) discoverCommDir(ctx context.Context) (string, error) {
+	ctx, cancel := context.WithTimeout(ctx, daemonsetMountReadyTimeout)
+	defer cancel()
+
+	for {
+		pods, err := dm.clientset.CoreV1().Pods("").List(ctx, metav1.ListOptions{
```
vladem: we're listing pod in all namespaces here, this requires broader RBAC than we need, lets adjust:
1. this operation to use a specific namespace (pass `.Release.Namespace` as an ENV)
2. reduce RBAC scope of `node.serviceAccount`

also consider testing that there is no change in templates rendered with default parameters as compared to main: 
```
helm template aws-mountpoint-s3-csi-driver --namespace kube-system charts/aws-mountpoint-s3-csi-driver
```

charts/aws-mountpoint-s3-csi-driver/values.yaml
```diff
@@ -135,3 +135,21 @@ experimental:
...
+  # In daemonset mode, one mounter daemonset pod handles all mountpoint processes instead of 
+  # one mountpoint pod per workload in v2. The controller, CRD, and mount-s3 namespace are not deployed.
+  mounterMode: "pod"
+
+# Configuration for the secondary mounter daemonset.
+# Only used when experimental.mounterMode is "daemonset".
+mounterDaemonset:
```
vladem: make it `daemonsetMounters` of array type, to be able to support multiple daemonset configs in future if we need to

charts/aws-mountpoint-s3-csi-driver/templates/mount-s3-namespace.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if ne .Values.experimental.mounterMode "daemonset" }}
```
vladem: lets reduce chart changes: temporarily disable only controller deployment (it requires changes, we'll address while adding V2 mounts maintenance to new mode), other things are deployed: mount-s3-namespace, pod-priority, pod attachments and even controller's SA

jet-tong: Good call. (I added this previously cause I was getting `namespaces mount-s3 not found` errors on helm validate in e2e tests - should've fixed it another way)

pkg/driver/node/mounter/daemonset_mounter.go
jet-tong: Good call, 15s for pod_mounter (pod attachment + MP readiness). Probably much faster in daemonset - I'll use 15s to match for now but can potentially be shorter.
```go
const (
	mountpointPodAttachmentWaitDuration = 15 * time.Second
	mountpointPodAttachmentPollInterval = 250 * time.Millisecond

	mountpointPodReadinessWaitDuration = 15 * time.Second
)
```

charts/aws-mountpoint-s3-csi-driver/templates/mounter-daemonset.yaml
```diff
@@ -0,0 +1,80 @@
...
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: s3-csi-daemonset-mounter
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "aws-mountpoint-s3-csi-driver.labels" . | nindent 4 }}
+spec:
+  updateStrategy:
+    type: OnDelete
```
vladem, line 17: please add a comment explaining the choice of the strategy (killing this pod disrupts active mounts, so it keeps running until manually deleted or the node is terminated), also we'd need to make this non-obvious update behaviour more prominent (may be an info log on driver startup and post-install message in helm)

charts/aws-mountpoint-s3-csi-driver/templates/mounter-daemonset.yaml
```diff
@@ -0,0 +1,80 @@
...
+        kubernetes.io/os: linux
+        {{- with $mounter.nodeSelector }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+      priorityClassName: system-node-critical
+      {{- with $mounter.affinity }}
+      affinity: {{- toYaml . | nindent 8 }}
+      {{- end }}
+      tolerations:
+        {{- if $mounter.tolerateAllTaints }}
```
vladem: I think we need default tolerations here too, which takes effect when tolerateAllTaints is false, similar to `.Values.node.defaultTolerations`

charts/aws-mountpoint-s3-csi-driver/templates/mounter-daemonset.yaml
```diff
@@ -0,0 +1,80 @@
...
+      containers:
+      - name: mounter
+        image: {{ include "csiDriverImageName" . }}
+        imagePullPolicy: {{ .Values.image.pullPolicy }}
+        command:
+          - /bin/aws-s3-csi-daemonset-mounter
+        args:
+          - --comm-dir=/comm
+          - --mountpoint-bin-dir=/mountpoint-s3/bin
+          - --v={{ .Values.node.logLevel }}
```
vladem: `.Values.node.logLevel` -> `$mounter.logLevel`

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -0,0 +1,442 @@
...
+const (
+	mounterPodLabel  = "app=s3-csi-daemonset-mounter"
+	CommVolumeName   = "comm"
+	MountSockName    = "mount.sock"
+	MountErrorSuffix = ".error"
+
+	// TODO: lower sendOptionsTimeout once secondary has concurrent accept to reduce blocks on Mount -> Send -> dialWithRetry
+	sendOptionsTimeout = 15 * time.Second
+
+	mountReadyTimeout      = 15 * time.Second
```
vladem: let's make it 2 minutes -- effectively maximum (NodePublishVolume timeout is 2 minutes); this is similar to pod_mounter, and I think longer wait is justified here: Mountpoint might be taking longer to startup, because of the internal retry logic (e.g. credential retrieval)

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -0,0 +1,442 @@
...
+	mount       *mpmounter.Mounter
+
+	// Comm dir discovery: commDir caches the path (nil = stale),
+	// rediscoverCh wakes the background watcher to re-discover immediately.
+	commDir      atomic.Pointer[string]
+	rediscoverCh chan struct{}
+
+	// Injectable for testing. nil = use default.
+	mountSyscall mountSyscallFunc
+	sendOptions  sendOptionsFunc
```
vladem: unused, let's remove (we never set this field)

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -0,0 +1,442 @@
...
+//  3. Waits for Mountpoint to start serving (or an error)
+func (dm *DaemonsetMounter) Mount(ctx context.Context, bucketName string, target string,
+	credentialCtx credentialprovider.ProvideContext, args mountpoint.Args, fsGroup string, userEnv envprovider.Environment) error {
+
+	// Mount identifier must be unique per mount (not per volume), since multiple pods
+	// can mount the same volume. Use workload pod UID to distinguish.
+	// TODO: change for pod/volume sharing
+	podUID := credentialCtx.WorkloadPodID
+	volumeId := credentialCtx.VolumeID
+	mountId := podUID + "-" + volumeId
```
vladem: nit: add helpers `GetMountId(podUID, volumeId)` and `GetErrorFileName(podUID, volumeId)`

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -0,0 +1,442 @@
...
+	err = dm.sendOptionsWithDefault(sendCtx, sockPath, mountoptions.Options{
+		Fd:         fd,
+		BucketName: bucketName,
+		Args:       args.SortedList(),
+		Env:        env.List(),
+		VolumeId:   mountId,
+	})
+	if err != nil {
+		// If send failed due to stale path, signal re-discovery and let Kubelet retry NodePublishVolume.
+		// TODO: add helpMessageForGettingMountpointLogs to help users on mount failures
```
vladem: are these todo-s addressed? can we remove them from code?

pkg/driver/node/mounter/daemonset_mounter_test.go
```diff
@@ -0,0 +1,505 @@
...
+			testCtx.mount.Mount("mountpoint-s3", target, "fuse", nil)
+
+			err := <-mountRes
+			assert.NoError(t, err)
+
+			gotFile := os.NewFile(uintptr(got.Fd), "fd")
+			t.Cleanup(func() { gotFile.Close() })
+			mountertest.AssertSameFile(t, devNull, gotFile)
+
+			for _, a := range got.Args {
```
vladem: let's verify here that `prefix` was received; also in `testCtx.mountSyscall` that `MountOptions.ReadOnly` was set

pkg/driver/node/mounter/daemonset_mounter_test.go
```diff
@@ -0,0 +1,505 @@
...
+			if mountSyscallCalled {
+				t.Error("mountSyscall should not be called for already-mounted target")
+			}
+		})
+
+		t.Run("Unmounts source if mounter does not receive mount options", func(t *testing.T) {
+			testCtx := setupDM(t)
+			target := filepath.Join(testCtx.kubeletPath, "target")
+
+			testCtx.mountSyscall = func(tgt string, opts mpmounter.MountOptions) (int, error) {
```
vladem: nit: use the default `mountSyscall` from `testCtx`

pkg/driver/node/mounter/daemonset_mounter.go
jet-tong: Good catch - previously added for staleness test to make Send function a bit faster by returning DeadlineExceeded immediately (Send -> dialWithRetry will retry until DeadlineExceeded), but didn't change it since it adds too many nil fields in NewDaemonsetMounter in driver.go.

pkg/driver/node/mounter/daemonset_mounter_test.go
jet-tong: Good catch, mistake from me not checking if all mount options are tested (was focused on making it work)

pkg/driver/node/mounter/daemonset_mounter.go
jet-tong: Will address 1st one. For the 2nd todo I was hoping for a comprehensive unit test covering edge cases, but I haven't - will remove comment (and can discuss here instead).

Aware that Send -> dialWithRetry retries ENOENT/ECONNREFUSED, but are these 3 errors checks enough for marking stale? Perhaps we want to mark staleness for unexpected errors? (Or NOT marking staleness to avoid breaking NPV calls, but signalling rediscovery returns early when `os.Stat(sockPath)` so it doesn't seem useful to do that).

charts/aws-mountpoint-s3-csi-driver/templates/NOTES.txt
```diff
@@ -5,4 +5,13 @@ Learn more about the file system operations Mountpoint supports: https://github.
 {{- if .Release.IsUpgrade }}
 
 While the CSI driver has been upgraded, existing Mountpoint Pods remain unchanged to avoid disrupting your active workloads. New Mountpoint Pods with the updated version are created only when workload pods are started or restarted. You can view which pods exist created by older V2 versions of the CSI driver using `kubectl get pods -n mount-s3 --label-columns="s3.csi.aws.com/mounted-by-csi-driver-version"`.
+{{- if eq .Values.experimental.mounterMode "daemonset"}}
```
vladem, line 5: I think we should print it always. Not only when `.Release.IsUpgrade`. Also I'd suggest something like:

```
IMPORTANT: After helm upgrade, mounter pods continue running the previous version (OnDelete strategy) to avoid disrupting active mounts. 
Node recycling (EKS node group update, Karpenter drift/expiration) completes the upgrade seamlessly. 
If node recycling is not feasible, manually delete the mounter pod — this will temporarily break all active mounts on that node:

kubectl delete pod -n {{ .Release.Namespace }} -l app=s3-csi-daemonset-mounter --field-selector spec.nodeName=<node-name>
```

pkg/driver/node/mounter/daemonset_mounter.go
vladem: `checkCommDir` handles the broad set of errors, its fine to handle a narrow set here (our main goal is to rediscover the socket path when mounter pod is deleted, ie when emptyDir is gone, and this is covered)

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -0,0 +1,442 @@
+// DaemonsetMounter implements Mounter for the two-daemonset architecture.
```
vladem: I'd suggest making this more readable for a person who looks at this code for the first time, e.g what is "primary", and keeping it brief -- just what this component does. Also would be good to cleanup todo-s before we merge.

jet-tong: I've updated it - let me know if I can further improve it!

Will keep this comment unresolved until I clean up the todos - good call.

vladem (approved):

PR #799, opened 2026-05-20
reconciler: recover from duplicate S3PodAttachments by deleting empty ones

yerzhan7:
*Issue:* N/A

*Description of changes:* 

**Problem**: In case of duplicate S3PAs with no MP pod attachments reconciler will get stuck with `found 2 MountpointS3PodAttachments when expecting 0 or 1` error.

**Solution**: Recover from duplicate S3PodAttachments by deleting empty ones directly in reconciler.

**Implementation**:
  - `getExistingS3PodAttachment`: on List returning >1 S3PA, hand off to `recoverFromDuplicateS3PodAttachments` instead of erroring.
  - `recoverFromDuplicateS3PodAttachments`:
    - Deletes duplicates that reference no Mountpoint Pods.
    - Returns the surviving S3PA if exactly one references Mountpoint Pods.
    - Returns `nil` if none did (caller re-creates via handleNewS3PodAttachment).
    - Errors if >1 references Mountpoint Pods -> return error as before (handling this case is out of scope of this PR/problem)
  - `setupLogger` refactor: logger is built before the List and enriched with the S3PA name afterward; no functional change.
  - Drive-by: Wrap Delete with a `ResourceVersion` precondition so a stale-cache `Delete` races to 409 instead of
  overwriting concurrent state. 

**Alternatives considered:**
 - Delete these duplicates in `stale_attachment_cleaner.go` periodic job. Discarded for now because: (we might consider doing in the future as well, in addition to this change)
    - In the worst case, customer workload will be stuck in the reconcile loop for up to 2min waiting for the periodic job cleanup
    - There are more edge cases to handle because we need to ensure that `Delete` (and clearing of expectation) does not interfere with reconciler. (Currently, it's not racing with reconciler because the cleaner only deletes if attachment is stale (>2min old))

### DIFF
--- a/cmd/aws-s3-csi-controller/csicontroller/reconciler.go
+++ b/cmd/aws-s3-csi-controller/csicontroller/reconciler.go
@@ -311,11 +311,14 @@ func (r *Reconciler) spawnOrDeleteMountpointPodIfNeeded(
 		return Requeue, err
 	}
 	fieldFilters := r.buildFieldFilters(workloadPod, pv, roleArn)
-	s3pa, err := r.getExistingS3PodAttachment(ctx, fieldFilters)
+	log := r.setupLogger(ctx, workloadPod, pvc, workloadUID, fieldFilters)
+	s3pa, err := r.getExistingS3PodAttachment(ctx, fieldFilters, log)
 	if err != nil {
 		return Requeue, err
 	}
-	log := r.setupLogger(ctx, workloadPod, pvc, workloadUID, fieldFilters, s3pa)
+	if s3pa != nil {
+		log = log.WithValues("s3pa", s3pa.Name)
+	}
 
 	if !isPodActive(workloadPod) {
 		return r.handleInactivePod(ctx, s3pa, workloadUID, fieldFilters, log)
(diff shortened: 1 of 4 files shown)


cmd/aws-s3-csi-controller/csicontroller/reconciler.go
```diff
@@ -420,8 +421,52 @@ func (r *Reconciler) getExistingS3PodAttachment(ctx context.Context, fieldFilter
...
+	for i := range duplicates {
+		if len(duplicates[i].Spec.MountpointS3PodAttachments) == 0 {
+			noMountpointPods = append(noMountpointPods, &duplicates[i])
+		} else {
+			withMountpointPods = append(withMountpointPods, &duplicates[i])
+		}
 	}
+
+	if len(withMountpointPods) > 1 {
+		return nil, fmt.Errorf("found %d MountpointS3PodAttachments referencing Mountpoint Pods when expecting 0 or 1", len(withMountpointPods))
```
vladem, line 448: is there a way to recover from this? this can happen in case of pod attachment's "timeout-but-actually-succeeded" creation (we don't `s3paExpectations.setPending(fieldFilters)` and attachment has a pod, next reconcile creates another attachment again with a pod => 2 attachments with a pod)

cmd/aws-s3-csi-controller/csicontroller/reconciler.go
```diff
@@ -420,8 +421,52 @@ func (r *Reconciler) getExistingS3PodAttachment(ctx context.Context, fieldFilter
...
+	}
+
+	if len(withMountpointPods) == 1 {
+		return withMountpointPods[0], nil
+	}
+
+	// All duplicates referenced no Mountpoint Pods AND every delete succeeded. Clear any stale pending
+	// creation expectation so the caller's handleNewS3PodAttachment can create a fresh S3PA instead of
+	// requeuing forever.
+	if len(noMountpointPods) > 0 && r.s3paExpectations.isPending(fieldFilters) {
```
vladem, line 465: there is a chance that PA just been created and not yet in cache and we clear expectations here? low risk?

vladem (approved):

cmd/aws-s3-csi-controller/csicontroller/reconciler.go
yerzhan7: It is possible to recover from it. It's more complex and out of scope of this PR.

cmd/aws-s3-csi-controller/csicontroller/reconciler.go
yerzhan7: That scenario should not be possible because it requires 2 outstanding Creates, which is not possible with single reconciler.

PR #814, opened 2026-06-01
Add controller leader election

yerzhan7:
*Issue #, if available:* N/A

*Description of changes:* During updates there is a chance that 2 controllers would co-exist briefly and would try to reconcile the same pod which would result in duplicate S3PA attachment. Fix this by introducing leader election so that only at most 1 controller is active at any time.

### DIFF
--- a/charts/aws-mountpoint-s3-csi-driver/templates/serviceaccount-csi-controller.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/serviceaccount-csi-controller.yaml
@@ -76,4 +76,36 @@ roleRef:
   kind: ClusterRole
   name: s3-csi-driver-controller-cluster-role
   apiGroup: rbac.authorization.k8s.io
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: s3-csi-driver-controller-leader-election-role
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "aws-mountpoint-s3-csi-driver.labels" . | nindent 4 }}
+rules:
+  - apiGroups: ["coordination.k8s.io"]
+    resources: ["leases"]
+    resourceNames: ["aws-s3-csi-controller"]
+    verbs: ["get", "update"]
+  - apiGroups: ["coordination.k8s.io"]
+    resources: ["leases"]
+    verbs: ["create"]
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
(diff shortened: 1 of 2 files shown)


muddyfish (commented):
What testing has been done? Is there an e2e test that we can do here?

yerzhan7:
> What testing has been done?

Performed one-off manual test:
- EKS cluster with 10 nodes, 10 PVs, 10 Deployments with 5 replica (50 pods) -> expecting 50 s3pas
- **Trigger**: `kubectl scale deploy/s3-csi-controller --replicas=2`
  - Note: This is not exactly the "2 pods during deployment rollout" scenario, but it allows for more consistent issue re-production of duplicate S3PAs
- **Before fix (v2.6.0):**
  - Total S3PAs: 68
  - Duplicates: 20
  - Error: `found 2 MountpointS3PodAttachments referencing Mountpoint Pods when expecting 0 or 1`
- **After the fix (this PR):**
  - Total S3PAs: 50
  - Duplicates: 0
  - No error messages in logs (both node and controller)
  - 50 Workloads and 50 MP Pods in Running state
  - No reconciler or stale attachment cleaner logs in the follower pod

Lease Object:
```yaml
metadata:
  name: aws-s3-csi-controller
  namespace: kube-system
spec:
  acquireTime: "2026-06-01T15:39:08.050405Z"
  holderIdentity: s3-csi-controller-5d8466fbd5-284dc_52fe04db-8186-4820-9633-698339937586
  leaseDurationSeconds: 15
```

Follower pod log:
```
I0601 15:39:08.583801  leaderelection.go:254] attempting to acquire leader lease kube-system/aws-s3-csi-controller...
```

Leader pod log:
```
I0601 15:39:08.050389  leaderelection.go:254] attempting to acquire leader lease kube-system/aws-s3-csi-controller...
I0601 15:39:08.146439  leaderelection.go:268] successfully acquired lease kube-system/aws-s3-csi-controller
```

Also, killed Leader Pod afterwards to validate auto-recovery (after 10s): `I0601 15:47:28.685192  successfully acquired lease kube-system/aws-s3-csi-controller`

> Is there an e2e test that we can do here?

Not sure we can add anything meaningful. We can add another controller test (`envtest`) with 2 controllers, but that would test the behaviour of leader-election/controller-runtime logic more rather than our code. Don't think it's worth the effort.

muddyfish (approved):
Approving without e2e tests given that the controller component is only needed for scheduling decisions for the MP pods, which goes away with the daemonset architecture approach. We may want to re-visit this if we ever need to work on a controller again for other reasons

PR #833, opened 2026-06-15
Update mp-dev cluster

renanmagagnin:
*Issue #, if available:* N/A

*Description of changes:* Update EKS version and instance type in cluster created by the `mp-dev.sh` script.

### DIFF
--- a/dev/mp-dev-cluster.yaml
+++ b/dev/mp-dev-cluster.yaml
@@ -4,7 +4,7 @@ kind: ClusterConfig
 metadata:
   name: mp-dev-cluster
   region: eu-north-1
-  version: "1.33"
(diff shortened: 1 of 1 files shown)


dev/mp-dev-cluster.yaml
```diff
@@ -3,8 +3,8 @@ kind: ClusterConfig
 
 metadata:
   name: mp-dev-cluster
-  region: eu-north-1
-  version: "1.33"
+  region: eu-west-1
```
jet-tong: I'd lean towards not changing the region - is there a reason why we want to change regions?

jet-tong (approved):

PR #835, opened 2026-06-18
Install driver in dedicated namespace for ROSA CI

jet-tong:
## Changes

OpenShift now [blocks ServiceAccount creation in kube-system via an admission webhook](https://github.com/openshift/managed-cluster-validating-webhooks/commit/e270e2b5d61009ff595c2443732c126fdd04fa2a) which [broke our ROSA CI](https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/27537579591/job/81399724260?pr=832). 

- Use the driver in s3-csi-driver namespace on ROSA CI
- Create the namespace and grant privileged SCC to the driver service account before helm install
- Updated csi_driver_irsa trust policy to use s3-csi-driver namespace (hardcode only though).
- Added docs for guidance for ROSA users (and a warning to install into namespace restricted cluster admins only)

## Testing

I used awslabs branch for this PR to invoke trusted workflow - tests are passing. 

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -59,6 +59,9 @@ You may deploy the Mountpoint for Amazon S3 CSI Driver via [Amazon EKS managed a
 > [!IMPORTANT]
 > When upgrading the CSI Driver, existing [Mountpoint Pods](ARCHITECTURE.md#the-mounter-component--mountpoint-pod-aws-s3-csi-mounter) remain unchanged to avoid disrupting your active workloads. New Mountpoint Pods with the updated version are created only when workload pods are started or restarted.
 
+> [!NOTE]
+> If you install the driver into any namespace other than `kube-system`, you must restrict that namespace to cluster administrators only. Do not grant untrusted users access to the driver namespace.
+
 #### Amazon EKS managed add-on
 
 See [the Amazon EKS guide](https://docs.aws.amazon.com/eks/latest/userguide/s3-csi.html) for more details on installing the Amazon EKS managed add-on of Mountpoint for Amazon S3 CSI Driver.
@@ -101,6 +104,22 @@ We recommend customers to separate the controller components from the nodes runn
 helm upgrade --install aws-mountpoint-s3-csi-driver \
     --namespace kube-system \
     --set controller.nodeSelector."kubernetes\.io/role"="control-plane" \
+    aws-mountpoint-s3-csi-driver/aws-mountpoint-s3-csi-driver
+```
+
+#### OpenShift / ROSA
+
+On OpenShift and ROSA ([Red Hat OpenShift Service on AWS](https://aws.amazon.com/rosa/)) clusters, the driver must be installed in a dedicated namespace rather than `kube-system`. This is because OpenShift's [managed-cluster-validating-webhooks](https://github.com/openshift/managed-cluster-validating-webhooks) block ServiceAccount creation in managed namespaces.
+
+> [!IMPORTANT]
+> The driver namespace must be restricted to cluster admins only. Do not grant untrusted users access to the driver namespace.
+
+```sh
+oc create namespace s3-csi-driver
+oc adm policy add-scc-to-user privileged -z s3-csi-driver-sa -n s3-csi-driver
+
+helm upgrade --install aws-mountpoint-s3-csi-driver \
+    --namespace s3-csi-driver \
     aws-mountpoint-s3-csi-driver/aws-mountpoint-s3-csi-driver
 ```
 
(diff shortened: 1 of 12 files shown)


tests/e2e-kubernetes/scripts/helm.sh
```diff
@@ -25,14 +25,14 @@ function helm_uninstall_driver() {
     if [[ "${CLUSTER_TYPE}" == "openshift" ]]; then
       echo "OpenShift cluster detected - using graceful Helm uninstall as ClusterRoleBindings and ServiceAccounts cannot be deleted due to admission webhooks."
```
yerzhan7: This no longer needed because now you have permissions to delete Service Account?

tests/e2e-kubernetes/testsuites/util.go
```diff
@@ -37,10 +37,16 @@ type jsonMap = map[string]any
...
-const (
-	csiDriverDaemonSetName      = "s3-csi-node"
-	csiDriverDaemonSetNamespace = "kube-system"
-)
+const csiDriverDaemonSetName = "s3-csi-node"
+
+var csiDriverDaemonSetNamespace = getDriverNamespace()
+
+func getDriverNamespace() string {
+	if namespace := os.Getenv("DRIVER_NAMESPACE"); namespace != "" {
```
yerzhan7: I think it would be cleaner to introduce driver namespace via CLI arg to the e2e test?

E.g. `ginkgo ... --csi-driver-namespace s3-csi-driver` (with kube-system as default value)

tests/e2e-kubernetes/scripts/helm.sh
jet-tong: Good catch, thanks!

tests/e2e-kubernetes/scripts/helm.sh
```diff
@@ -65,7 +56,15 @@ function helm_install_driver() {
     IRSA_FLAG=""
   fi
 
-  $HELM_BIN upgrade --install $RELEASE_NAME --namespace kube-system ./charts/aws-mountpoint-s3-csi-driver --values \
+  # OpenShift blocks SA creation in kube-system via admission webhook.
+  # So we install into a dedicated namespace and grant privileged SCC.
+  # https://github.com/openshift/managed-cluster-validating-webhooks/commit/e270e2b5d61009ff595c2443732c126fdd04fa2a
+  if [[ "${CLUSTER_TYPE}" == "openshift" ]]; then
+    $KUBECTL_BIN create namespace $DRIVER_NAMESPACE --kubeconfig $KUBECONFIG 2>/dev/null || true
+    oc adm policy add-scc-to-user privileged -z s3-csi-driver-sa -n $DRIVER_NAMESPACE
```
yerzhan7, line 64: np: `oc adm policy add-scc-to-user privileged -z s3-csi-driver-sa -n $DRIVER_NAMESPACE` is this idempotent also like previous line?

yerzhan7 (dismissed):

tests/e2e-kubernetes/scripts/run.sh
yerzhan7: Should also add `--driver-namespace=${DRIVER_NAMESPACE}` here?

tests/e2e-kubernetes/scripts/helm.sh
jet-tong: Good call to check, thanks. Workflow already ran multiple times so it should be idempotent. Couldn't find docs about idempotency in https://docs.redhat.com/en/documentation/openshift_container_platform; but with some AI help found [this line](https://github.com/openshift/oc/blob/c639a3143e8d86763c185a5afc6c9b38d24c00a7/pkg/cli/admin/policy/modify_scc.go#L246) that shows idempotency for AddSCC since it tries to Get it first before adding.

docs/INSTALL.md
```diff
@@ -104,6 +107,22 @@ helm upgrade --install aws-mountpoint-s3-csi-driver \
     aws-mountpoint-s3-csi-driver/aws-mountpoint-s3-csi-driver
 ```
 
+#### OpenShift / ROSA
+
+On OpenShift and ROSA ([Red Hat OpenShift Service on AWS](https://aws.amazon.com/rosa/)) clusters, the driver must be installed in a dedicated namespace rather than `kube-system`. This is because OpenShift's [managed-cluster-validating-webhooks](https://github.com/openshift/managed-cluster-validating-webhooks) block ServiceAccount creation in managed namespaces.
```
yerzhan7, line 112: Are you sure that this for general OpenShift clusters? I thought that the issue is only ROSA clusters (i.e. only "managed" clusters) as per code name of the issue [ROSAENG-963](github.com/openshift/managed-cluster-validating-webhooks/pull/543).

Also, I see that that change is reverted now: https://github.com/openshift/managed-cluster-validating-webhooks/commit/e1a9e37dd309de7dbbc73ab473ef53dd2beac83d

jet-tong: Sorry these were big misses by me.

Will try to rerun https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/27537579591/job/82708122488?pr=832 to check if the change is reverted now.

jet-tong:
Closing PR since we OpenShift has reverted the change, and we no longer need this PR.

PR #842, opened 2026-06-26
Keep S3 Pod Attachments whilst pods are running and terminating

muddyfish:
*Issue #, if available:* N/A

*Description of changes:*

When a workload pod is terminating after previously running, keep S3 Pod Attachments associated until the pod terminates.

Ran a manual test and saw that with a grace period, the S3PA stays around until force deletion, where it lasts another ~2 minutes until the StaleAttachmentCleaner kicks in

### DIFF
--- a/cmd/aws-s3-csi-controller/csicontroller/reconciler.go
+++ b/cmd/aws-s3-csi-controller/csicontroller/reconciler.go
@@ -298,6 +298,10 @@ type workloadVolume struct {
 // the Mountpoint Pod will be scheduled for termination as well. This is because if `workloadPod` never transition into its `Running` state,
 // the Mountpoint Pod might never got a successful mount operation, and thus it might never get unmount operation to cleanly exit
 // and might hang there until it reaches its timeout. We just terminate it in this case to prevent unnecessary waits.
+//
+// If `workloadPod` is `Running` and scheduled for termination (i.e., `DeletionTimestamp` is non-nil during graceful shutdown),
+// it is still treated as active. The pod may still be accessing the S3 volume during its termination grace period
+// (e.g., if it ignores SIGTERM), so the Mountpoint Pod and S3PodAttachment must remain until the pod actually exits.
 func (r *Reconciler) spawnOrDeleteMountpointPodIfNeeded(
 	ctx context.Context,
 	workloadPod *corev1.Pod,
@@ -1108,12 +1112,11 @@ func extractCSISpecFromPV(pv *corev1.PersistentVolume) *corev1.CSIPersistentVolu
 	return csi
 }
 
-// isPodActive returns whether given the Pod is active and not in the process of termination.
-// Copied from https://github.com/kubernetes/kubernetes/blob/8770bd58d04555303a3a15b30c245a58723d0f4a/pkg/controller/controller_utils.go#L1009-L1013.
+// isPodActive returns whether the given Pod is active.
+// A pod is active if it is running (regardless of terminating status, to allow shut-downs
+// which access the volume), or if it is pending and not terminating.
 func isPodActive(p *corev1.Pod) bool {
-	return corev1.PodSucceeded != p.Status.Phase &&
-		corev1.PodFailed != p.Status.Phase &&
-		p.DeletionTimestamp == nil
+	return isPodRunning(p) || (p.Status.Phase == corev1.PodPending && p.DeletionTimestamp == nil)
 }
 
 // isPodTerminating returns whether the given Pod is terminating.
--- a/tests/controller/controller_test.go
+++ b/tests/controller/controller_test.go
@@ -957,6 +957,30 @@ var _ = Describe("Mountpoint Controller", func() {
 			waitForObjectToDisappear(pod.Pod)
 			waitForObjectToDisappear(s3pa)
 		})
+
+		It("should keep S3 Pod Attachment while workload pod is terminating but still running", func() {
+			vol := createVolume()
+			vol.bind()
+
+			gracePeriod := int64(300)
+			pod := createPod(withPVC(vol.pvc), func(p *corev1.Pod) {
+				p.Spec.TerminationGracePeriodSeconds = &gracePeriod
+			})
+			pod.schedule(testNode)
+
+			// Pod gets a Mountpoint Pod and transitions to Running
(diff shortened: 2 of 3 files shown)


tests/controller/controller_test.go
```diff
@@ -957,6 +957,30 @@ var _ = Describe("Mountpoint Controller", func() {
...
+				p.Spec.TerminationGracePeriodSeconds = &gracePeriod
+			})
+			pod.schedule(testNode)
+
+			// Pod gets a Mountpoint Pod and transitions to Running
+			s3pa, _ := waitAndVerifyS3PodAttachmentAndMountpointPod(testNode, vol, pod)
+			pod.running()
+
+			// Pod is deleted but ignores SIGTERM — still Running with DeletionTimestamp set
+			pod.terminate()
```
yerzhan7, line 976: np: Not sure I understand the mechanism behind `"ignores SIGTERM"` in this test. Or are you describing a real-world scenario?

cmd/aws-s3-csi-controller/csicontroller/reconciler.go
```diff
@@ -1108,12 +1112,11 @@ func extractCSISpecFromPV(pv *corev1.PersistentVolume) *corev1.CSIPersistentVolu
...
-// isPodActive returns whether given the Pod is active and not in the process of termination.
-// Copied from https://github.com/kubernetes/kubernetes/blob/8770bd58d04555303a3a15b30c245a58723d0f4a/pkg/controller/controller_utils.go#L1009-L1013.
+// isPodActive returns whether the given Pod is active.
+// A pod is active if it is running (regardless of terminating status, to allow shut-downs
+// which access the volume), or if it is pending and not terminating.
 func isPodActive(p *corev1.Pod) bool {
-	return corev1.PodSucceeded != p.Status.Phase &&
-		corev1.PodFailed != p.Status.Phase &&
-		p.DeletionTimestamp == nil
+	return isPodRunning(p) || (p.Status.Phase == corev1.PodPending && p.DeletionTimestamp == nil)
```
yerzhan7, line 1119: np: Here we kinda here switching from denylist to allowlist, and it silently changes behavior for  `Unknown` phase (active->inactive when not terminating).

Something like this would be cleaner/safer:
```
switch p.Status.Phase {
case corev1.PodSucceeded, corev1.PodFailed:
    return false
case corev1.PodRunning:
    return true
default: // Pending/Unknown: active unless terminating
    return !isPodTerminating(p)
}
```

But `Unknown` phase should be [deprecated](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go#L3516), so we should be safe.

jet-tong, line 1119: AI caught this - It seems inverting the logic also swapsies Unknown phase from true to false (or any other unknown phases). But Unknown isn't being set anyways since 2015 anyways (deprecated) so it should be fine to not having an S3PA for it.

References:
- https://github.com/kubernetes/api/blob/e272d17aacf435af22bc8e293a80cf98895cb3d6/core/v1/types.go#L3495-L3518
- https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase

cmd/aws-s3-csi-controller/csicontroller/reconciler.go
```diff
@@ -298,6 +298,10 @@ type workloadVolume struct {
 // the Mountpoint Pod will be scheduled for termination as well. This is because if `workloadPod` never transition into its `Running` state,
 // the Mountpoint Pod might never got a successful mount operation, and thus it might never get unmount operation to cleanly exit
 // and might hang there until it reaches its timeout. We just terminate it in this case to prevent unnecessary waits.
+//
+// If `workloadPod` is `Running` and scheduled for termination (i.e., `DeletionTimestamp` is non-nil during graceful shutdown),
+// it is still treated as active. The pod may still be accessing the S3 volume during its termination grace period
+// (e.g., if it ignores SIGTERM), so the Mountpoint Pod and S3PodAttachment must remain until the pod actually exits.
```
jet-tong, line 304: Might be good to describe problem scenario (I guess it's long terminationGracePeriod + Running + terminating pod + ignores sigterm) in the PR description and highlight `{"Running + terminating", corev1.PodRunning, isDeleting, true},` case in unit test that got swapped from false to true after this change. 

AI generated mermaid diagram that encapsulates this comment + what we chatted about offline, perhaps useful to add to PR description :) (may be missing arrows though)

```mermaid
stateDiagram-v2
    [*] --> Pending: Pod created
    Pending --> Running: Containers started
    Running --> Succeeded: Graceful exit
    Running --> Failed: Crash
    Pending --> PendingTerminating: DeletionTimestamp set
    Running --> RunningTerminating: DeletionTimestamp set
    
    state "isPodActive = true" as active {
        Pending
        Running
        RunningTerminating
    }
    
    state "isPodActive = false" as inactive {
        PendingTerminating
        Succeeded
        Failed
    }

    note right of RunningTerminating
        NEW: Keep S3PA alive
        Pod may still use volume
        during graceful shutdown
    end note

    note right of PendingTerminating
        Remove S3PA immediately
        Pod never ran, wont mount
    end note
```

yerzhan7 (approved):

jet-tong (approved):
LGTM, good fix! Added non-blocking comments only.

tests/controller/controller_test.go
muddyfish: It's just describing a real-world scenario - the grace period specified causes it to stick around. I can remove it if it's not clear

cmd/aws-s3-csi-controller/csicontroller/reconciler.go
muddyfish: I added some unit tests to verify the behaviour remained the same, aside from the unknown case which I checked and it said it shouldn't appear in the real world. 

I was thinking the current check is actually a lot more clear than originally, but can change it if you'd prefer.

tests/controller/controller_test.go
yerzhan7: Got it, but it just was a bit confusing - as if you are saying that the test sends SIGTERM signal. Suggestion:

```
// Deleting with a grace period leaves the pod Running+terminating (no kubelet in envtest to finalize it)
```

PR #845, opened 2026-06-29
Implement credential provisioning for DaemonsetMounter

vladem:
*Issue #, if available:* N/A

*Description of changes:*

- Implement credential provisioning for `DaemonsetMounter` (credentials provisioned into a per-mount subdirectory, e.g. `/comm/59eaf75e-4782-405a-8d04-3edb5a4373d7-s3-e2e-pv`):
   - `pkg/driver/node/credentialprovider/provider_pod.go` uses fixed name for the token file
   - `pkg/driver/node/credentialprovider/provider_driver.go` provides/cleans up token file **per mount**, also uses fixed name for long-term creds file 
- Fix `GetMountId` to use `pv.name` instead of `volumeHandle` (reason: `volumeHandle` is an opaque string, which might contain `/` and other special symbols, so its not safe to use as a path component; `pv.name` is a validated DNS subdomain name and used by kubelet itself as a path component)
- Extract mount & sendOptions into `mountS3AtTarget`
- Reorder function definitions in `DaemonsetMounter`

---

### DIFF
--- a/pkg/driver/node/credentialprovider/provider.go
+++ b/pkg/driver/node/credentialprovider/provider.go
@@ -17,7 +17,6 @@ import (
 	k8sstrings "k8s.io/utils/strings"
 
 	"github.com/awslabs/mountpoint-s3-csi-driver/pkg/driver/node/envprovider"
-	"github.com/awslabs/mountpoint-s3-csi-driver/pkg/util"
 )
 
 // CredentialFilePerm is the default permissions to be used for credential files.
@@ -95,9 +94,6 @@ type ProvideContext struct {
 
 	// MountKind indicates whether the mount is managed by systemd or pod mounter
 	MountKind MountKind
-
-	// FileOwnership specifies UID/GID for credential files. Nil uses process defaults.
-	FileOwnership *util.FileOwnership
 
 	// The following values are provided from CSI volume context.
 	AuthenticationSource     AuthenticationSource
--- a/pkg/driver/node/credentialprovider/provider.go
+++ b/pkg/driver/node/credentialprovider/provider.go
@@ -17,6 +17,7 @@ import (
 	k8sstrings "k8s.io/utils/strings"
 
 	"github.com/awslabs/mountpoint-s3-csi-driver/pkg/driver/node/envprovider"
+	"github.com/awslabs/mountpoint-s3-csi-driver/pkg/util"
 )
 
 // CredentialFilePerm is the default permissions to be used for credential files.
@@ -28,6 +29,11 @@ const CredentialFilePerm = fs.FileMode(0640)
 // It's only readable, listable (execute bit), and writeable by the owner and group.
 // Group access is needed as Mountpoint Pod is run as non-root user
 const CredentialDirPerm = fs.FileMode(0750)
+
+const (
+	webIdentityServiceAccountTokenName    = "token"
+	eksPodIdentityServiceAccountTokenName = "eks-pod-identity-token"
+)
 
 // An AuthenticationSource represents the source (i.e., driver-level or pod-level) where the credentials was obtained.
 type AuthenticationSource = string
@@ -49,6 +55,8 @@ const (
 	MountKindPod MountKind = "pod"
 	// MountKindSystemd indicates the mount is managed by systemd
 	MountKindSystemd MountKind = "systemd"
+	// MountKindDaemonset indicates the mount is managed by DaemonsetMounter
+	MountKindDaemonset MountKind = "daemonset"
 )
 
 // A Provider provides methods for accessing AWS credentials.
@@ -87,6 +95,9 @@ type ProvideContext struct {
 
 	// MountKind indicates whether the mount is managed by systemd or pod mounter
 	MountKind MountKind
+
+	// FileOwnership specifies UID/GID for credential files. Nil uses process defaults.
+	FileOwnership *util.FileOwnership
 
 	// The following values are provided from CSI volume context.
 	AuthenticationSource     AuthenticationSource
@@ -134,6 +145,15 @@ func (ctx *ProvideContext) IsSystemDMountpoint() bool {
 // IsPodMountpoint returns true if this context is managed by pod mounter.
 func (ctx *ProvideContext) IsPodMountpoint() bool {
 	return ctx.MountKind == MountKindPod
+}
+
+func (ctx *ProvideContext) ToCleanupCtx() CleanupContext {
+	return CleanupContext{
+		WritePath: ctx.WritePath,
+		PodID:     ctx.GetCredentialPodID(),
+		VolumeID:  ctx.VolumeID,
+		MountKind: ctx.MountKind,
+	}
 }
 
 // GetCredentialPodID returns the appropriate Pod ID for credential operations.
--- a/pkg/driver/node/credentialprovider/provider.go
+++ b/pkg/driver/node/credentialprovider/provider.go
@@ -92,7 +92,7 @@ type ProvideContext struct {
 	MountpointPodID string
 	VolumeID        string
 
-	// MountKind indicates whether the mount is managed by systemd or pod mounter
+	// MountKind indicates whether the mount is managed by systemd, pod mounter or daemonset mounter
 	MountKind MountKind
 
 	// The following values are provided from CSI volume context.
@@ -143,6 +143,7 @@ func (ctx *ProvideContext) IsPodMountpoint() bool {
 	return ctx.MountKind == MountKindPod
 }
 
+// ToCleanupCtx constructs CleanupContext from ProvideContext.
 func (ctx *ProvideContext) ToCleanupCtx() CleanupContext {
 	return CleanupContext{
 		WritePath: ctx.WritePath,
--- a/pkg/driver/node/credentialprovider/provider_driver.go
+++ b/pkg/driver/node/credentialprovider/provider_driver.go
@@ -107,7 +107,7 @@ func (c *Provider) cleanupFromDriver(cleanupCtx CleanupContext) error {
 func provideStsWebIdentityCredentialsFromDriver(provideCtx ProvideContext) (envprovider.Environment, error) {
 	driverServiceAccountTokenFile := os.Getenv(envprovider.EnvWebIdentityTokenFile)
 	tokenFile := filepath.Join(provideCtx.WritePath, webIdentityServiceAccountTokenName)
-	err := util.ReplaceFile(tokenFile, driverServiceAccountTokenFile, CredentialFilePerm, provideCtx.FileOwnership)
+	err := util.ReplaceFile(tokenFile, driverServiceAccountTokenFile, CredentialFilePerm)
 	if err != nil {
 		return nil, fmt.Errorf("credentialprovider: sts-web-identity: failed to copy driver's service account token: %w", err)
 	}
@@ -122,7 +122,7 @@ func provideStsWebIdentityCredentialsFromDriver(provideCtx ProvideContext) (envp
 // It basically copies driver's injected service account token to [provideCtx.WritePath].
 func provideContainerCredentialsFromDriver(provideCtx ProvideContext, containerAuthorizationTokenFile string, containerCredentialsFullURI string) (envprovider.Environment, error) {
 	tokenFile := filepath.Join(provideCtx.WritePath, eksPodIdentityServiceAccountTokenName)
-	err := util.ReplaceFile(tokenFile, containerAuthorizationTokenFile, CredentialFilePerm, provideCtx.FileOwnership)
+	err := util.ReplaceFile(tokenFile, containerAuthorizationTokenFile, CredentialFilePerm)
 	if err != nil {
 		return nil, fmt.Errorf("credentialprovider: container: failed to copy driver's service account token: %w", err)
 	}
--- a/pkg/driver/node/credentialprovider/provider_driver.go
+++ b/pkg/driver/node/credentialprovider/provider_driver.go
@@ -15,11 +15,6 @@ import (
 	"github.com/awslabs/mountpoint-s3-csi-driver/pkg/util"
 )
 
-const (
-	webIdentityServiceAccountTokenName    = "token"
-	eksPodIdentityServiceAccountTokenName = "eks-pod-identity-token"
-)
-
 // provideFromDriver provides driver-level AWS credentials.
 func (c *Provider) provideFromDriver(provideCtx ProvideContext) (envprovider.Environment, error) {
 	klog.V(4).Infof("credentialprovider: Using driver identity and %s mount kind", provideCtx.MountKind)
@@ -70,7 +65,7 @@ func (c *Provider) provideFromDriver(provideCtx ProvideContext) (envprovider.Env
 	// Container credential provider (EKS Pod Identity)
 	containerAuthorizationTokenFile := os.Getenv(envprovider.EnvContainerAuthorizationTokenFile)
 	containerCredentialsFullURI := os.Getenv(envprovider.EnvContainerCredentialsFullURI)
-	if provideCtx.IsPodMountpoint() && containerAuthorizationTokenFile != "" && containerCredentialsFullURI != "" {
+	if !provideCtx.IsSystemDMountpoint() && containerAuthorizationTokenFile != "" && containerCredentialsFullURI != "" {
 		klog.V(4).Infof("Providing credentials from driver with Container credential provider (EKS Pod Identity)")
 		containerCredsEnv, err := provideContainerCredentialsFromDriver(provideCtx, containerAuthorizationTokenFile, containerCredentialsFullURI)
 		if err != nil {
@@ -85,14 +80,14 @@ func (c *Provider) provideFromDriver(provideCtx ProvideContext) (envprovider.Env
 
 // cleanupFromDriver removes any credential files that were created for driver-level authentication via [Provider.provideFromDriver].
 func (c *Provider) cleanupFromDriver(cleanupCtx CleanupContext) error {
-	prefix := driverLevelLongTermCredentialsProfilePrefix(cleanupCtx.PodID, cleanupCtx.VolumeID)
+	prefix := driverLevelLongTermCredentialsProfilePrefix(cleanupCtx.PodID, cleanupCtx.VolumeID, cleanupCtx.MountKind)
 	errLongTerm := awsprofile.Cleanup(awsprofile.Settings{
 		Basepath: cleanupCtx.WritePath,
 		Prefix:   prefix,
 	})
 
 	var errSTS, errEKS error
-	if cleanupCtx.IsPodMountpoint() {
+	if !cleanupCtx.IsSystemDMountpoint() {
 		errSTS = c.cleanupToken(cleanupCtx.WritePath, webIdentityServiceAccountTokenName)
 		if errSTS != nil {
 			errSTS = status.Errorf(codes.Internal, "Failed to cleanup driver-level service account STS token: %v", errSTS)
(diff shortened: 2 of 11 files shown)


pkg/driver/node/credentialprovider/provider.go
rajdchak: update comment here to include daemonset

pkg/driver/node/credentialprovider/provider.go
```diff
@@ -136,6 +143,15 @@ func (ctx *ProvideContext) IsPodMountpoint() bool {
 	return ctx.MountKind == MountKindPod
 }
 
+func (ctx *ProvideContext) ToCleanupCtx() CleanupContext {
```
rajdchak, line 147: nit: add a one liner comment about this method

pkg/driver/node/credentialprovider/provider_driver.go
```diff
@@ -169,6 +164,9 @@ func provideLongTermCredentialsFromDriver(provideCtx ProvideContext, accessKeyID
 
 // driverLevelLongTermCredentialsProfilePrefix generates a prefix for AWS credential profile names
 // when using driver-level authentication. The prefix includes both pod and volume IDs to ensure uniqueness.
```
rajdchak, line 166: need to update the comment to include the daemonset logic

pkg/driver/node/mounter/daemonset_mounter_test.go
```diff
@@ -117,7 +124,14 @@ func setupDM(t *testing.T) *dmTestCtx {
 	}
 
 	t.Setenv("CONTAINER_KUBELET_PATH", kubeletPath)
-	dm := mounter.NewDaemonsetMounter(client, nodeName, mpmounter.NewWithMount(fakeMounter), mountSyscall)
+	mockCtl := gomock.NewController(t)
+	mockCredProvider := mock_credentialprovider.NewMockProviderInterface(mockCtl)
+	mockCredProvider.EXPECT().Provide(gomock.Any(), gomock.Any()).
+		Return(envprovider.Environment{}, credentialprovider.AuthenticationSourceDriver, nil).
```
rajdchak: will we have tests for AuthenticationSourcePod as well?

pkg/driver/node/credentialprovider/provider_pod.go
```diff
@@ -149,6 +149,7 @@ func (c *Provider) findPodServiceAccountRole(ctx context.Context, provideCtx Pro
 	podServiceAccount := provideCtx.ServiceAccountName
 
 	// In PodMounter we get IAM Role ARN from MountpointS3PodAttachment custom resource
+	// TODO(vlaad): In DaemonsetMounter we should get IAM Role ARN from in-memory cache (or from volume's .meta)
```
rajdchak, line 152: do we plan to do this after the pod sharing changes?

pkg/driver/node/mounter/daemonset_mounter_test.go
rajdchak: sorry I later realised that we are disabling AuthenticationSourcePod in V3

vladem: yeah, currently we rely on integration tests for this (`s3CSICredentialsTestSuite`), I'll look into adding a simple unit test

vladem: > we are disabling AuthenticationSourcePod in V3

Oh, no, I don't think we're disabling it. There likely will be a constraint that the same SA must be used for all workloads sharing a PV with `authenticationSource: pod`. But credentials can be provided from workload's SA.

For these tests its somewhat irrelevant (mounter delegates to CredentialProvider all the logic). Provider's tests [cover](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/pkg/driver/node/credentialprovider/provider_test.go#L410) this.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -111,50 +116,134 @@ func NewDaemonsetMounter(clientset kubernetes.Interface, nodeID string, mount *m
...
+
+	err = dm.cleanupCredentials(commDir, mountId, cleanupCtx)
+	if err != nil {
+		return fmt.Errorf("failed to cleanup credentials: %w", err)
 	}
 
-	// Ensure target directory exists (kubelet may not have created it yet)
-	if err := os.MkdirAll(target, targetDirPerm); err != nil {
-		return fmt.Errorf("failed to create target directory %q: %w", target, err)
+	// finally unmount (this order ensures retrying cleanup)
```
jet-tong, line 214: Just to make sure I understand - putting unmount last is so that we allow cleanupCredentials + 'error file removal' to retry via NodeUnpublishVolume calls, because after unmount succeeds the next NodeUnpublishVolume would call `mounted, err := ns.Mounter.IsMountPoint(targetContainer)` and find it's already unmounted and it doesn't retry?

(Noted [pod_unmounter.go](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/14dff6293ecc872bed712d173c0c0acf784fb588/pkg/driver/node/mounter/pod_unmounter.go#L144) has reverse logic)

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -111,50 +116,134 @@ func NewDaemonsetMounter(clientset kubernetes.Interface, nodeID string, mount *m
...
-	mountId := GetMountId(podUID, volumeId)
+	mountSuccess := false
+
+	commDir, err := dm.GetCommDir()
+	if err != nil {
+		return fmt.Errorf("failed to find s3-csi-daemonset-mounter pod: %w. %s", err, helpMessageForCheckingMounterPodStatus())
+	}
+
+	// Provision credentials before the isMounted early return so republish refreshes them
+	credsEnv, err := dm.provideCredentials(ctx, commDir, mountId, &credentialCtx)
```
jet-tong: Potential problem (probably a design problem from me in previous primary daemonset PR):
GetCommDir() above only returns the cached commdir (rename as GetCachedCommDir?), and if s3-csi-daemonset-mounter restarted in between NodePublishVolume calls and the 5s interval commDir watcher hasn't updated yet (possible, since kubelet retries with [exponential backoff from 500ms, 1s, 2s, till 122s cap](https://github.com/kubernetes/kubernetes/blob/cbe5e3096ec29950198ed0b77ed9c611f3183c14/pkg/util/goroutinemap/exponentialbackoff/exponential_backoff.go#L27-L37)), we would write credentials in a stale file and they would exist there before defer function is called (e.g. 15s waiting for socket timeout).

It was okay before this PR because we don't interact with commDir until mountoptions.Send, which is the line that detects staleness, so we can let the StartCommDirWatch -> checkCommDir 5s cycle to fully handle the commDir staleness issue. 

Potential mitigations: 
1. we could Stat the socket path right after GetCommDir so we're sure it's not stale (recommended, probably will work) - perhaps we can write a function to handle the duplicate stat part in checkCommDir). 
2. Trigger rediscoverCh to checkCommDir which also stats the socket path; not recommended since if it's stale, and if 1000 NodePublishVolume calls happen within the same second, there'd be 1000 calls to tryDiscoverCommDir calling k8s API to list pods.

pkg/driver/node/mounter/daemonset_mounter.go
vladem: correct, this is defence in depth though, the primary mechanism should be the periodic [cleanup](https://github.com/vladem/mountpoint-s3-csi-driver/blob/daemonset-credentials/pkg/driver/node/mounter/daemonset_mounter.go#L185) which will also retry deleting the token left from mount failure; the invariant which we need to maintain is that no mount is started before the token/cache from previous mount with same `mountId` was cleaned up (currently it doesn't hold, it requires tracking mounts and their tokens/cache via `.meta` files)

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -163,49 +252,32 @@ func (dm *DaemonsetMounter) Mount(ctx context.Context, bucketName string, target
...
 	defer func() {
-		if !fdClosed {
-			dm.closeFUSEDevFD(fd)
-		}
-		if unmount {
-			if umErr := dm.mount.Unmount(target); umErr != nil {
-				klog.Errorf("Failed to unmount %q during cleanup: %v", target, umErr)
-			}
+		if err := mpmounter.CloseFD(fd); err != nil {
+			klog.V(4).Infof("DaemonsetMounter: failed to close /dev/fuse fd %d: %v", fd, err)
```
rajdchak: shouldn't this be an error log?

rajdchak: what are the consequences of fd leak?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -111,50 +116,134 @@ func NewDaemonsetMounter(clientset kubernetes.Interface, nodeID string, mount *m
...
+	credsEnv, err := dm.provideCredentials(ctx, commDir, mountId, &credentialCtx)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if mountSuccess {
+			return
+		}
+		if err := dm.cleanupCredentials(commDir, mountId, credentialCtx.ToCleanupCtx()); err != nil {
+			klog.Errorf("DaemonsetMounter: failed to clean up credential directory for mount %s: %v", mountId, err)
```
rajdchak, line 150: is there any security issue if creds fail to get cleaned up?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -224,52 +296,115 @@ func (dm *DaemonsetMounter) Mount(ctx context.Context, bucketName string, target
...
+//
+//	.../pods/<pod-uid>/volumes/kubernetes.io~csi/<pv-name>/mount
+//
+// Both components are escaped as defense in depth against unexpected characters.
+func GetMountId(target string) (string, error) {
+	tp, err := targetpath.Parse(target)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse target path %q: %w", target, err)
+	}
+	return k8sstrings.EscapeQualifiedName(tp.PodID) + "-" + k8sstrings.EscapeQualifiedName(tp.VolumeID), nil
```
rajdchak, line 391: we won't need pod id in the name after the pod sharing changes, am I right?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -111,50 +116,134 @@ func NewDaemonsetMounter(clientset kubernetes.Interface, nodeID string, mount *m
...
 
+	// Ensure target directory exists (kubelet may not have created it yet)
+	if err := os.MkdirAll(target, targetDirPerm); err != nil {
+		return fmt.Errorf("failed to create target directory %q: %w", target, err)
+	}
+
+	// Remove old error file if exists
+	errFilePath := filepath.Join(commDir, GetErrorFileName(mountId))
+	if err = removeIfExists(errFilePath); err != nil {
+		return err
```
jet-tong: np: Perhaps we can add fmt.Errorf to say "failed to remove old error file"?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -224,52 +296,115 @@ func (dm *DaemonsetMounter) Mount(ctx context.Context, bucketName string, target
 			default:
 			}
 		}
+		dm.mount.Unmount(target)
```
jet-tong: Noticed this is new and the error also isn't handled. 

Perhaps in Mount, we can simply move the defer func above mountS3AtTarget so the defer function can do the cleanup in case of any errors in mountoptions.Send?
```go
	// Perform FUSE mount and send options to secondary daemonset
	if err := dm.mountS3AtTarget(ctx, target, bucketName, args, mountId, volumeId, commDir, userEnv, credsEnv); err != nil {
		return err
	}

	// Clean up the FUSE mount if waitForMount fails
	defer func() {
		if !mountSuccess {
			// TODO(vlaad): might leak a mount, implement periodic stale mount cleanup
			if umErr := dm.mount.Unmount(target); umErr != nil {
				klog.Errorf("Failed to unmount %q during cleanup: %v", target, umErr)
			}
		}
	}()
```

pkg/driver/node/mounter/daemonset_mounter.go
jet-tong: np: we can update to add provideCredentials (and perhaps mount + send can be combined to mountS3AtTarget)

Perhaps this part is duplicate documentation to comments above Mount() and I shouldn't have added it...?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -111,50 +116,134 @@ func NewDaemonsetMounter(clientset kubernetes.Interface, nodeID string, mount *m
...
+	if err != nil {
+		return fmt.Errorf("failed to find s3-csi-daemonset-mounter pod: %w. %s", err, helpMessageForCheckingMounterPodStatus())
+	}
+
+	// Provision credentials before the isMounted early return so republish refreshes them
+	credsEnv, err := dm.provideCredentials(ctx, commDir, mountId, &credentialCtx)
+	if err != nil {
+		return err
+	}
+	defer func() {
```
jet-tong, line 145: np: Would it be better to put the cleanup defer function before dm.provideCredentials to clean up in case of any errors in provideCredentials?

pkg/driver/node/credentialprovider/provider_pod.go
jet-tong: np: (AI noted) this will need to change to !provideCtx.IsSystemDMountpoint() (like other changes in provider_driver.go), but we'll do pod sharing in a future PR anyways.

pkg/driver/node/mounter/daemonset_mounter.go
vladem: yes, name of the persistent volume may be used to identify a mount on a given node in future

pkg/driver/node/mounter/daemonset_mounter_test.go
```diff
@@ -274,12 +297,53 @@ func TestDaemonsetMounter(t *testing.T) {
...
+			}
+
+			err := testCtx.dm.Mount(testCtx.ctx, testCtx.bucketName, target, credentialprovider.ProvideContext{
+				WorkloadPodID: testCtx.podUID,
+				VolumeID:      testCtx.volumeID,
+			}, mountpoint.ParseArgs(nil), "", nil)
+			if err == nil {
+				t.Fatal("mount should fail")
+			}
+			assert.Contains(t, err.Error(), "simulated mount failure")
```
jet-tong, line 354: Perhaps can check credentials directory doesn't exist also to test `os.RemoveAll(mountCredDir)`?

e.g.: `assert.ErrorIs(t, os.Stat(expectedWritePath), os.ErrNotExist)`

pkg/driver/node/mounter/daemonset_mounter_test.go
```diff
@@ -274,12 +297,53 @@ func TestDaemonsetMounter(t *testing.T) {
 			// verify Unmount was called via FakeMounter log.
 			testCtx.assertUnmounted(target)
 		})
+
+		t.Run("Credentials cleaned up on mount failure", func(t *testing.T) {
```
jet-tong, line 316: I was thinking adding unit tests for what happens to credentials during Unmount + republish would be super useful to cover the rest of the PR code.

If you think these aren't overkill to add perhaps they might be useful?
- Modify Unmount test "Removes mount from target" to "Removes mount from target and cleans up credentials" and include the `mockCredProvider.EXPECT().Cleanup` block
- Modify "Does not duplicate mounts if target is already mounted" to "Does not duplicate mounts if target is already mounted and refreshes credentials" to check for republish refreshing credentials (since in e2e credentials.go it's only a TODO for now, or perhaps just implement the republish e2e test for more comprehensive testing?)
- Do we also want a happy path credentials unit test? (They should be covered in e2e tests so probably not)

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -111,50 +116,134 @@ func NewDaemonsetMounter(clientset kubernetes.Interface, nodeID string, mount *m
...
-	// Idempotency: if target is already a healthy Mountpoint mount, return early.
-	// Kubelet may call NodePublishVolume repeatedly (requiresRepublish, retries).
+	// Idempotency: if target is already a healthy Mountpoint mount, return early
 	isMounted, err := dm.IsMountPoint(target)
-	// If target doesn't exist (ErrNotExist), defer directory creation to MkdirAll step.
 	if err != nil && !errors.Is(err, fs.ErrNotExist) {
-		// Note: Skipped handling dm.mount.IsMountpointCorrupted(err) case - we accept that for old
-		// pods they won't recover after daemonset-mounter restart, unlike PodMounter:verifyOrSetupMountTarget
-		// TODO: daemonset mounter pod recovery
-		if dm.mount.IsMountpointCorrupted(err) {
```
jet-tong, line 135: Do we need to do anything to check and handle IsMountpointCorrupted elsewhere?

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak, line 69: qq: why do we fetch this from an env var? why can't we have a constant value for this?

jet-tong: s3-csi-daemonset-mounter is in .Release.Namespace same as s3-csi-node (node.yaml), and user could be installing it to any namespaces not just default kube-system, so can't use constant value.

I added env var so s3-csi-node knows which namespace to find s3-csi-daemonset-mounter pod (in [this commit](https://github.com/awslabs/mountpoint-s3-csi-driver/pull/797/changes/dafa52cea92f1ad5b9649aaa851cf943cdf6bd71)). (Looking back perhaps s3-csi-node can just check which namespace it is in and query that namespace, not sure if it'll be cleaner code though.

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: I agree with this

rajdchak: I dont think it is a nitpick

pkg/driver/node/mounter/daemonset_mounter.go
vladem: Decided to accept this behaviour. We can not guarantee that the `commDir` is not stale at the moment when we write credentials (the mounter pod might be gone a moment before). We accept that in this case we will try to write to stale `commDir`, which will typically fail because directory doesn't exist anymore. If the write succeeds, we rely on the cleanup mechanism to ensure no token file is lost: cleanup will run either at the end of `NodePublishVolume` call or at the next cycle of the periodic cleanup.

pkg/driver/node/mounter/daemonset_mounter.go
vladem: On this error, typically FD is closed but something went wrong while closing (e.g. FD already was closed / never opened). Possible consequence is that our process can run out of FD-s and all future mounts will fail. I think its a low risk, given the behaviour of `close(devFuseFd)`.

pkg/driver/node/mounter/daemonset_mounter.go
vladem: Yes, decided to have a periodic cleanup goroutine. We'll track all token files as attributes of mounts (in RAM and on disk). Addressing as a follow-up.

pkg/driver/node/mounter/daemonset_mounter.go
vladem: correct, we'd like `MOUNTER_NAMESPACE` to be configurable, we might choose to make it different to `csi-node`'s namespace

pkg/driver/node/mounter/daemonset_mounter.go
vladem: We can not recover a corrupted target mount, so removed for now. We do need to detect and recover a corrupted source mount, that should be handled with mount sharing PR.

pkg/driver/node/credentialprovider/provider.go
jet-tong, line 173: np: now can use "MountKind indicates whether the mount is managed by systemd, pod mounter or daemonset mounter" comment as in line 95 above.

pkg/driver/node/mounter/daemonset_mounter.go
jet-tong: Gotcha. Saw IsSourceHealthy in #862 that should handle this
FYI @rajdchak

rajdchak (approved):

jet-tong (approved):

pkg/driver/node/credentialprovider/provider.go
vladem: will update the comment in the follow up PR

PR #851, opened 2026-06-30
Add MaxVolumesPerNode to enable scheduler volume limit enforcement

jet-tong:
### Description

This PR advertises `MaxVolumesPerNode` ([KEP 554](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/554-maximum-volume-count)) via `NodeGetInfo` in daemonset mode, and the scheduler's existing [NodeVolumeLimit plugin](https://github.com/kubernetes/kubernetes/blob/eb51fbf7c6de0e361504729feabdadd178cd2b7b/pkg/scheduler/framework/plugins/nodevolumelimits/csi.go#L257) reads this and rejects workload pods when the node's volume capacity is full, which allows autoscaler scale-ups. 

#### Changes

Implementation:
- values.yaml: maxVolumesPerNode field and sizing guidance
- node.yaml: Emit env var + Helm validation (fail if < 0), 
  - Allow 0 for unbounded volume limits, with warning in node logs / values.yaml comment. Note kubelet [doesn't write allocatable value if 0](https://github.com/kubernetes/kubernetes/blob/6a5cc912a3c2a2c241fe05abea563079bf574abd/pkg/volume/csi/nodeinfomanager/nodeinfomanager.go#L634-L643)
- driver.go: Parse MAX_VOLUMES_PER_NODE env var via util.GetEnvAsInt (referenced from k8s [here](https://github.com/kubernetes/kubernetes/blob/release-1.36/pkg/util/env/env.go))
- node.go: Add MaxVolumesPerNode field to S3NodeServer, returns in NodeGetInfo

Tests:
- Unit test: Very basic passthrough verification and parsing logic test adapted from k8s [env_test.go](https://github.com/kubernetes/kubernetes/blob/v1.36.2/pkg/util/env/env_test.go)
- E2E test: Adapted from [upstream volumelimits.go](https://github.com/kubernetes/kubernetes/blob/v1.36.2/test/e2e/storage/testsuites/volumelimits.go) for our existing static provisioning pattern
  - "should support volume limits" - fill node to capacity, extra pod should be rejected
  - "should admit a pod reusing an already-mounted PV without counting it twice" - fills node to capacity, node that uses existing PV should be accepted. (volume sharing case)
  - "should verify that all csinodes have volume limits" - deployment health check

#### Important follow-up PR after this PR merges
- Prevent / Warn users when they have non-unique volumeHandles (map based on PV name in pod sharing PR doesn't encapsulate volumeHandles, but we can reference some of the restart guarantees in the PR to write similar map including volumeHandles to repopulate on restart)

#### Important Notes
- For Karpenter integration - this startup taint might be needed for race condition where pods are scheduled into node before the driver starts advertising the maxVolumesPerNode volume limit.
```yaml
      startupTaints:
        - key: s3.csi.aws.com/agent-not-ready
          effect: NoSchedule
```
- Upgrade note: switching an existing node from pod mode to daemonset mode updates the advertised limit correctly - [kubelet overwrites CSINode.allocatable.count](https://github.com/kubernetes/kubernetes/blob/bbedc834eb4e2509cb913a5e0dcf5236fab6fac9/pkg/volume/csi/nodeinfomanager/nodeinfomanager.go#L619) on driver re-registration, when performing helm upgrade. Verified in this test:
  - Deployed in pod mode, and added 10 s3 workloads
  - helm upgrade to daemonset mode (volume limit 4), and added 1 s3 workload
  - All pod mode workload pods stay working, and the new s3 workload in daemonset mode Pending state with reason "FailedScheduling".

---

#### Motivation

In current pod mode, per-mount resources are per-PV volumeAttributes (`mountpointContainerResourcesRequestsMemory`, `cacheEphemeralStorageResourceRequest`) and the scheduler sees each Mountpoint pod individually. In daemonset architecture, all Mountpoint processes share a daemonset pod with fixed resource budget. Without a volume limit (per node), the scheduler, with no visibility into per-mount resource usage, would keep placing workloads until the mount resources are exhausted (e.g. OOM when out of RAM, or fills disk with emptyDir cache). We decided to use `MaxVolumesPerNode` to mitigate this problem.

Alternatives considered:
- Driver computes via memory per mount - but it forces all mounts to use identical resources, and is in conflict with PV --memory-target for example.
- [DRA consumable capacity](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/#consumable-capacity) - but `ResourceClaims` are pod-scoped so pod-sharing mounts are double-counted, and requires mutating webhook.
- Upstream KEP for multi-dimensional volume limits - extend `CSINode.Allocatable` from a flat count to resource budgets, and add per-PV cost to `PersistentVolumeSpec` so scheduler sums costs instead of counting. Significant upstream changes required (CSI spec, k8s API, scheduler changes) and is something we could still consider if configuring `MaxVolumesPerNode` is a big pain point for customers.

#### `maxVolumesPerNode` flow

Data path of maxVolumesPerNode flow (with AI help finding the path)
- Helm value (`maxVolumesPerNode: 10`)
- -> [node.yaml template](./charts/aws-mountpoint-s3-csi-driver/templates/node.yaml) (env `MAX_VOLUMES_PER_NODE=10`)
- -> [driver.go](./pkg/driver/driver.go) parses `maxVolumesPerNode`
- -> [node.go NodeGetInfo](./pkg/driver/node/node.go) ([docs](https://github.com/container-storage-interface/spec/blob/master/spec.md#nodegetinfo)) returns `MaxVolumesPerNode: ns.MaxVolumesPerNode`
- -> [kubelet calls NodeGetInfo](https://github.com/kubernetes/kubernetes/blob/6a5cc912a3c2a2c241fe05abea563079bf574abd/pkg/volume/csi/csi_plugin.go#L149) and gets `maxVolumePerNode`
- -> [kubelet writes `CSINode.allocatable.count`](https://github.com/kubernetes/kubernetes/blob/6a5cc912a3c2a2c241fe05abea563079bf574abd/pkg/volume/csi/csi_plugin.go#L157): through [`InstallCSIDriver`](https://github.com/kubernetes/kubernetes/blob/6a5cc912a3c2a2c241fe05abea563079bf574abd/pkg/volume/csi/nodeinfomanager/nodeinfomanager.go#L112), [`updateCSINode`](https://github.com/kubernetes/kubernetes/blob/6a5cc912a3c2a2c241fe05abea563079bf574abd/pkg/volume/csi/nodeinfomanager/nodeinfomanager.go#L360), [`tryUpdateCSINode`](https://github.com/kubernetes/kubernetes/blob/6a5cc912a3c2a2c241fe05abea563079bf574abd/pkg/volume/csi/nodeinfomanager/nodeinfomanager.go#L385), [`installDriverToCSINode`](https://github.com/kubernetes/kubernetes/blob/6a5cc912a3c2a2c241fe05abea563079bf574abd/pkg/volume/csi/nodeinfomanager/nodeinfomanager.go#L591), and writes allocatable count in [`driverSpec.Allocatable = &storagev1.VolumeNodeResources{Count: &m}`](https://github.com/kubernetes/kubernetes/blob/6a5cc912a3c2a2c241fe05abea563079bf574abd/pkg/volume/csi/nodeinfomanager/nodeinfomanager.go#L639-L640)
- -> [scheduler reads `allocatable.count`](https://github.com/kubernetes/kubernetes/blob/6a5cc912a3c2a2c241fe05abea563079bf574abd/pkg/scheduler/framework/plugins/nodevolumelimits/csi.go#L204)
- -> [scheduler rejects](https://github.com/kubernetes/kubernetes/blob/6a5cc912a3c2a2c241fe05abea563079bf574abd/pkg/scheduler/framework/plugins/nodevolumelimits/csi.go#L348) pods when volume count is over the limit
- Autoscalers' responses
  - -> [Karpenter gets the volume limit in populateVolumeLimits](https://github.com/kubernetes-sigs/karpenter/blob/43964dc54782bc68f15b7da6414c50931dadd07b/pkg/controllers/state/cluster.go#L845-L857), and [Karpenter actions based on `ExceedsLimits`](https://github.com/kubernetes-sigs/karpenter/blob/43964dc54782bc68f15b7da6414c50931dadd07b/pkg/scheduling/volumeusage.go#L201-L208)
  - -> [KEP-5030: Integrate Volume Attach limit into cluster autoscaler](https://github.com/kubernetes/enhancements/tree/master/keps/sig-autoscaling/5030-attach-limit-autoscaler) allows Cluster Autoscaler to be aware of volume limits.

#### Unique volumeHandles needed

For this feature to work, volumeHandles must be unique. This is because for static provisioning, k8s count PVs with their volumeHandles in [nodevolumelimits/csi.go](https://github.com/kubernetes/kubernetes/blob/6a5cc912a3c2a2c241fe05abea563079bf574abd/pkg/scheduler/framework/plugins/nodevolumelimits/csi.go#L499), and uses `fmt.Sprintf("%s/%s", driverName, volumeHandle)
` in [getVolumeUniqueName](https://github.com/kubernetes/kubernetes/blob/6a5cc912a3c2a2c241fe05abea563079bf574abd/pkg/scheduler/framework/plugins/nodevolumelimits/csi.go#L688).

### Misc Notes

- Node pod limit can be queried if we want via instance metadata: [source of formula](https://github.com/awslabs/amazon-eks-ami/blob/main/nodeadm/internal/kubelet/eni_max_pods.go): instanceInfo.DefaultMaxENIs*(instanceInfo.Ipv4AddressesPerInterface-1) + 2

#### Note for Reviewer(s)

Some follow-ups / TODOs:
- Prevent / Warn users when they have non-unique volumeHandles (map based on PV name in pod sharing PR doesn't encapsulate volumeHandles, but we can reference some of the restart guarantees in the PR to write similar map including volumeHandles to repopulate on restart)
- Warn if maxMountsPerNode * --max-memory-target exceeds the pod resources.requests.memory - TODO as a follow up task after memory limit is integrated with daemonset architecture
- Add documentation for volume limits (we might want to do this as part of follow up too since we need to clarify user experience on it)

For Reviewer: Things to consider when reviewing:
- Unit test doesn't test much - do we need driver_test.go?
- Copies from k8s package - I tried to follow existing test suites (e.g. multivolumes.go) porting from k8s testsuite to our static provisioning integration tests, but perhaps we can further simplify the test file. My thinking was to reduce the changes from the original in case we later support dynamic provisioning which would mean we can use the original testsuites adding our new (e.g. pod sharing) tests for volume limits.
  - 1 pod N PVs in integration test, or 10 pod each with 1 PV? (chosen 1 pod N PVs for simplicity / match k8s volumelimits.go testsuite)

### DIFF
--- a/charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
@@ -105,10 +105,16 @@ spec:
             - name: INSTALLATION_TYPE
               value: {{ if .Values.isEKSAddon }}eks-addon{{ else }}helm{{ end }}
             {{- if eq .Values.experimental.mounterMode "daemonset" }}
+            {{- $mounter := index .Values.daemonsetMounters 0 }}
+            {{- if lt (int $mounter.maxVolumesPerNode) 1 }}
+            {{- fail "daemonsetMounters[].maxVolumesPerNode must be >= 1" }}
+            {{- end }}
             - name: MOUNTER_MODE
               value: daemonset
             - name: MOUNTER_NAMESPACE
               value: {{ .Release.Namespace }}
+            - name: MAX_VOLUMES_PER_NODE
+              value: "{{ $mounter.maxVolumesPerNode }}"
             {{- end }}
             {{- with .Values.awsAccessSecret }}
             - name: AWS_ACCESS_KEY_ID
--- a/charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
@@ -105,9 +105,12 @@ spec:
             - name: INSTALLATION_TYPE
               value: {{ if .Values.isEKSAddon }}eks-addon{{ else }}helm{{ end }}
             {{- if eq .Values.experimental.mounterMode "daemonset" }}
+            {{- if ne (len .Values.daemonsetMounters) 1 }}
+            {{- fail "daemonsetMounters must have exactly one entry" }}
+            {{- end }}
             {{- $mounter := index .Values.daemonsetMounters 0 }}
-            {{- if lt (int $mounter.maxVolumesPerNode) 1 }}
-            {{- fail "daemonsetMounters[].maxVolumesPerNode must be >= 1" }}
+            {{- if lt (int $mounter.maxVolumesPerNode) 0 }}
+            {{- fail "daemonsetMounters[].maxVolumesPerNode must be >= 0" }}
             {{- end }}
             - name: MOUNTER_MODE
               value: daemonset
--- a/charts/aws-mountpoint-s3-csi-driver/values.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/values.yaml
@@ -156,12 +156,17 @@ daemonsetMounters:
     # Advertised to the Kubernetes scheduler via MaxVolumesPerNode.
     # Shared volumes (same PV used by multiple pods) are counted once.
   - maxVolumesPerNode: 10
-    # Resource requests and limits for the mounter DaemonSet pod.
-    # requests.memory should be sized to maxMountsPerNode × per-mount memory
-    # (e.g., 10 mounts × 2Gi = 20Gi).
+    # Total resource requests and limits for the mounter DaemonSet pod.
+    # requests.memory should be sized to maxVolumesPerNode × per-mount memory
+    # (e.g., 4 mounts × 512Mi = 2Gi, or 10 mounts × 4Gi = 40Gi).
+    # Please refer to Mountpoint documentation for recommended resource configurations.
+    # TODO: Add link for Mountpoint documentation
+    # requests.cpu should also be sized to maxVolumesPerNode × per-mount CPU
+    # (e.g.  4 mounts × 500Mi = 2000m, or 10 mounts × 500m = 5000m).
+    # TODO: update guidances with memory limiter integration
     resources:
       requests:
-        memory: "1Gi"
+        memory: "2Gi"
         cpu: "500m"
     logLevel: 4
     podLabels: {}
--- a/charts/aws-mountpoint-s3-csi-driver/values.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/values.yaml
@@ -154,16 +154,14 @@ experimental:
 daemonsetMounters:
     # Maximum number of distinct S3 volumes (Mountpoint processes) per node.
     # Advertised to the Kubernetes scheduler via MaxVolumesPerNode.
-    # Shared volumes (same PV used by multiple pods) are counted once.
-  - maxVolumesPerNode: 10
+    # Set to 0 to disable the limit (unbounded); this risks resource exhaustion
+    # (memory OOM, or disk exhaustion with cache) on the mounter DaemonSet pod.
+    # IMPORTANT: volumeHandle must be unique per PV as the scheduler uses it to 
+    # track volume usage.
+  - maxVolumesPerNode: 4
     # Total resource requests and limits for the mounter DaemonSet pod.
     # requests.memory should be sized to maxVolumesPerNode × per-mount memory
     # (e.g., 4 mounts × 512Mi = 2Gi, or 10 mounts × 4Gi = 40Gi).
-    # Please refer to Mountpoint documentation for recommended resource configurations.
-    # TODO: Add link for Mountpoint documentation
-    # requests.cpu should also be sized to maxVolumesPerNode × per-mount CPU
-    # (e.g.  4 mounts × 500Mi = 2000m, or 10 mounts × 500m = 5000m).
-    # TODO: update guidances with memory limiter integration
     resources:
       requests:
         memory: "2Gi"
--- a/charts/aws-mountpoint-s3-csi-driver/values.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/values.yaml
@@ -152,10 +152,17 @@ experimental:
 # Configuration for the secondary mounter daemonset(s).
 # Only used when experimental.mounterMode is "daemonset".
 daemonsetMounters:
-  - resources:
+    # Maximum number of distinct S3 volumes (Mountpoint processes) per node.
+    # Advertised to the Kubernetes scheduler via MaxVolumesPerNode.
+    # Shared volumes (same PV used by multiple pods) are counted once.
+  - maxVolumesPerNode: 10
+    # Resource requests and limits for the mounter DaemonSet pod.
+    # requests.memory should be sized to maxMountsPerNode × per-mount memory
+    # (e.g., 10 mounts × 2Gi = 20Gi).
+    resources:
       requests:
-        cpu: 100m
-        memory: 256Mi
+        memory: "1Gi"
+        cpu: "500m"
     logLevel: 4
     podLabels: {}
     nodeSelector: {}
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -50,6 +50,7 @@ import (
 	"github.com/awslabs/mountpoint-s3-csi-driver/pkg/driver/version"
 	mpmounter "github.com/awslabs/mountpoint-s3-csi-driver/pkg/mountpoint/mounter"
 	"github.com/awslabs/mountpoint-s3-csi-driver/pkg/podmounter/mppod/watcher"
+	"github.com/awslabs/mountpoint-s3-csi-driver/pkg/util"
 )
 
 const (
@@ -61,11 +62,12 @@ const (
 )
 
 var (
-	mountpointPodNamespace = os.Getenv("MOUNTPOINT_NAMESPACE")
-	mounterMode            = os.Getenv("MOUNTER_MODE")
-	mounterModeDaemonset   = "daemonset"
-	podWatcherResyncPeriod = time.Minute
-	scheme                 = runtime.NewScheme()
+	mountpointPodNamespace   = os.Getenv("MOUNTPOINT_NAMESPACE")
+	mounterMode              = os.Getenv("MOUNTER_MODE")
+	mounterModeDaemonset     = "daemonset"
+	maxVolumesPerNodeEnvName = "MAX_VOLUMES_PER_NODE"
+	podWatcherResyncPeriod   = time.Minute
+	scheme                   = runtime.NewScheme()
 )
 
 func init() {
@@ -119,7 +121,7 @@ func NewDriver(endpoint string, mpVersion string, nodeID string) (*Driver, error
 
 	stopCh := make(chan struct{})
 
-	var nodeMounter mounter.Mounter
+	var nodeServer *node.S3NodeServer
 
 	if mounterMode == mounterModeDaemonset {
 		klog.Info("Using daemonset mounter mode")
@@ -139,7 +141,15 @@ func NewDriver(endpoint string, mpVersion string, nodeID string) (*Driver, error
 		}
 		go dm.StartCommDirWatch(stopCh)
 
-		nodeMounter = dm
+		maxVolumesPerNode, err := util.GetEnvAsIntOrFallback(maxVolumesPerNodeEnvName, 0)
+		if err != nil {
+			return nil, err
+		}
+		if maxVolumesPerNode < 1 {
+			return nil, fmt.Errorf("%s must be >= 1, got %d", maxVolumesPerNodeEnvName, maxVolumesPerNode)
+		}
+
+		nodeServer = node.NewS3NodeServer(nodeID, dm, int64(maxVolumesPerNode))
 	} else {
 		mpMounter := mpmounter.New()
 		podWatcher := watcher.New(clientset, mountpointPodNamespace, nodeID, podWatcherResyncPeriod)
@@ -161,10 +171,8 @@ func NewDriver(endpoint string, mpVersion string, nodeID string) (*Driver, error
 		if err != nil {
 			klog.Fatalln(err)
 		}
-		nodeMounter = podMounter
+		nodeServer = node.NewS3NodeServer(nodeID, podMounter, 0) // maxVolumesPerNode = 0 means no limit
 	}
-
-	nodeServer := node.NewS3NodeServer(nodeID, nodeMounter)
 
 	return &Driver{
 		Endpoint:   endpoint,
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -141,7 +141,7 @@ func NewDriver(endpoint string, mpVersion string, nodeID string) (*Driver, error
 		}
 		go dm.StartCommDirWatch(stopCh)
 
-		maxVolumesPerNode, err := util.GetEnvAsIntOrFallback(maxVolumesPerNodeEnvName, 0)
+		maxVolumesPerNode, err := util.GetEnvAsInt(maxVolumesPerNodeEnvName)
 		if err != nil {
 			return nil, err
 		}
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -145,8 +145,14 @@ func NewDriver(endpoint string, mpVersion string, nodeID string) (*Driver, error
 		if err != nil {
 			return nil, err
 		}
-		if maxVolumesPerNode < 1 {
-			return nil, fmt.Errorf("%s must be >= 1, got %d", maxVolumesPerNodeEnvName, maxVolumesPerNode)
+		if maxVolumesPerNode < 0 {
+			return nil, fmt.Errorf("%s must be >= 0, got %d", maxVolumesPerNodeEnvName, maxVolumesPerNode)
+		}
+		if maxVolumesPerNode == 0 {
+			klog.Warningf("%s is 0: volume limit is disabled (unbounded). The scheduler will not "+
+				"limit the number of S3 volumes per node, which may lead to resource exhaustion "+
+				"(memory OOM, or disk exhaustion when cache is enabled) on s3-csi-daemonset-mounter.",
+				maxVolumesPerNodeEnvName)
 		}
 
 		nodeServer = node.NewS3NodeServer(nodeID, dm, int64(maxVolumesPerNode))
--- a/pkg/util/env.go
+++ b/pkg/util/env.go
@@ -1,7 +1,24 @@
 package util
 
-import "os"
+import (
+	"os"
+	"strconv"
+)
 
 func SupportLegacySystemdMounts() bool {
 	return os.Getenv("SUPPORT_LEGACY_SYSTEMD_MOUNTS") == "true"
+}
+
+// GetEnvAsIntOrFallback returns the env variable (parsed as integer) for
+// the given key and falls back to the given defaultValue if not set.
+// Copied from https://github.com/kubernetes/kubernetes/blob/release-1.36/pkg/util/env/env.go
+func GetEnvAsIntOrFallback(key string, defaultValue int) (int, error) {
+	if v := os.Getenv(key); v != "" {
+		value, err := strconv.Atoi(v)
+		if err != nil {
+			return defaultValue, err
+		}
+		return value, nil
+	}
+	return defaultValue, nil
 }
--- a/pkg/util/env.go
+++ b/pkg/util/env.go
@@ -1,6 +1,7 @@
 package util
 
 import (
+	"fmt"
 	"os"
 	"strconv"
 )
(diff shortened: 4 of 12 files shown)


charts/aws-mountpoint-s3-csi-driver/values.yaml
```diff
@@ -146,10 +146,17 @@ experimental:
...
 # Only used when experimental.mounterMode is "daemonset".
 daemonsetMounters:
-  - resources:
+    # Maximum number of distinct S3 volumes (Mountpoint processes) per node.
+    # Advertised to the Kubernetes scheduler via MaxVolumesPerNode.
+    # Shared volumes (same PV used by multiple pods) are counted once.
+  - maxVolumesPerNode: 10
+    # Resource requests and limits for the mounter DaemonSet pod.
+    # requests.memory should be sized to maxMountsPerNode × per-mount memory
+    # (e.g., 10 mounts × 2Gi = 20Gi).
```
rajdchak: should we also define how to size requests.cpu?

pkg/driver/driver.go
```diff
@@ -131,7 +133,15 @@ func NewDriver(endpoint string, mpVersion string, nodeID string) (*Driver, error
 		}
 		go dm.StartCommDirWatch(stopCh)
 
-		nodeMounter = dm
+		maxVolumesPerNode, err := util.GetEnvAsIntOrFallback(maxVolumesPerNodeEnvName, 0)
```
rajdchak: instead of defaulting to 0 here, can we check whether the env var is set and fail in case not set? This will be a more specific msg " environment variable is required" rather than failing later with a msg "must be >=1 got 0"
```
val, ok := os.LookupEnv(maxVolumesPerNodeEnvName)
if !ok {
    return nil, fmt.Errorf("%s environment variable is required", maxVolumesPerNodeEnvName)
}
```

charts/aws-mountpoint-s3-csi-driver/values.yaml
```diff
@@ -146,10 +146,17 @@ experimental:
 # Configuration for the secondary mounter daemonset(s).
 # Only used when experimental.mounterMode is "daemonset".
 daemonsetMounters:
-  - resources:
+    # Maximum number of distinct S3 volumes (Mountpoint processes) per node.
+    # Advertised to the Kubernetes scheduler via MaxVolumesPerNode.
+    # Shared volumes (same PV used by multiple pods) are counted once.
+  - maxVolumesPerNode: 10
```
rajdchak: how did we come up with this value?

pkg/util/env.go
```diff
@@ -1,7 +1,24 @@
...
+)
 
 func SupportLegacySystemdMounts() bool {
 	return os.Getenv("SUPPORT_LEGACY_SYSTEMD_MOUNTS") == "true"
 }
+
+// GetEnvAsIntOrFallback returns the env variable (parsed as integer) for
+// the given key and falls back to the given defaultValue if not set.
+// Copied from https://github.com/kubernetes/kubernetes/blob/release-1.36/pkg/util/env/env.go
+func GetEnvAsIntOrFallback(key string, defaultValue int) (int, error) {
```
rajdchak: in case we have the logic for checking whether the env var is set or not, I don't think we will need a default value.

jet-tong: Thanks, good point - I copied GetEnvAsIntOrFallback from k8s so it could be used for pod mode code as well (which needs 0) but decided to split them in the if else block for more clarity, so we no longer need the fallback. I think I'll just remove the fallback logic there and raise errors instead.

charts/aws-mountpoint-s3-csi-driver/values.yaml
jet-tong: Good call, will check out how to guide users on requests.cpu field.

charts/aws-mountpoint-s3-csi-driver/values.yaml
jet-tong: Will need to look further, but based on [this](https://github.com/awslabs/mountpoint-s3/blob/fc04a2533e8088a48b029903239f09e41e569645/mountpoint-s3/src/cli.rs#L192) hard limit we might want to reduce this number to 4 (512MB * 4 = 2 GB).

jet-tong: It's best we update the recommendations after memory limit integration, so have added TODOs there for now.

charts/aws-mountpoint-s3-csi-driver/values.yaml
jet-tong: Added placeholder guidance there for now mirroring memory ones, we should update the guidance after more testing with memory limiter feature.

charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
```diff
@@ -105,10 +105,16 @@ spec:
             - name: INSTALLATION_TYPE
               value: {{ if .Values.isEKSAddon }}eks-addon{{ else }}helm{{ end }}
             {{- if eq .Values.experimental.mounterMode "daemonset" }}
+            {{- $mounter := index .Values.daemonsetMounters 0 }}
```
vladem, line 111: do we check somewhere that `.Values.daemonsetMounters` has exactly 1 element? how do we fail if 0 or 2 are provided?

charts/aws-mountpoint-s3-csi-driver/values.yaml
```diff
@@ -152,10 +152,22 @@ experimental:
 # Configuration for the secondary mounter daemonset(s).
 # Only used when experimental.mounterMode is "daemonset".
 daemonsetMounters:
-  - resources:
+    # Maximum number of distinct S3 volumes (Mountpoint processes) per node.
+    # Advertised to the Kubernetes scheduler via MaxVolumesPerNode.
+    # Shared volumes (same PV used by multiple pods) are counted once.
```
vladem: "sharing" or deduplication in k8s scheduler is done based on `volumeHandle` property of the PV, consider adding a check to `NodePublishVolume` whether a given `volumeHandle` is already in use by another PV (its a misconfiguration, `volumeHandle` must be unique per PV, but failing explicitly would be the best CX, as opposed to silently creating a volume violating the limit)

charts/aws-mountpoint-s3-csi-driver/values.yaml
```diff
@@ -152,10 +152,22 @@ experimental:
...
+  - maxVolumesPerNode: 10
+    # Total resource requests and limits for the mounter DaemonSet pod.
+    # requests.memory should be sized to maxVolumesPerNode × per-mount memory
+    # (e.g., 4 mounts × 512Mi = 2Gi, or 10 mounts × 4Gi = 40Gi).
```
vladem, line 164: consider removing everything after this line in this comment, todo-s tend to be never addressed

pkg/driver/driver.go
```diff
@@ -131,7 +133,15 @@ func NewDriver(endpoint string, mpVersion string, nodeID string) (*Driver, error
 		}
 		go dm.StartCommDirWatch(stopCh)
 
-		nodeMounter = dm
+		maxVolumesPerNode, err := util.GetEnvAsInt(maxVolumesPerNodeEnvName)
+		if err != nil {
+			return nil, err
+		}
+		if maxVolumesPerNode < 1 {
```
vladem: discussion point: don't we want to allow customers disable the limit in daemonset architecture?

tests/e2e-kubernetes/testsuites/volumelimits.go
```diff
@@ -0,0 +1,312 @@
...
+	})
+
+	// This checks that CSIMaxVolumeLimitChecker works as expected.
+	// A randomly chosen node should be able to handle as many CSI volumes as
+	// it claims to handle in CSINode.Spec.Drivers[x].Allocatable.
+	// The test uses one single pod with a lot of volumes to work around any
+	// max pod limit on a node.
+	// And one extra pod with a CSI volume should get Pending with a condition
+	// that says it's unschedulable because of volume limit.
+	// BEWARE: the test may create lot of volumes and it's really slow.
```
vladem: how slow? also we assume that the node has no s3 volumes currently present, is it possible to verify that assumption at the start of the test?

tests/e2e-kubernetes/testsuites/volumelimits.go
```diff
@@ -0,0 +1,312 @@
...
+			}
+			if pod.Status.Phase != v1.PodPending {
+				return true, fmt.Errorf("expected pod to be in phase Pending, but got phase: %v", pod.Status.Phase)
+			}
+			return false, nil
+		})
+		framework.ExpectNoError(err)
+	})
+
+	// TODO: Uncomment when pod sharing is implemented. Alternatively just add to test above
```
vladem: may be a WIP branch with this? I am a bit concerned with TODO scattered around the code base, which might be easily forgotten

pkg/driver/driver.go
jet-tong: That's a good point. I added it to avoid misconfiguration leading to e.g. OOM when number of PVs are much greater than requested memory for the daemonset mounter. But I was also thinking we should allow customers to disable the limit which reflects the default behaviour in pod mode as well. 

I'll see whether / how I can do that. Suggestion: default 4 volumes 2Gi memory, and allow users to configure 0 to disable volume limits - that way users / cluster admins are aware of possible OOM when setting this behaviour.

tests/e2e-kubernetes/testsuites/volumelimits.go
jet-tong: 1-2s each for 11 pods from what I see which is not very slow. 
(I copied this line from k8s [volumelimits.go](https://github.com/kubernetes/kubernetes/blob/24e2b02af5543d7910c2bb074c7264df5a8f0467/test/e2e/storage/testsuites/volumelimits.go#L188)).

GitHub test took 30s, see [this run](https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/29030911091/job/86774012659#step:9:1677).

Asserting there are no other s3 volumes would be a good addition, will check, thanks.

tests/e2e-kubernetes/testsuites/volumelimits.go
jet-tong: I was hoping we can catch and address all the TODOs before we merge back to main via git diff, but that's a valid concern actually. since pod sharing PR https://github.com/awslabs/mountpoint-s3-csi-driver/pull/862 is still open I'll keep this for now and if required remove it depending on how quickly these PRs will be merged.

charts/aws-mountpoint-s3-csi-driver/values.yaml
jet-tong: Thanks, that's a very good point.

Confirmed that k8s count PVs with their volumeHandles in [nodevolumelimits/csi.go](https://github.com/kubernetes/kubernetes/blob/6a5cc912a3c2a2c241fe05abea563079bf574abd/pkg/scheduler/framework/plugins/nodevolumelimits/csi.go#L499), and uses `fmt.Sprintf("%s/%s", driverName, volumeHandle)` in [getVolumeUniqueName](https://github.com/kubernetes/kubernetes/blob/6a5cc912a3c2a2c241fe05abea563079bf574abd/pkg/scheduler/framework/plugins/nodevolumelimits/csi.go#L688).

charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
jet-tong: Good catch, I did put this in mounter-daemonset.yaml:
```yaml
{{- if eq .Values.experimental.mounterMode "daemonset" }}
{{- if gt (len .Values.daemonsetMounters) 1 }}
{{- fail "Only one entry in daemonsetMounters is supported currently" }}
{{- end }}
```
but didn't add to node.yaml.

I'll use `if ne (len .Values.daemonsetMounters) 1` to fail the 0 case too.

tests/e2e-kubernetes/testsuites/volumelimits.go
jet-tong: Removed that comment, and added countS3VolumesOnNode helper to assert S3 volume count before/after.

tests/e2e-kubernetes/testsuites/volumelimits.go
jet-tong: On second thought you're right and the TODO will also pollute the git diff with untested test, so I've removed it for now.

FYI @rajdchak if we merge volumelimits first we can add pod sharing case for volume limits back later.

charts/aws-mountpoint-s3-csi-driver/values.yaml
jet-tong: Added this comment in values.yaml;
```yaml
    # IMPORTANT: volumeHandle must be unique per PV as the scheduler uses it to 
    # track volume usage.
```

Will keep this comment open for now.

pkg/driver/driver.go
jet-tong: I've now allowed 0 maxVolumesPerNode to mean unbounded behaviour, with comment + warning in code.

Will keep this comment unresolved until we agree this is the behaviour we want (and if the warnings are sufficient or if there's more we can do)

tests/e2e-kubernetes/testsuites/volumelimits.go
jet-tong: Unresolved comment for me to address, as pod sharing PR now landed.

jet-tong: Re-added this test now, resolving comment.

charts/aws-mountpoint-s3-csi-driver/values.yaml
jet-tong: I'll be doing volumeHandle uniqueness check in a follow-up PR, so will resolve the comment for now.

pkg/driver/driver.go
jet-tong: For now let's allow 0 maxVolumesPerNode since that is a valid knob customers can set anyways by introducing a really large number; allowing an explicit option allows us to warn them. 

When we have a better picture on the user experience on setting these values, we can determine whether to allow disabling volume limits or not (and/or readjust the warning messages).

charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
```diff
@@ -105,10 +105,19 @@ spec:
             - name: INSTALLATION_TYPE
               value: {{ if .Values.isEKSAddon }}eks-addon{{ else }}helm{{ end }}
             {{- if eq .Values.experimental.mounterMode "daemonset" }}
+            {{- if ne (len .Values.daemonsetMounters) 1 }}
+            {{- fail "daemonsetMounters must have exactly one entry" }}
+            {{- end }}
+            {{- $mounter := index .Values.daemonsetMounters 0 }}
+            {{- if lt (int $mounter.maxVolumesPerNode) 0 }}
+            {{- fail "daemonsetMounters[].maxVolumesPerNode must be >= 0" }}
```
rajdchak, line 113: what does 0 volumes on a node mean?

charts/aws-mountpoint-s3-csi-driver/values.yaml
```diff
@@ -152,10 +152,20 @@ experimental:
...
 # Only used when experimental.mounterMode is "daemonset".
 daemonsetMounters:
-  - resources:
+    # Maximum number of distinct S3 volumes (Mountpoint processes) per node.
+    # Advertised to the Kubernetes scheduler via MaxVolumesPerNode.
+    # Set to 0 to disable the limit (unbounded); this risks resource exhaustion
+    # (memory OOM, or disk exhaustion with cache) on the mounter DaemonSet pod.
+    # IMPORTANT: volumeHandle must be unique per PV as the scheduler uses it to 
+    # track volume usage.
+  - maxVolumesPerNode: 4
```
rajdchak, line 161: when customers upgrade to V3 this is the max volumes they can have on a node. But if customers have already mounted volumes for V1/V2 on the nodes and the count of those is more than the value we set here, what happens then?

charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
jet-tong: It means disabling volume limits -  kubelet [doesn't write allocatable value if 0](https://github.com/kubernetes/kubernetes/blob/6a5cc912a3c2a2c241fe05abea563079bf574abd/pkg/volume/csi/nodeinfomanager/nodeinfomanager.go#L634-L643). I hope it's clear from values.yaml how to set this (or could it be improved?)

rajdchak: if 0, do we need to add a warning message that this means unlimited and might cause resource exhaustion?

charts/aws-mountpoint-s3-csi-driver/values.yaml
jet-tong: Great question - it only affects scheduling decisions; already scheduled pods and volumes shouldn't be affected, and I think it'll just reject new pods in that node. 

I can verify this manually too to make sure - I didn't consider v2 pods in the PR much. Could we maybe verify this in a new upgrade test from v2 to v3 with existing pods?

charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
jet-tong: I added warning in logs in [this commit](https://github.com/awslabs/mountpoint-s3-csi-driver/pull/851/changes/6863aec8061faf87c01d48b34feef5cf53ed1b72) - is it worth adding warning in NOTES.txt also to show on installation?

rajdchak: If we are warning already, then we don't have to add the warning here again.

charts/aws-mountpoint-s3-csi-driver/values.yaml
jet-tong: I've ran this test and confirmed this is the behaviour.

- Deployed in pod mode, and added 10 s3 workloads
- helm upgrade to daemonset mode (volume limit 4), and added 1 s3 workload
- All pod mode workload pods stay working, and the new s3 workload in daemonset mode Pending state with reason "FailedScheduling".

rajdchak (approved):

PR #852, opened 2026-07-01
Merge main to feature daemonset workflow

rajdchak:
*Issue #, if available:*

*Description of changes:*

Workflow to merge changes from main to daemonset feature branch biweekly.

### DIFF
--- a/.github/workflows/merge-main-to-feature-daemonset.yaml
+++ b/.github/workflows/merge-main-to-feature-daemonset.yaml
@@ -1,0 +1,114 @@
+name: Merge main to feature/daemonset-architecture branch
+
+on:
+  schedule:
+    # Runs bi-weekly on the 1st and 15th of each month at 9:00 UTC
+    - cron: '0 9 1,15 * *'
+  workflow_dispatch: # Allow manual trigger
+
+permissions:
+  contents: write
+  pull-requests: write
+  issues: write
+
+jobs:
+  merge-main-to-feature:
+    runs-on: ubuntu-latest
+    env:
+      FEATURE_BRANCH: feature/daemonset-architecture
+      BASE_BRANCH: main
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ env.FEATURE_BRANCH }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Configure git
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+
+      - name: Fetch latest main
+        run: |
+          git fetch origin ${{ env.BASE_BRANCH }}
+
+      - name: Check if main has new commits
+        id: check
+        run: |
+          MERGE_BASE=$(git merge-base HEAD origin/${{ env.BASE_BRANCH }})
+          MAIN_HEAD=$(git rev-parse origin/${{ env.BASE_BRANCH }})
+          if [ "$MERGE_BASE" = "$MAIN_HEAD" ]; then
+            echo "up_to_date=true" >> "$GITHUB_OUTPUT"
+            echo "Feature branch already contains all commits from main. Nothing to do."
+          else
+            echo "up_to_date=false" >> "$GITHUB_OUTPUT"
+            echo "Main has new commits to merge."
+          fi
+
+      - name: Attempt merge
+        if: steps.check.outputs.up_to_date == 'false'
+        id: merge
+        run: |
+          if git merge origin/${{ env.BASE_BRANCH }} --no-edit; then
+            echo "conflict=false" >> "$GITHUB_OUTPUT"
+          else
+            git merge --abort
+            echo "conflict=true" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Create PR with clean merge
+        if: steps.check.outputs.up_to_date == 'false' && steps.merge.outputs.conflict == 'false'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          MERGE_BRANCH="merge-main-to-feature-daemonset-$(date +%Y%m%d)"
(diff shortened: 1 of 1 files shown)


.github/workflows/merge-main-to-feature-daemonset.yaml
```diff
@@ -0,0 +1,114 @@
...
+jobs:
+  merge-main-to-feature:
+    runs-on: ubuntu-latest
+    env:
+      FEATURE_BRANCH: feature/daemonset-architecture
+      BASE_BRANCH: main
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
jet-tong, line 23: nit: can use newer v7 checkout

.github/workflows/merge-main-to-feature-daemonset.yaml
```diff
@@ -0,0 +1,114 @@
+name: Merge main to feature/daemonset-architecture branch
+
+on:
+  schedule:
+    # Runs bi-weekly on the 1st and 15th of each month at 9:00 UTC
+    - cron: '0 9 1,15 * *'
```
jet-tong, line 6: throwaway comment: I would prefer it to run on specific day of week (e.g. Sunday) if possible, but I couldn't find a way to do that. Every Sunday ('0 9 * * 0') feels quite frequent though. 
Reason: it triggers trusted workflow which might come unexpected if it comes on a weekday 9am (I guess dependabot already does that though)

nit: bi-weekly comment is not quite accurate, perhaps 'twice a month' is better (bi-monthly with the other bi- meaning is very ambiguous too I guess)

.github/workflows/merge-main-to-feature-daemonset.yaml
```diff
@@ -0,0 +1,114 @@
...
+          else
+            git merge --abort
+            echo "conflict=true" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Create PR with clean merge
+        if: steps.check.outputs.up_to_date == 'false' && steps.merge.outputs.conflict == 'false'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
```
jet-tong, line 66: Do we want to run a check to see if we can simply update existing dependabot PR?

Use case: before running integration tests, a new commit lands on main, OR we want to add an extra change to main and port it over to feature branch before we merge this PR.

It would also solve problems where we manually dispatched workflow while the PR is still open and there's a branch name conflict (branch already exists)

.github/workflows/merge-main-to-feature-daemonset.yaml
```diff
@@ -0,0 +1,114 @@
...
+          This is an automated bi-weekly merge of \`main\` into \`feature/daemonset-architecture\`.
+
+          No conflicts detected. Please review and merge.
+
+          > **Note:** Please delete the \`${MERGE_BRANCH}\` branch after merging this PR.
+
+          ---
+          *Created automatically by the bi-weekly merge workflow.*"
+
+      - name: Open issue for conflict resolution
```
jet-tong, line 86: Perhaps slack message > public github issue? (e.g. https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/.github/workflows/notify-slack.yaml) I think we should reserve github issues for the public. 

Also, would be good to prevent duplicate issues too like PR part.

.github/workflows/merge-main-to-feature-daemonset.yaml
```diff
@@ -0,0 +1,114 @@
...
+      FEATURE_BRANCH: feature/daemonset-architecture
+      BASE_BRANCH: main
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ env.FEATURE_BRANCH }}
+          token: ${{ secrets.GITHUB_TOKEN }}
```
jet-tong, line 27: May I ask why token is needed here?

.github/workflows/merge-main-to-feature-daemonset.yaml
```diff
@@ -0,0 +1,114 @@
...
+            git merge --abort
+            echo "conflict=true" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Create PR with clean merge
+        if: steps.check.outputs.up_to_date == 'false' && steps.merge.outputs.conflict == 'false'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          MERGE_BRANCH="merge-main-to-feature-daemonset-$(date +%Y%m%d)"
```
vladem, line 67: how about adding a timestamp to the branch name? otherwise it'll fail if triggered twice in a day

.github/workflows/merge-main-to-feature-daemonset.yaml
```diff
@@ -0,0 +1,114 @@
...
+          git checkout -b "$MERGE_BRANCH"
+          git push origin "$MERGE_BRANCH"
+
+          gh pr create \
```
yerzhan7, line 71: FYI: Currently our repo setting for PRs only allows "squash commits". But you would need proper merge commits to properly handle multiple periodic merges.

.github/workflows/merge-main-to-feature-daemonset.yaml
rajdchak: As discussed, to keep a track of this will open a PR instead of an issue with a dummy commit

.github/workflows/merge-main-to-feature-daemonset.yaml
rajdchak: As discussed, this kind of scenario might not be frequent. Let's start simple for now we can then update the workflow if needed

jet-tong: Can use existing approach as discussed. 
(I was hoping to reuse the same branch so consecutive workflow runs would simply update existing PR - but we can always close PR and open a new one when we want to take in new commits from main before mergin).

Issue #855, opened 2026-07-06
Workload pod is not ungated when the pod has both EBS and S3 type PVC's


Kathir163:
/kind bug
**What happened?**
When a workload pod has both EBS and S3 PVC's and when the headroom pod feature is enabled the workload pod is never ungated. The root cause is determined to be this PR - https://github.com/awslabs/mountpoint-s3-csi-driver/pull/708. The check added here does not differentiate between EBS volumes and S3 volumes. EBS volumes are created only after the pod is scheduled but the s3-csi-controller doesn't ungate the pod until all pvc's volumes are bound hence ending up in a deadlock.

**What you expected to happen?**
The pod to get ungated after the headroom pods are created.

**How to reproduce it (as minimally and precisely as possible)?**
Create a workload pod having both a EBS and S3 type PVC's with headroom feature enabled in the driver.
**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`):1.30
- Driver version:1.27

PR #856, opened 2026-07-06
Update the logic for checking unboundPVC's of the workload pod 

Kathir163:
*Issue:#855*
    
    *Description of changes:*
    Add extra check while checking for unbound PVC's to ensure only S3 type PVC's are taken into account.
    If we have non S3 PVC's whose storage class has binding mode set to 'WaitForFirstConsumer' the PVC would be in pending state until the pod gets scheduled.
    This results in a deadlock situation where the non S3PVC is waiting for the pod to get scheduled but the controller has scheduleGated the workload pod
    for the pvc to get bound.
    

cmd/aws-s3-csi-controller/csicontroller/reconciler.go
```diff
@@ -260,12 +260,16 @@ func (r *Reconciler) getWorkloadVolumes(
...
 
 		// If PVC has no bound PVs yet, `getBoundPVForPodClaim` will return `errPVCIsNotBoundToAPV`.
-		// In this case we'll just return `reconcile.Result{Requeue: true}` here, which will bubble up to the
-		// original `Reconcile` call and will cause a retry for this Pod with an exponential backoff.
+		// We only block ungating (via `status = Requeue`) for PVCs that are likely S3-backed.
 		pvc, pv, err := r.getBoundPVForPodClaim(ctx, workloadPod, podPVC)
 		if err != nil {
 			if errors.Is(err, errPVCIsNotBoundToAPV) {
-				status = Requeue
+				isS3, checkErr := r.isUnboundS3PVC(ctx, workloadPod.Namespace, podPVC.ClaimName)
```
vladem: How about embedding the S3/non-S3 check into `getBoundPVForPodClaim`? Something like:

```go
// getBoundS3PVForPodClaim tries to find bound PV and PVC from given `claim`.
// It `errPVCIsNotBoundToAPV` if PVC is not bound to a PV yet to be eventually retried.
// It `errPVCIsNotS3` if PVC is guaranteed not be backed by our driver, we skip such PVC-s.
func (r *Reconciler) getBoundS3PVForPodClaim(
	ctx context.Context,
	pod *corev1.Pod,
	claim *corev1.PersistentVolumeClaimVolumeSource,
) (*workloadVolume, error) {
```

And then:

```go
// If PVC has no bound PVs yet, `getBoundPVForPodClaim` will return `errPVCIsNotBoundToAPV`.
// In this case we'll just return `reconcile.Result{Requeue: true}` here, which will bubble up to the
// original `Reconcile` call and will cause a retry for this Pod with an exponential backoff.
// If PVC is not backed by S3 driver, we ignore it, even if it's unbound.
volume, err := r.getBoundS3PVForPodClaim(ctx, workloadPod, podPVC)
if err != nil {
	// Ignore non-S3 PVC-s
	if errors.Is(err, errPVCIsNotS3) {
		continue
	}
	if errors.Is(err, errPVCIsNotBoundToAPV) {
		status = Requeue
	} else {
		errs = append(errs, err)
	}
	continue
}

volumes = append(volumes, volume)
```

vladem (commented):
Thanks for reporting the issue and working on a fix for it! The logic looks good to me, my only suggestion is about code structure. Also need to address failing tests, before we can merge this, in particular `late PV and PVC binding`.

cmd/aws-s3-csi-controller/csicontroller/reconciler.go
```diff
@@ -918,6 +922,25 @@ func (r *Reconciler) getBoundPVForPodClaim(
...
+// isUnboundS3PVC checks whether the given unbound PVC is backed by the S3 CSI Driver.
+// It fetches the PVC to get `spec.volumeName`, then fetches the PV to inspect the CSI driver.
+// Returns false if the PVC has no volume name (dynamically provisioned, not yet bound)
+// or if the PV is not backed by the S3 CSI Driver.
+func (r *Reconciler) isUnboundS3PVC(ctx context.Context, namespace, claimName string) (bool, error) {
+	pvc := &corev1.PersistentVolumeClaim{}
+	if err := r.Get(ctx, types.NamespacedName{Namespace: namespace, Name: claimName}, pvc); err != nil {
+		return false, fmt.Errorf("failed to get PVC %s/%s: %w", namespace, claimName, err)
+	}
+	if pvc.Spec.VolumeName == "" {
```
vladem: Please add an explanation and TODO: 

```
// S3 CSI Driver doesn't support dynamic provisioning, so the claim has to reference a PV
// TODO: support dynamic provisioning
```

Kathir163: Have added the comments

Kathir163:
@vladem I have tested the E2E tests locally and they do pass.

PR #862, opened 2026-07-15
Sharing mountpoint processes changes

rajdchak:
*Issue #, if available:*

*Description of changes:*
1. Multiple pods on the same node using the same PV now share a single mount-s3 FUSE process via bind mounts instead of each getting a separate one
2 First pod triggers a FUSE mount at a stable source path (/var/lib/kubelet/plugins/s3.csi.aws.com/mnt/<volumeID>), subsequent pods get bind mounts from that source
3. Reference-counted MountMap (lock-free sync.Map with per-volume mutex) tracks active consumers; last unmount tears down the FUSE source
4. Sharing is validated: pods must have matching mount options, auth source, service account, EKS role ARN, namespace, and fsGroup — mismatches are rejected
5. .meta.json files persisted alongside source mounts enable state recovery after CSI node pod restart via RebuildMountMap() (scans meta files + /proc/self/mountinfo to reconstruct refcounts)
6. After mounter pod crash: existing pods have dead mounts (require recreation), but new pods automatically get a fresh FUSE source
7. E2E test suite covers: basic sharing, refcount correctness, parameter validation, concurrent mount/unmount, pod churn, CSI node restart recovery, persistence lifecycle, mounter crash recovery

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -125,7 +125,15 @@ func NewDriver(endpoint string, mpVersion string, nodeID string) (*Driver, error
 		klog.Info("Using daemonset mounter mode")
 		klog.Info("Note: s3-csi-daemonset-mounter uses OnDelete update strategy - helm upgrade will not restart/upgrade mounter pods automatically")
 
-		dm := mounter.NewDaemonsetMounter(clientset, nodeID, mpmounter.New(), credProvider, nil)
+		dm := mounter.NewDaemonsetMounter(clientset, nodeID, mpmounter.New(), credProvider, nil, nil, nil)
+
+		// Rebuild mount map from persisted .meta.json files + kernel mount table.
+		// Fatal on failure: operating with an empty map while FUSE mounts exist
+		// would cause duplicate mounts or incorrect unmounts.
+		if err := dm.RebuildMountMap(); err != nil {
+			klog.Fatalf("Failed to rebuild mount map from disk: %v", err)
+		}
+
 		if err := dm.DiscoverCommDir(context.Background()); err != nil {
 			klog.Fatalf("Failed to discover mounter pod: %v", err)
 		}
--- a/pkg/driver/node/mounter/daemonset_mounter.go
+++ b/pkg/driver/node/mounter/daemonset_mounter.go
@@ -28,6 +28,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync/atomic"
 	"time"
 
@@ -36,8 +37,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
-
-	k8sstrings "k8s.io/utils/strings"
 
 	"github.com/awslabs/mountpoint-s3-csi-driver/pkg/driver/node/credentialprovider"
 	"github.com/awslabs/mountpoint-s3-csi-driver/pkg/driver/node/envprovider"
@@ -79,6 +78,12 @@ var (
 // mountSyscallFunc performs the FUSE mount and returns the fd. Injectable for testing.
 type mountSyscallFunc func(target string, opts mpmounter.MountOptions) (int, error)
 
+// bindMountSyscallFunc performs a bind mount. Injectable for testing.
+type bindMountSyscallFunc func(source, target string) error
+
+// mountInfoProviderFunc reads the kernel mount table. Injectable for testing.
+type mountInfoProviderFunc func() ([]mountInfoEntry, error)
+
 // DaemonsetMounter is a [Mounter] that delegates Mountpoint process management
 // to a secondary daemonset running on the same node. It communicates via the
 // secondary pod's emptyDir volume, accessed through the kubelet pod directory.
@@ -95,183 +100,244 @@ type DaemonsetMounter struct {
 	rediscoverCh chan struct{}
 
 	// Injectable for testing. nil = use default.
-	mountSyscall mountSyscallFunc
+	mountSyscall      mountSyscallFunc
+	bindMountSyscall  bindMountSyscallFunc
+	mountInfoProvider mountInfoProviderFunc
+
+	// mountMap tracks shared source mounts for pod-sharing.
+	// Mount/Unmount use reference-counted sharing via this map.
+	mountMap *MountMap
 }
 
 // NewDaemonsetMounter creates a new [DaemonsetMounter].
-// mountSyscall may be nil, in which case the default FUSE mount implementation is used.
+// mountSyscall, bindMountSyscall, and mountInfoProvider may be nil,
+// in which case the default implementations are used.
 func NewDaemonsetMounter(clientset kubernetes.Interface, nodeID string, mount *mpmounter.Mounter,
-	credProvider credentialprovider.ProviderInterface, mountSyscall mountSyscallFunc) *DaemonsetMounter {
+	credProvider credentialprovider.ProviderInterface, mountSyscall mountSyscallFunc, bindMountSyscall bindMountSyscallFunc, mountInfoProvider mountInfoProviderFunc) *DaemonsetMounter {
 	return &DaemonsetMounter{
-		clientset:    clientset,
-		nodeID:       nodeID,
-		kubeletPath:  util.ContainerKubeletPath(),
-		mount:        mount,
-		credProvider: credProvider,
-		rediscoverCh: make(chan struct{}, 1),
-		mountSyscall: mountSyscall,
+		clientset:         clientset,
+		nodeID:            nodeID,
+		kubeletPath:       util.ContainerKubeletPath(),
+		mount:             mount,
+		credProvider:      credProvider,
+		rediscoverCh:      make(chan struct{}, 1),
+		mountSyscall:      mountSyscall,
+		bindMountSyscall:  bindMountSyscall,
+		mountInfoProvider: mountInfoProvider,
+		mountMap:          NewMountMap(),
 	}
 }
 
-// Mount mounts the given S3 bucket at the target path.
+// Mount mounts the given S3 bucket at the target path with pod-sharing support.
 //
-// It performs the following steps:
-//  1. Provides credentials for the mount
-//  2. Opens /dev/fuse and performs the kernel FUSE mount on target
-//  3. Sends mount options (including FUSE FD) to the secondary daemonset via UDS
-//  4. Waits for Mountpoint to start serving (or an error)
+// Flow:
+//  1. If target is already mounted (republish): refresh credentials without locking, return early
+//  2. Acquire per-volume lock via MountMap (serializes concurrent NodePublishVolume for same PV)
+//  3. Provision credentials (under lock to avoid race with cleanup on failure)
+//  4. If an existing healthy source mount exists for this volumeID → bind mount to target
+//  5. If no source exists (or dead) → FUSE mount at source path, send FD to mounter, wait for readiness
+//  6. Bind mount source → target
+//  7. Track the target in MountMap for refcounting
+//
+// On initial mount failure (fuseMount or bindMount), credentials are cleaned up immediately
+// under the lock. On unmount of last consumer, the FUSE source is unmounted (causing mount-s3
+// to exit via kernel FUSE teardown) and credentials are removed.
 func (dm *DaemonsetMounter) Mount(ctx context.Context, bucketName string, target string,
 	credentialCtx credentialprovider.ProvideContext, args mountpoint.Args, fsGroup string, userEnv envprovider.Environment) error {
-	mountId, err := GetMountId(target)
-	if err != nil {
-		return err
-	}
-	volumeId := credentialCtx.VolumeID
-	mountSuccess := false
 
-	commDir, err := dm.GetCommDir()
+	// Extract PV name from target path to use as the volume identifier for sharing and filesystem paths.
+	// PV names are Kubernetes resource names — guaranteed DNS-safe (no '/', no '..', alphanumeric + '-' + '.').
+	// Target path format: /var/lib/kubelet/pods/<podUID>/volumes/kubernetes.io~csi/<pv-name>/mount
+	parsedTarget, err := targetpath.Parse(target)
 	if err != nil {
-		return fmt.Errorf("failed to find s3-csi-daemonset-mounter pod: %w. %s", err, helpMessageForCheckingMounterPodStatus())
+		return fmt.Errorf("failed to parse target path %q: %w", target, err)
 	}
+	volumeID := parsedTarget.VolumeID // This is the PV name
 
-	alreadyMounted, err := dm.IsMountPoint(target)
+	// Idempotency: if target is already a healthy Mountpoint mount, refresh creds and return.
+	// Kubelet may call NodePublishVolume repeatedly (requiresRepublish, retries).
+	isMounted, err := dm.IsMountPoint(target)
 	if err != nil && !errors.Is(err, fs.ErrNotExist) {
 		return fmt.Errorf("failed to check if target %q is a mount point (mount target is possibly"+
 			" corrupted, manual pod re-creation %s might be required for mount recovery): %w",
 			target, credentialCtx.WorkloadPodID, err)
 	}
-
-	// Cleanup provisioned credentials only if this is a new mount (not a republish) and mount failed
-	defer func() {
-		if alreadyMounted || mountSuccess {
-			return
+	if isMounted {
+		// Republish: refresh credentials for the existing mount (token rotation).
+		commDir, err := dm.GetCommDir()
+		if err == nil {
+			if _, err := dm.provideCredentials(ctx, commDir, volumeID, &credentialCtx); err != nil {
+				klog.Errorf("DaemonsetMounter: failed to refresh credentials on republish for %s: %v", target, err)
+			}
 		}
-		if err := dm.cleanupCredentials(commDir, mountId, credentialCtx.ToCleanupCtx()); err != nil {
-			klog.Errorf("DaemonsetMounter: failed to clean up credential directory for mount %s: %v", mountId, err)
-		}
-	}()
-
-	// Provision credentials before the alreadyMounted early return so republish refreshes them
-	credsEnv, err := dm.provideCredentials(ctx, commDir, mountId, &credentialCtx)
-	if err != nil {
-		return err
-	}
-
-	// Idempotency: if target is already a healthy Mountpoint mount, return early
-	if alreadyMounted {
 		klog.V(4).Infof("DaemonsetMounter: target %s is already mounted, credentials refreshed", target)
-		mountSuccess = true
 		return nil
 	}
 
-	// Ensure target directory exists (kubelet may not have created it yet)
-	if err := os.MkdirAll(target, targetDirPerm); err != nil {
-		return fmt.Errorf("failed to create target directory %q: %w", target, err)
-	}
-
-	defer func() {
-		if mountSuccess {
-			return
-		}
-		// TODO(vlaad): might leak a mount, implement periodic stale mount cleanup
-		if umErr := dm.mount.Unmount(target); umErr != nil {
-			klog.Errorf("Failed to unmount %q during cleanup: %v", target, umErr)
-		}
-	}()
-
-	// Perform FUSE mount and send options to secondary daemonset
-	if err := dm.mountS3AtTarget(ctx, target, bucketName, args, mountId, volumeId, commDir, userEnv, credsEnv); err != nil {
-		return err
-	}
-
-	mountSuccess = true
-	klog.V(4).Infof("DaemonsetMounter: volume %s (mount %s) mounted at %s", volumeId, mountId, target)
-	return nil
+	// Pod-sharing: use MountMap to share a single Mountpoint process per volume.
+	// Credential provisioning happens inside mountOrShareSource under the per-volume lock
+	// to avoid races between concurrent NodePublishVolume calls for the same PV.
+	return dm.mountOrShareSource(ctx, bucketName, target, volumeID, credentialCtx, args, fsGroup, userEnv)
 }
 
-// Unmount unmounts the FUSE filesystem at target.
-// This causes the Mountpoint process in the secondary daemonset to exit.
-func (dm *DaemonsetMounter) Unmount(ctx context.Context, target string, cleanupCtx credentialprovider.CleanupContext) error {
-	// cleanup the error file and credentials
+// mountOrShareSource implements the pod-sharing Mount flow using MountMap.
+func (dm *DaemonsetMounter) mountOrShareSource(ctx context.Context, bucketName string, target string,
+	volumeID string, credentialCtx credentialprovider.ProvideContext, args mountpoint.Args, fsGroup string, userEnv envprovider.Environment) error {
+
+	// Get or create the per-volume entry, then lock it.
+	// Retry loop ensures we hold the canonical entry — not one orphaned by a concurrent unmount/delete.
+	var entry *MountEntry
+	for {
+		entry, _ = dm.mountMap.GetOrCreate(volumeID)
+		entry.mu.Lock()
+		if dm.mountMap.Get(volumeID) == entry {
+			break // we hold the canonical entry
+		}
+		entry.mu.Unlock() // orphaned entry (deleted by concurrent unmount), retry
+	}
+	defer entry.mu.Unlock()
+
+	// Provision credentials under the lock to avoid races between concurrent
+	// NodePublishVolume calls for the same PV (all share one credential file).
 	commDir, err := dm.GetCommDir()
 	if err != nil {
-		return fmt.Errorf("failed to find s3-csi-daemonset-mounter pod: %w. %s", err, helpMessageForCheckingMounterPodStatus())
+		return fmt.Errorf("connection to s3-csi-daemonset-mounter not yet established, allowing kubelet to retry NodePublishVolume: %w. %s", err, helpMessageForCheckingMounterPodStatus())
 	}
-	mountId, err := GetMountId(target)
+	credsEnv, err := dm.provideCredentials(ctx, commDir, volumeID, &credentialCtx)
 	if err != nil {
 		return err
 	}
 
-	if err = removeIfExists(filepath.Join(commDir, GetErrorFileName(mountId))); err != nil {
-		return fmt.Errorf("failed to remove the error file: %w", err)
+	// Build mount params for this request — used for validation and stored on first mount.
+	incomingParams := MountParams{
+		MountOptions:             args.SortedList(),
+		AuthenticationSource:     credentialCtx.AuthenticationSource,
+		ServiceAccountName:       credentialCtx.ServiceAccountName,
+		ServiceAccountEKSRoleARN: credentialCtx.ServiceAccountEKSRoleARN,
+		PodNamespace:             credentialCtx.PodNamespace,
+		FSGroup:                  fsGroup,
 	}
 
-	err = dm.cleanupCredentials(commDir, mountId, cleanupCtx)
-	if err != nil {
-		return fmt.Errorf("failed to cleanup credentials: %w", err)
+	sourcePath := SourceMountPath(dm.kubeletPath, volumeID)
+
+	if entry.initialized {
+		// Existing source — validate compatibility before sharing.
+		if err := entry.Params.ValidateCompatibility(&incomingParams); err != nil {
+			return fmt.Errorf("cannot share mount for volume %s: %w", volumeID, err)
+		}
+
+		// Check health before sharing.
+		if dm.IsSourceHealthy(entry.SourcePath) {
+			// Healthy source: bind mount to new target.
+			if err := dm.BindMount(entry.SourcePath, target); err != nil {
+				return err
+			}
+			entry.RefCount++
+			entry.Targets = append(entry.Targets, target)
+			klog.V(4).Infof("DaemonsetMounter: shared existing mount for volume %s → %s (refcount=%d)",
+				volumeID, target, entry.RefCount)
+			return nil
+		}
+		// Dead source: clean up and fall through to new mount.
+		klog.V(2).Infof("DaemonsetMounter: source %s is dead for volume %s, recovering", entry.SourcePath, volumeID)
+		if err := dm.mount.Unmount(entry.SourcePath); err != nil {
+			if !os.IsNotExist(err) {
+				return fmt.Errorf("failed to unmount dead source %q for volume %s, will retry: %w", entry.SourcePath, volumeID, err)
+			}
+		}
+		os.Remove(entry.SourcePath)
+		entry.initialized = false
 	}
 
-	// finally unmount (this order ensures retrying cleanup)
-	err = dm.mount.Unmount(target)
-	if err != nil {
-		return fmt.Errorf("failed to unmount %q: %w", target, err)
+	// New mount: FUSE mount at source, then bind to target.
+	mountId := volumeID // In sharing mode, one mount per volume
+	if err := dm.fuseMount(ctx, bucketName, sourcePath, mountId, args, userEnv, credsEnv); err != nil {
+		dm.cleanupCredentials(commDir, volumeID, credentialCtx.ToCleanupCtx())
+		return err
 	}
 
-	klog.V(4).Infof("DaemonsetMounter: volume %s unmounted from %s", cleanupCtx.VolumeID, target)
+	// Bind mount source → target.
+	if err := dm.BindMount(sourcePath, target); err != nil {
+		// Cleanup source and credentials on bind failure.
+		dm.mount.Unmount(sourcePath)
+		os.Remove(sourcePath)
+		dm.cleanupCredentials(commDir, volumeID, credentialCtx.ToCleanupCtx())
+		return err
+	}
+
+	// Populate the entry.
+	entry.SourcePath = sourcePath
+	entry.MountID = mountId
+	entry.Params = incomingParams
+	entry.RefCount = 1
+	entry.Targets = []string{target}
+	entry.initialized = true
+
+	// Persist meta to disk for recovery after driver restart.
+	if err := WriteMeta(dm.kubeletPath, entry); err != nil {
+		klog.Errorf("DaemonsetMounter: failed to write meta for volume %s: %v (non-fatal)", volumeID, err)
+	}
+
+	klog.V(4).Infof("DaemonsetMounter: new shared mount for volume %s at source %s → %s", volumeID, sourcePath, target)
 	return nil
 }
 
-// IsMountPoint returns whether the given target is a Mountpoint mount.
-func (dm *DaemonsetMounter) IsMountPoint(target string) (bool, error) {
-	return dm.mount.CheckMountpoint(target)
-}
+// fuseMount performs the FUSE mount + FD send + wait cycle at the given path.
+// Credentials are already provisioned by the caller (Mount).
+func (dm *DaemonsetMounter) fuseMount(ctx context.Context, bucketName string, mountPath string,
+	mountId string, args mountpoint.Args, userEnv envprovider.Environment, credsEnv envprovider.Environment) error {
 
-// mountS3AtTarget performs the kernel FUSE mount at target and sends mount options
-// (including the FUSE fd) to the secondary daemonset. On success the mount remains
-// at target; on failure any partial mount is cleaned up.
-func (dm *DaemonsetMounter) mountS3AtTarget(ctx context.Context, target string, bucketName string,
-	args mountpoint.Args, mountId string, volumeId string, commDir string,
-	userEnv envprovider.Environment, credsEnv envprovider.Environment) error {
-	// Remove old error file if exists
-	errFilePath := filepath.Join(commDir, GetErrorFileName(mountId))
-	if err := removeIfExists(errFilePath); err != nil {
-		return fmt.Errorf("failed to remove stale error file %q: %w", errFilePath, err)
+	commDir, err := dm.GetCommDir()
+	if err != nil {
+		return fmt.Errorf("connection to s3-csi-daemonset-mounter not yet established, allowing kubelet to retry NodePublishVolume: %w. %s", err, helpMessageForCheckingMounterPodStatus())
+	}
+
+	if err := os.MkdirAll(mountPath, targetDirPerm); err != nil {
+		return fmt.Errorf("failed to create mount directory %q: %w", mountPath, err)
 	}
 
 	mountOpts := mpmounter.MountOptions{
 		ReadOnly:   args.Has(mountpoint.ArgReadOnly),
 		AllowOther: args.Has(mountpoint.ArgAllowOther) || args.Has(mountpoint.ArgAllowRoot),
 	}
-	fd, err := dm.mountSyscallWithDefault(target, mountOpts)
+	fd, err := dm.mountSyscallWithDefault(mountPath, mountOpts)
 	if err != nil {
-		return fmt.Errorf("failed to mount FUSE at %q: %w", target, err)
+		return fmt.Errorf("failed to mount FUSE at %q: %w", mountPath, err)
 	}
+
+	fdClosed := false
+	unmount := true
 	defer func() {
-		if err := mpmounter.CloseFD(fd); err != nil {
-			klog.Errorf("DaemonsetMounter: failed to close /dev/fuse fd %d: %v", fd, err)
+		if !fdClosed {
+			dm.closeFUSEDevFD(fd)
+		}
+		if unmount {
+			if umErr := dm.mount.Unmount(mountPath); umErr != nil {
+				klog.Errorf("Failed to unmount %q during cleanup: %v", mountPath, umErr)
+			}
 		}
 	}()
 
-	// TODO: add --user-agent-prefix for S3 request telemetry
+	args.Remove(mountpoint.ArgReadOnly)
 
-	// Build environment
 	env := envprovider.Environment{}
 	env.Merge(userEnv)
 	env.Merge(envprovider.Default())
 	env.Merge(credsEnv)
 
-	// Move --aws-max-attempts to env if provided
 	if maxAttempts, ok := args.Remove(mountpoint.ArgAWSMaxAttempts); ok {
 		env.Set(envprovider.EnvMaxAttempts, maxAttempts)
 	}
-	// Remove read-only from args since we already passed MS_RDONLY in mount syscall
-	args.Remove(mountpoint.ArgReadOnly)
 
-	// Send mount options to secondary daemonset
 	sockPath := filepath.Join(commDir, MountSockName)
-	klog.V(4).Infof("DaemonsetMounter: sending mount options for volume %s (mount %s) to %s", volumeId, mountId, sockPath)
+	errFilePath := filepath.Join(commDir, GetErrorFileName(mountId))
+	os.Remove(errFilePath)
+
+	klog.V(4).Infof("DaemonsetMounter: sending mount options (mount %s) to %s", mountId, sockPath)
+
 	sendCtx, sendCancel := context.WithTimeout(ctx, sendOptionsTimeout)
 	defer sendCancel()
+
 	err = mountoptions.Send(sendCtx, sockPath, mountoptions.Options{
 		Fd:         fd,
 		BucketName: bucketName,
@@ -280,8 +346,6 @@ func (dm *DaemonsetMounter) mountS3AtTarget(ctx context.Context, target string, 
 		VolumeId:   mountId,
 	})
 	if err != nil {
-		// If send failed due to stale path, signal re-discovery and let Kubelet retry NodePublishVolume.
-		// Send -> dialWithRetry retries ENOENT/ECONNREFUSED, so we only check these errors
 		if errors.Is(err, fs.ErrNotExist) || os.IsPermission(err) || errors.Is(err, context.DeadlineExceeded) {
 			klog.V(4).Infof("DaemonsetMounter: comm dir may be stale, signaling re-discovery")
 			dm.commDir.Store(nil)
@@ -290,52 +354,147 @@ func (dm *DaemonsetMounter) mountS3AtTarget(ctx context.Context, target string, 
 			default:
 			}
 		}
-		return fmt.Errorf("failed to send mount options for volume %s (mount %s): %w. %s", volumeId, mountId, err, helpMessageForGettingMounterLogs())
+		return fmt.Errorf("failed to send mount options (mount %s): %w. %s", mountId, err, helpMessageForGettingMounterLogs())
 	}
 
-	// Wait for mount readiness or error
-	err = dm.waitForMount(ctx, target, mountId, errFilePath)
+	dm.closeFUSEDevFD(fd)
+	fdClosed = true
+
+	err = dm.waitForMount(ctx, mountPath, mountId, errFilePath)
 	if err != nil {
 		return err
 	}
 
+	unmount = false
 	return nil
 }
 
-// waitForMount waits until Mountpoint is serving at target or an error occurs.
-func (dm *DaemonsetMounter) waitForMount(parentCtx context.Context, target, mountId, errFilePath string) error {
-	ctx, cancel := context.WithTimeout(parentCtx, mountReadyTimeout)
-	defer cancel()
+// Unmount unmounts the target path with pod-sharing awareness.
+//
+// Flow:
+//   - Unmounts the bind mount at target
+//   - Decrements refcount in MountMap
+//   - If refcount reaches 0, unmounts the FUSE source and removes the entry
+func (dm *DaemonsetMounter) Unmount(ctx context.Context, target string, credentialCtx credentialprovider.CleanupContext) error {
+	parsedTarget, err := targetpath.Parse(target)
+	if err != nil {
+		return fmt.Errorf("failed to parse target path %q: %w", target, err)
+	}
+	volumeID := parsedTarget.VolumeID // This is the PV name
+	return dm.releaseTarget(target, volumeID)
+}
 
-	mountResultCh := make(chan error, 2)
+// releaseTarget handles the Unmount flow with MountMap refcounting.
+func (dm *DaemonsetMounter) releaseTarget(target string, volumeID string) error {
+	entry := dm.mountMap.Get(volumeID)
+	if entry == nil {
+		// No entry means we never mounted this volume. Return success (idempotency).
+		klog.V(4).Infof("DaemonsetMounter: no mount map entry for volume %s, unmount of %s is a no-op", volumeID, target)
+		return nil
+	}
 
-	// Poll for error file
-	go func() {
-		wait.PollUntilContextCancel(ctx, mountReadyPollInterval, true, func(ctx context.Context) (bool, error) {
-			content, err := os.ReadFile(errFilePath)
-			if err != nil {
-				return false, nil
-			}
-			os.Remove(errFilePath)
-			mountResultCh <- fmt.Errorf("Mountpoint for mount %s failed: %s", mountId, string(content))
-			return true, nil
-		})
-	}()
+	entry.mu.Lock()
+	defer entry.mu.Unlock()
 
-	// Poll for mount readiness
-	go func() {
-		err := wait.PollUntilContextCancel(ctx, mountReadyPollInterval, true, func(ctx context.Context) (bool, error) {
-			isMounted, _ := dm.mount.CheckMountpoint(target)
-			return isMounted, nil
-		})
-		if err != nil {
-			mountResultCh <- fmt.Errorf("timed out waiting for Mountpoint to serve mount %s at %s. %s", mountId, target, helpMessageForGettingMounterLogs())
-		} else {
-			mountResultCh <- nil
+	// Unmount the bind mount at target.
+	if err := dm.mount.Unmount(target); err != nil {
+		if !os.IsNotExist(err) {
+			return fmt.Errorf("failed to unmount bind mount at %q: %w", target, err)
 		}
-	}()
+	}
 
-	return <-mountResultCh
+	// Remove target from entry.
+	for i, t := range entry.Targets {
+		if t == target {
+			entry.Targets = append(entry.Targets[:i], entry.Targets[i+1:]...)
+			entry.RefCount--
+			break
+		}
+	}
+
+	if entry.RefCount == 0 {
+		// Last consumer: clean up credentials first, then unmount the FUSE source.
+		klog.V(4).Infof("DaemonsetMounter: last consumer for volume %s, unmounting source %s", volumeID, entry.SourcePath)
+
+		// Clean up error file and credentials before unmount (ensures retrying cleanup on failure).
+		if dir := dm.commDir.Load(); dir != nil {
+			os.Remove(filepath.Join(*dir, GetErrorFileName(entry.MountID)))
+			if err := dm.cleanupCredentials(*dir, entry.MountID, credentialprovider.CleanupContext{
+				VolumeID:  volumeID,
+				MountKind: credentialprovider.MountKindDaemonset,
+			}); err != nil {
+				klog.Errorf("DaemonsetMounter: failed to cleanup credentials for volume %s: %v", volumeID, err)
+			}
+		}
+
+		// Unmount FUSE source — causes mount-s3 to exit via kernel FUSE teardown.
+		if err := dm.mount.Unmount(entry.SourcePath); err != nil {
+			klog.Errorf("DaemonsetMounter: failed to unmount source %q: %v (best-effort cleanup)", entry.SourcePath, err)
+		}
+		os.Remove(entry.SourcePath)
+
+		// Delete entry from the map — the retry loop in Mount handles the race.
+		dm.mountMap.Delete(volumeID)
+
+		// Remove persisted meta file.
+		RemoveMeta(dm.kubeletPath, volumeID)
+	}
+
+	klog.V(4).Infof("DaemonsetMounter: volume %s unmounted from %s", volumeID, target)
+	return nil
+}
+
+// GetErrorFileName returns the error file name for a given mount ID.
+func GetErrorFileName(mountId string) string {
+	return mountId + MountErrorSuffix
+}
+
+// helpMessageForGettingMounterLogs returns a help message with a command to get mounter logs.
+func helpMessageForGettingMounterLogs() string {
+	return fmt.Sprintf("You can see mounter logs by running: `kubectl logs -n %s -l app=s3-csi-daemonset-mounter`", mounterNamespace)
+}
+
+// helpMessageForCheckingMounterPodStatus returns a help message with a command to check mounter pod status.
+func helpMessageForCheckingMounterPodStatus() string {
+	return fmt.Sprintf("You can check mounter pod status by running: `kubectl get pods -n %s -l app=s3-csi-daemonset-mounter`", mounterNamespace)
+}
+
+// IsMountPoint returns whether the given target is a Mountpoint mount.
+func (dm *DaemonsetMounter) IsMountPoint(target string) (bool, error) {
+	return dm.mount.CheckMountpoint(target)
+}
+
+// IsSourceHealthy checks if the FUSE mount at sourcePath is alive and serving.
+
+func (dm *DaemonsetMounter) IsSourceHealthy(sourcePath string) bool {
+	return dm.mount.IsHealthyMountpoint(sourcePath)
+}
+
+// BindMount performs a bind mount from source to target, creating the target directory if needed.
+func (dm *DaemonsetMounter) BindMount(source, target string) error {
+	if err := os.MkdirAll(target, targetDirPerm); err != nil {
+		return fmt.Errorf("failed to create bind mount target directory %q: %w", target, err)
+	}
+	if err := dm.bindMountSyscallWithDefault(source, target); err != nil {
+		return fmt.Errorf("failed to bind mount %q → %q: %w", source, target, err)
+	}
+	klog.V(4).Infof("DaemonsetMounter: bind mounted %s → %s", source, target)
+	return nil
+}
+
+// bindMountSyscallWithDefault delegates to bindMountSyscall if set, or falls back to dm.mount.BindMount.
+func (dm *DaemonsetMounter) bindMountSyscallWithDefault(source, target string) error {
+	if dm.bindMountSyscall != nil {
+		return dm.bindMountSyscall(source, target)
+	}
+	return dm.mount.BindMount(source, target)
+}
+
+// TODO: refactor closeFUSEDevFD into a shared helper (duplicated in pod_mounter.go)
+func (dm *DaemonsetMounter) closeFUSEDevFD(fd int) {
+	if err := mpmounter.CloseFD(fd); err != nil {
+		klog.V(4).Infof("DaemonsetMounter: failed to close /dev/fuse fd %d: %v", fd, err)
+	}
 }
 
 // mountSyscallWithDefault delegates to mountSyscall if set, or falls back to dm.mount.Mount.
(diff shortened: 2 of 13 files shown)


pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -144,64 +153,138 @@ func (dm *DaemonsetMounter) Mount(ctx context.Context, bucketName string, target
...
+			}
+			entry.RefCount++
+			entry.Targets = append(entry.Targets, target)
+			klog.V(4).Infof("DaemonsetMounter: shared existing mount for volume %s → %s (refcount=%d)",
+				volumeID, target, entry.RefCount)
+			return nil
+		}
+		// Dead source: clean up and fall through to new mount.
+		klog.V(2).Infof("DaemonsetMounter: source %s is dead for volume %s, recovering", entry.SourcePath, volumeID)
+		dm.mount.Unmount(entry.SourcePath)
```
vladem: we should handle the unmount error here (do not reset `initialised`)

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -144,64 +153,138 @@ func (dm *DaemonsetMounter) Mount(ctx context.Context, bucketName string, target
...
+	incomingParams := MountParams{
+		MountOptions:             args.SortedList(),
+		AuthenticationSource:     credentialCtx.AuthenticationSource,
+		ServiceAccountName:       credentialCtx.ServiceAccountName,
+		ServiceAccountEKSRoleARN: credentialCtx.ServiceAccountEKSRoleARN,
+		PodNamespace:             credentialCtx.PodNamespace,
+		FSGroup:                  fsGroup,
+	}
+
+	sourcePath := SourceMountPath(dm.kubeletPath, volumeID)
```
vladem: what is volumeID? pv.name or volumeHandle? can it contain '/' or other dangerous symbols?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -144,64 +153,138 @@ func (dm *DaemonsetMounter) Mount(ctx context.Context, bucketName string, target
...
+		}
+
+		// Check health before sharing.
+		if dm.IsSourceHealthy(entry.SourcePath) {
+			// Healthy source: bind mount to new target.
+			if err := dm.BindMount(entry.SourcePath, target); err != nil {
+				return err
+			}
+			entry.RefCount++
+			entry.Targets = append(entry.Targets, target)
```
vladem: why do we need this field?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -285,6 +407,53 @@ func (dm *DaemonsetMounter) IsMountPoint(target string) (bool, error) {
...
+		klog.V(4).Infof("DaemonsetMounter: source %s health check error: %v", sourcePath, err)
+		return false
+	}
+	if !isMounted {
+		return false
+	}
+
+	// Mount entry exists in the table — now verify the FUSE daemon is alive
+	// by attempting to read the directory. A dead mount returns ENOTCONN.
+	_, err = os.ReadDir(sourcePath)
```
vladem: what if the root mount directory contains millions of files?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -285,6 +407,53 @@ func (dm *DaemonsetMounter) IsMountPoint(target string) (bool, error) {
...
 }
 
+// IsSourceHealthy checks if the FUSE mount at sourcePath is alive and serving.
+// Returns true only if sourcePath is a healthy Mountpoint FUSE mount AND the
+// FUSE daemon is actually responding to I/O.
+//
+// CheckMountpoint alone is insufficient — a dead FUSE mount (after mounter pod crash)
+// stays in the mount table but returns ENOTCONN on any I/O. We must attempt real I/O
+// to distinguish alive from dead mounts.
+func (dm *DaemonsetMounter) IsSourceHealthy(sourcePath string) bool {
```
vladem: shouldnt this logic live in `pkg/mountpoint/mounter/mount.go`?

pkg/driver/node/mounter/mount_map_persistence.go
```diff
@@ -0,0 +1,149 @@
...
+	return filepath.Join(kubeletPath, "plugins", "s3.csi.aws.com", "mnt", volumeID+".meta.json")
+}
+
+// MountMeta is the JSON-serializable structure persisted alongside each source mount.
+// It records the parameters used to create the mount — enabling validation on recovery
+// and for subsequent share requests after driver restart.
+type MountMeta struct {
+	VolumeID                 string   `json:"volumeID"`
+	MountID                  string   `json:"mountID"`
+	SourcePath               string   `json:"sourcePath"`
```
vladem: aren't these 2 fields derivable from `VolumeID`?

pkg/driver/node/mounter/mount_map_persistence.go
```diff
@@ -0,0 +1,149 @@
...
+		PodNamespace:             entry.Params.PodNamespace,
+		FSGroup:                  entry.Params.FSGroup,
+	}
+
+	data, err := json.Marshal(meta)
+	if err != nil {
+		return fmt.Errorf("failed to marshal mount meta for volume %s: %w", entry.VolumeID, err)
+	}
+
+	metaPath := MetaFileName(kubeletPath, entry.VolumeID)
```
vladem, line 54: again: we need to ensure `VolumeID` is escaped before using it as a path component (also account for possible collisions, e.g. `volume/handle` shouldn't be escaped to the same value as `volume-handle`

pkg/driver/node/mounter/mount_map_persistence.go
```diff
@@ -0,0 +1,149 @@
...
+	if err := os.MkdirAll(filepath.Dir(metaPath), 0750); err != nil {
+		return fmt.Errorf("failed to create meta directory: %w", err)
+	}
+
+	tmpPath := metaPath + ".tmp"
+	if err := os.WriteFile(tmpPath, data, 0640); err != nil {
```
vladem: probably harden to `0600`?

pkg/driver/node/mounter/mount_map_persistence.go
```diff
@@ -0,0 +1,149 @@
...
+		return fmt.Errorf("failed to write temp meta file: %w", err)
+	}
+
+	if err := os.Rename(tmpPath, metaPath); err != nil {
```
vladem, line 64: don't think atomic writing is important here (no process may be reading this file while we write), but up to you if you'd like to keep this extra operation

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -144,64 +153,138 @@ func (dm *DaemonsetMounter) Mount(ctx context.Context, bucketName string, target
...
+		klog.V(2).Infof("DaemonsetMounter: source %s is dead for volume %s, recovering", entry.SourcePath, volumeID)
+		dm.mount.Unmount(entry.SourcePath)
+		os.Remove(entry.SourcePath)
+		entry.initialized = false
+	}
+
+	// New mount: FUSE mount at source, then bind to target.
+	mountId := volumeID // In sharing mode, one mount per volume
+	if err := dm.fuseMount(ctx, bucketName, sourcePath, mountId, args, userEnv); err != nil {
+		return err
```
vladem, line 277: garbage collect the inserted un-initialized entry? or better with periodic cleaner to ensure resources assosiated with this entry (mountpoint process, mount itself, creds, cacheDir) are also cleaned up?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -144,64 +153,138 @@ func (dm *DaemonsetMounter) Mount(ctx context.Context, bucketName string, target
...
+	// Populate the entry.
+	entry.SourcePath = sourcePath
+	entry.MountID = mountId
+	entry.Params = incomingParams
+	entry.RefCount = 1
+	entry.Targets = []string{target}
+	entry.initialized = true
+
+	// Persist meta to disk for recovery after driver restart.
+	if err := WriteMeta(dm.kubeletPath, entry); err != nil {
```
vladem: the fact that we persist after we create the associated resources (e.g. the MP process, or mount itself) might cause a leak: created a mount but crashed before persisting => leaked entry? or even an unrecoverable mount (fuseMount will always fail with `already mounted`)?

pkg/driver/node/mounter/mount_map_persistence.go
```diff
@@ -0,0 +1,149 @@
...
+
+// mountInfoEntry represents a single line from /proc/self/mountinfo.
+type mountInfoEntry struct {
+	MountPoint string // the mount point path
+	DeviceID   string // major:minor device ID
+}
+
+// parseMountInfoFromProc reads /proc/self/mountinfo and returns all entries.
+// This is the default implementation used in production (Linux).
+func parseMountInfoFromProc() ([]mountInfoEntry, error) {
```
vladem: don't we have a helper for that in `k8s.io/mount-utils`?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -144,64 +153,138 @@ func (dm *DaemonsetMounter) Mount(ctx context.Context, bucketName string, target
...
+	volumeID string, credentialCtx credentialprovider.ProvideContext, args mountpoint.Args, fsGroup string, userEnv envprovider.Environment) error {
+
+	// Get or create the per-volume entry, then lock it.
+	entry, _ := dm.mountMap.GetOrCreate(volumeID)
+	entry.mu.Lock()
+	defer entry.mu.Unlock()
+
+	// Build mount params for this request — used for validation and stored on first mount.
+	incomingParams := MountParams{
+		MountOptions:             args.SortedList(),
```
vladem, line 198: nit: do mountOptions require normalisation? e.g. `--metadata-ttl=3600`, `--metadata-ttl 3600` and `metadata-ttl=3600` should probably be all the same thing

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -144,64 +153,138 @@ func (dm *DaemonsetMounter) Mount(ctx context.Context, bucketName string, target
...
+		AuthenticationSource:     credentialCtx.AuthenticationSource,
+		ServiceAccountName:       credentialCtx.ServiceAccountName,
+		ServiceAccountEKSRoleARN: credentialCtx.ServiceAccountEKSRoleARN,
```
vladem, line 201: please consider the case when `ServiceAccountEKSRoleARN` annotation on the SA changes, what will happen to the existing source mount? to new mount requests which attempt to share it?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -224,45 +305,86 @@ func (dm *DaemonsetMounter) Mount(ctx context.Context, bucketName string, target
...
+			entry.Targets = append(entry.Targets[:i], entry.Targets[i+1:]...)
+			entry.RefCount--
+			break
+		}
+	}
+
+	if entry.RefCount == 0 {
+		// Last consumer: unmount the FUSE source and reset entry.
+		klog.V(4).Infof("DaemonsetMounter: last consumer for volume %s, unmounting source %s", volumeID, entry.SourcePath)
+		if err := dm.mount.Unmount(entry.SourcePath); err != nil {
```
vladem: consider waiting for MP process to terminate to avoid doubling the resource usage (e.g. what if MP process using 8GB of RAM is taking 30s to terminate for some reason? would we create a brand new process on the next request for the same PV?)

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -224,45 +305,86 @@ func (dm *DaemonsetMounter) Mount(ctx context.Context, bucketName string, target
...
+		}
+		os.Remove(entry.SourcePath)
+
+		// Clean up error file if comm dir is available.
+		if dir := dm.commDir.Load(); dir != nil {
+			os.Remove(filepath.Join(*dir, GetErrorFileName(entry.MountID)))
+		}
+
+		// Reset entry instead of deleting — avoids race with concurrent mount.
+		entry.initialized = false
```
vladem: we never delete the entry from the map, is that sustainable? what if driver runs for a year and millions of entries pile up? e.g. in case of dynamic provisioning, each workload pod will use a brand new PV with a unique `volumeId`

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: There is also a risk to persist the mount metadata before we actually mount. What if it crashed before creating the mount we will wrongly persist data  for the mount which did not happen.

Such scenarios, I think can be handled by the background resource cleanup task which can scan /proc/self/mountinfo for mountpoint-s3 FUSE mounts under our mnt/ directory that aren't tracked in the MountMap, and unmount them?

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: ack

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: I realised this is the volume handle. To keep it consistent with the credential provisioning changes I will also make change to use PV name. PV names follow http://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names naming standards.

pkg/driver/node/mounter/mount_map_persistence.go
rajdchak: yes

pkg/driver/node/mounter/mount_map_persistence.go
rajdchak: VolumeID will now be the PV name

pkg/driver/node/mounter/mount_map_persistence.go
rajdchak: ack

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: They're already normalized. args.SortedList() comes from mountpoint.ParseArgs() https://github.com/awslabs/mountpoint-s3-csi-driver/blob/feature/daemonset-architecture/pkg/mountpoint/args.go#L58

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: dont we have a duplicate check here while launching mountpoint process?
```
// Hold lock across duplicate check and process start to prevent races.
	pm.mu.Lock()
	if _, exists := pm.processes[mountId]; exists {
		pm.mu.Unlock()
		fuseDev.Close()
		return fmt.Errorf("mount %s already has a running process", mountId)
	}
```

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: ack

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: As discussed the existing source mount will keep running with the old arn. New mount requests using the same PV will be rejected. Customers will need to create new PV for successful new mount or will need to delete the older pods before creating the new pods with changed arn.

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: We decrement RefCount if this target is actually in the list. If kubelet sends a duplicate NodeUnpublishVolume for the same target, the second call won't find it in the list (already removed), so RefCount stays unchanged. Without Targets, you'd blindly RefCount-- on every unmount call
The following unmount snippet will not fail for duplicate NodeUnpublishVolume for the same target so we need to do a check before decrementing RefCount.
```
// Unmount the bind mount at target.
	if err := dm.mount.Unmount(target); err != nil {
		if !os.IsNotExist(err) {
			return fmt.Errorf("failed to unmount bind mount at %q: %w", target, err)
		}
	}
```

pkg/driver/node/mounter/mount_map_persistence.go
rajdchak: no

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -290,52 +354,147 @@ func (dm *DaemonsetMounter) mountS3AtTarget(ctx context.Context, target string,
...
+		if !os.IsNotExist(err) {
+			return fmt.Errorf("failed to unmount bind mount at %q: %w", target, err)
+		}
+	}
+
+	// Remove target from entry.
+	for i, t := range entry.Targets {
+		if t == target {
+			entry.Targets = append(entry.Targets[:i], entry.Targets[i+1:]...)
+			entry.RefCount--
```
vladem, line 417: what would happen if target was unmounted by external actor? note, that `node.go` performs `IsMountPoint` check

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: made changes to delete the entry

rajdchak: also sharing is not applicable for DP as we agreed

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: I made the change which will persist metadata before actual mount. But if mount fails that metadata will be deleted. In case node crashes in between persisting data and mount, our rebuild mount map logic is good enough to delete any stale metadata for which mount does not exist.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -95,183 +100,253 @@ type DaemonsetMounter struct {
...
-		if alreadyMounted || mountSuccess {
-			return
-		}
-		if err := dm.cleanupCredentials(commDir, mountId, credentialCtx.ToCleanupCtx()); err != nil {
-			klog.Errorf("DaemonsetMounter: failed to clean up credential directory for mount %s: %v", mountId, err)
+	if isMounted {
+		// Republish: refresh credentials for the existing mount (token rotation).
+		commDir, err := dm.GetCommDir()
+		if err == nil {
+			if _, err := dm.provideCredentials(ctx, commDir, volumeID, &credentialCtx); err != nil {
```
vladem: We need to ensure that mount options match and that we hold the mount lock before providing credentials. Consider a case with 2 pods, using the same PV with `authenticationSource: pod`, but having different service accounts. We'll refuse to share the mount after `ValidateCompatibility`, but before that we'll overwrite the token with one for a different SA.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -95,193 +100,259 @@ type DaemonsetMounter struct {
...
-	return nil
-}
-
-// Unmount unmounts the FUSE filesystem at target.
-// This causes the Mountpoint process in the secondary daemonset to exit.
-func (dm *DaemonsetMounter) Unmount(ctx context.Context, target string, cleanupCtx credentialprovider.CleanupContext) error {
-	// cleanup the error file and credentials
+	// Provision credentials under the lock to avoid races between concurrent
+	// NodePublishVolume calls for the same PV (all share one credential file).
 	commDir, err := dm.GetCommDir()
```
vladem: Consider making `commDir` a property of the `MountEntry`. All mount resources (MP process, credentials, cache dir) are associated with a specific mounter (fixed `commDir`). If `commDir`  changes we don't want to provide credentials to a new mounter, since there is no MP process running there. We also want to attempt cleanup in the same `commDir` where we created resources. 

If mounter dies and `commDir` becomes invalid, we'll re-initialise the mount (after `IsSourceHealthy` returns false) with a fresh `commDir`.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -95,193 +100,259 @@ type DaemonsetMounter struct {
...
-// (including the FUSE fd) to the secondary daemonset. On success the mount remains
-// at target; on failure any partial mount is cleaned up.
-func (dm *DaemonsetMounter) mountS3AtTarget(ctx context.Context, target string, bucketName string,
-	args mountpoint.Args, mountId string, volumeId string, commDir string,
-	userEnv envprovider.Environment, credsEnv envprovider.Environment) error {
-	// Remove old error file if exists
-	errFilePath := filepath.Join(commDir, GetErrorFileName(mountId))
-	if err := removeIfExists(errFilePath); err != nil {
-		return fmt.Errorf("failed to remove stale error file %q: %w", errFilePath, err)
+	commDir, err := dm.GetCommDir()
```
vladem: consider calling `GetCommDir` at most once per NodePublishVolume, we don't want to launch MP inside a mounter which doesn't have creds (could happen if mounter pod dies between `provideCredentials(commDir1)` and  `fuseMount(commDir2)`)

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -95,193 +100,259 @@ type DaemonsetMounter struct {
...
+	// fails, we clean up the meta file. An orphan meta file (crash after meta-write but
+	// before fuseMount) is harmless — RebuildMountMap detects no matching mount in
+	// /proc/self/mountinfo and deletes the stale meta file.
+	entry.Params = incomingParams
+	if err := WriteMeta(dm.kubeletPath, entry); err != nil {
+		klog.Errorf("DaemonsetMounter: failed to write meta for volume %s: %v (non-fatal)", volumeID, err)
+	}
+
+	if err := dm.fuseMount(ctx, bucketName, sourcePath, volumeID, args, userEnv, credsEnv); err != nil {
+		dm.cleanupCredentials(commDir, volumeID, credentialCtx.ToCleanupCtx())
```
vladem: Cleanup errors are mostly ignored in this function (`cleanupCredentials`, `os.Remove`), I recommend that we at least log them for debuggability. Probably we also want to avoid `RemoveMeta` unless cleanup was confirmed successful.

Would it be possible to have a single idempotent `func (dm *DaemonsetMounter) cleanupMount(entry *MountEntry) error`? This would return success for the given entry iff: 

1. No credential files exist
2. No error file exist
3. `sourcePath` is not mounted
4. `sourcePath` doesn't exist
5. [in future] no cache dir exist
6. [in future] no MP process running with assosiated UID exist

pkg/driver/node/mounter/mount_map_persistence.go
```diff
@@ -0,0 +1,145 @@
...
+	if err := os.MkdirAll(filepath.Dir(metaPath), 0750); err != nil {
+		return fmt.Errorf("failed to create meta directory: %w", err)
+	}
+
+	tmpPath := metaPath + ".tmp"
+	if err := os.WriteFile(tmpPath, data, 0600); err != nil {
+		return fmt.Errorf("failed to write temp meta file: %w", err)
+	}
+
+	if err := os.Rename(tmpPath, metaPath); err != nil {
```
vladem, line 64: What if `metaPath` already exists for some reason? Its a high risk if we overwrite with a different SA. Consider using `RENAME_NOREPLACE`. (With sqlite we'd probably get it for free)

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -95,193 +100,259 @@ type DaemonsetMounter struct {
...
-		return fmt.Errorf("failed to unmount %q: %w", target, err)
+	// New mount: FUSE mount at source, then bind to target.
+
+	// Persist meta to disk BEFORE mounting so that RebuildMountMap can find this volume
+	// if the driver crashes between fuseMount and the end of this function. If the mount
+	// fails, we clean up the meta file. An orphan meta file (crash after meta-write but
+	// before fuseMount) is harmless — RebuildMountMap detects no matching mount in
+	// /proc/self/mountinfo and deletes the stale meta file.
+	entry.Params = incomingParams
+	if err := WriteMeta(dm.kubeletPath, entry); err != nil {
```
vladem: Consider `WriteMeta` before providing credentials. Meta file exists => credentials file might also exist. Never the reverse.

pkg/driver/node/mounter/daemonset_mounter_test.go
```diff
@@ -194,28 +198,13 @@ func TestDaemonsetMounter(t *testing.T) {
 				BucketName: testCtx.bucketName,
 				Args:       []string{"--prefix=data/"},
 				Env:        env.List(),
-				VolumeId:   mustGetMountId(t, testCtx.targetPath),
+				VolumeId:   testCtx.volumeID,
 			}, got)
 		})
 
-		t.Run("Does not duplicate mounts if target is already mounted and refreshes credentials", func(t *testing.T) {
```
vladem, line 201: This file wasn't merged properly. All credentials-related tests were reverted.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -95,193 +100,259 @@ type DaemonsetMounter struct {
...
+		}
+		// Dead source: clean up and fall through to new mount.
+		klog.V(2).Infof("DaemonsetMounter: source %s is dead for volume %s, recovering", entry.SourcePath, volumeID)
+		if err := dm.mount.Unmount(entry.SourcePath); err != nil {
+			if !os.IsNotExist(err) {
+				return fmt.Errorf("failed to unmount dead source %q for volume %s, will retry: %w", entry.SourcePath, volumeID, err)
+			}
+		}
+		os.Remove(entry.SourcePath)
+		entry.initialized = false
```
vladem: consider renaming this flag to `mounted`, `mp_running` or similar

pkg/driver/node/mounter/mount_map_persistence.go
vladem: consider using this https://pkg.go.dev/k8s.io/mount-utils#ParseMountInfo

pkg/driver/node/mounter/daemonset_mounter.go
vladem: Consider providing better diagnostics to customer for that case in future. Currently [error](https://github.com/rajdchak/rajdchak-fork-mountpoint-s3-csi-driver/blob/podsharing/cmd/aws-s3-csi-daemonset-mounter/process_manager.go#L76) will only be visible in mounter log. We could propagate it to `NodePublishVolume` response via the `.error` file

pkg/driver/node/mounter/daemonset_mounter.go
vladem: Also now, since we store `ServiceAccountEKSRoleARN`, may be we can avoid [fetching](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/feature/daemonset-architecture/pkg/driver/node/credentialprovider/provider_pod.go#L152) from API each time?

pkg/driver/node/mounter/daemonset_mounter.go
jet-tong: Reproduced 1 case or orphaned source directory which I think this advice would help:
- kubelet calls NPV for a PV and driver creates source dir and FUSE syscall mount
- [mounter pod restarted in between]
- driver tries to send fd to mounter pod pocket, retries for 15s and timeouts and fuseMount fails
- Found orphaned `mnt/<pv-name>` source directory

(unlikely case and was only reproduced when deleting mounter pod while creating 50 volumes or so)

pkg/driver/node/mounter/mount_map_persistence.go
vladem: Actually, after thinking about this more, probably its the right behaviour to replace the stale meta file on disk. We rely on the in-memory map to be an accurate representation of volume's metadata. Data on disk may become out-of-sync if entry was removed from in-memory map, but disk operation failed. Entry is removed from in-memory map strictly after confirming `unmount(source)` succeeded. This makes it safe to create the mount (possibly with different SA) and overwrite on-disk `.meta`.

rajdchak: The overwrite is intentional for the dead-source recovery case (stale meta from previous mount). ValidateCompatibility prevents a different SA from reaching WriteMeta — incompatible requests are rejected before we get here. The entry lock serializes all writes for the same PV

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: I made a change in node.go to always call unmount in case of daemonset mode.

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: nice catch, corrected this. Even corrected for the republish/refresh path where also we should check the params still match before refreshing the creds.

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: made this change

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: ack

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: made this change.

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: ack

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: ack, we can improve observability later. I will check if we have a task for it or will create one.

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: Also now, since we store ServiceAccountEKSRoleARN, may be we can avoid [fetching](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/feature/daemonset-architecture/pkg/driver/node/credentialprovider/provider_pod.go#L152) from API each time? - will do this in a follow up PR

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -290,52 +359,178 @@ func (dm *DaemonsetMounter) mountS3AtTarget(ctx context.Context, target string,
...
+			errs = append(errs, err)
+		}
+	}
+
+	//[future] Clean cache dir
+	// [future] Verify no MP process running
+
+	if len(errs) > 0 {
+		return fmt.Errorf("incomplete cleanup for volume %s (%d errors)", entry.VolumeID, len(errs))
+	}
```
jet-tong, line 481: This reports only error count, shall we surface the errors somehow?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -290,52 +359,178 @@ func (dm *DaemonsetMounter) mountS3AtTarget(ctx context.Context, target string,
...
+
+// bindMountSyscallWithDefault delegates to bindMountSyscall if set, or falls back to dm.mount.BindMount.
+func (dm *DaemonsetMounter) bindMountSyscallWithDefault(source, target string) error {
+	if dm.bindMountSyscall != nil {
+		return dm.bindMountSyscall(source, target)
+	}
+	return dm.mount.BindMount(source, target)
+}
+
+// TODO: refactor closeFUSEDevFD into a shared helper (duplicated in pod_mounter.go)
```
jet-tong: qq: I also sprayed some TODOs to refactor in earlier PRs; do we need a plan to refactor to share with pod_mounter, or should we just keep them separate since we'll start maintaining them separately?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -95,193 +101,256 @@ type DaemonsetMounter struct {
...
+		return fmt.Errorf("connection to s3-csi-daemonset-mounter not yet established, allowing kubelet to retry NodePublishVolume: %w. %s", err, helpMessageForCheckingMounterPodStatus())
 	}
 
-	alreadyMounted, err := dm.IsMountPoint(target)
-	if err != nil && !errors.Is(err, fs.ErrNotExist) {
-		return fmt.Errorf("failed to check if target %q is a mount point (mount target is possibly"+
-			" corrupted, manual pod re-creation %s might be required for mount recovery): %w",
-			target, credentialCtx.WorkloadPodID, err)
-	}
+	// All paths (republish, share, new mount) go through mountOrShareSource
```
jet-tong, line 173: If all paths go through the same function that's not reused anyways, could we put it in the same function for readability?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -290,52 +359,178 @@ func (dm *DaemonsetMounter) mountS3AtTarget(ctx context.Context, target string,
...
+			klog.Errorf("DaemonsetMounter: unmount source %s: %v", entry.SourcePath, err)
+			errs = append(errs, err)
+		}
+		if err := os.Remove(entry.SourcePath); err != nil && !os.IsNotExist(err) {
+			klog.Errorf("DaemonsetMounter: remove source dir %s: %v", entry.SourcePath, err)
+			errs = append(errs, err)
+		}
+	}
+
+	//[future] Clean cache dir
```
jet-tong: np: missing space, and perhaps we can use TODO instead so we scan for it before merging back to main?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -290,52 +359,178 @@ func (dm *DaemonsetMounter) mountS3AtTarget(ctx context.Context, target string,
...
+	return fmt.Sprintf("You can check mounter pod status by running: `kubectl get pods -n %s -l app=s3-csi-daemonset-mounter`", mounterNamespace)
+}
+
+// IsMountPoint returns whether the given target is a Mountpoint mount.
+func (dm *DaemonsetMounter) IsMountPoint(target string) (bool, error) {
+	return dm.mount.CheckMountpoint(target)
+}
+
+// IsSourceHealthy checks if the FUSE mount at sourcePath is alive and serving.
+
```
jet-tong: np: extra line

pkg/mountpoint/mounter/mount.go
```diff
@@ -119,6 +121,30 @@ func (m *Mounter) FindReferencesToMountpoint(target Target) ([]string, error) {
 	return m.mount.GetMountRefs(target)
 }
 
+// IsHealthyMountpoint checks whether `target` is both a valid Mountpoint mount AND the FUSE daemon is alive.
+// A dead FUSE mount (after mounter pod crash) stays in the mount table but returns ENOTCONN on any I/O.
+func (m *Mounter) IsHealthyMountpoint(target Target) bool {
```
jet-tong, line 131: np: Would IsMountpointHealthy be better name? (c.f. IsSourceHealthy which calls this function)

pkg/mountpoint/mounter/mount.go
```diff
@@ -119,6 +121,30 @@ func (m *Mounter) FindReferencesToMountpoint(target Target) ([]string, error) {
...
+	isMounted, err := m.CheckMountpoint(target)
```
jet-tong, line 132: qq: Noted CheckMountpoint above says 
```
// Some notable errors that requires callers to perform some operations are:
//   - If `errors.Is(err, fs.ErrNotExist)` - it means the `target` does not exists, and the caller should create the target folder
//   - If `mounter.IsMountpointCorrupted(err)` - it means the `target` is corrupted, and the caller should `Unmount` and `Mount` the file system
```
I think daemonset_mounter.go handles fs.ErrNotExist, but should we handle IsMountpointCorrupted somewhere? (I don't know which actual case this would handle, perhaps mounter restart? Or perhaps it's not a problem?)

pkg/driver/node/mounter/mount_map_persistence_test.go
```diff
@@ -0,0 +1,567 @@
...
+// It sets kubeletPath and mountInfoProvider — no clientset/mount needed for RebuildMountMap.
+func newTestDMWithMountInfo(kubeletPath string, provider mountInfoProviderFunc) *DaemonsetMounter {
+	return &DaemonsetMounter{
+		kubeletPath:       kubeletPath,
+		mountInfoProvider: provider,
+		mountMap:          NewMountMap(),
+	}
+}
+
+func TestWriteMeta_CreatesFile(t *testing.T) {
```
jet-tong, line 59: style nit (might require huge refactor if taken): I think the project likes nested t.Run more than individual func for each test, and I also thought e.g. grouping TestWriteMeta as 1 func and t.Run for each sub-tests would be good to match style (rather than underscore in `TestWriteMeta_CreatesFile`). Do you think we switch to that style or keep current? 

(This is for full mount_map_test.go and mount_map_persistence_test.go)

pkg/driver/driver.go
```diff
@@ -125,7 +125,15 @@ func NewDriver(endpoint string, mpVersion string, nodeID string) (*Driver, error
 		klog.Info("Using daemonset mounter mode")
 		klog.Info("Note: s3-csi-daemonset-mounter uses OnDelete update strategy - helm upgrade will not restart/upgrade mounter pods automatically")
 
-		dm := mounter.NewDaemonsetMounter(clientset, nodeID, mpmounter.New(), credProvider, nil)
+		dm := mounter.NewDaemonsetMounter(clientset, nodeID, mpmounter.New(), credProvider, nil, nil, nil)
+
+		// Rebuild mount map from persisted .meta.json files + kernel mount table.
+		// Fatal on failure: operating with an empty map while FUSE mounts exist
+		// would cause duplicate mounts or incorrect unmounts.
+		if err := dm.RebuildMountMap(); err != nil {
```
jet-tong, line 133: Like dm.DiscoverCommDir, should we add a wait.ExponentialBackoffWithContext wrapper for RebuildMountMap so it can retry any potentially flaky errors (not sure if `mountutils.ParseMountInfo("/proc/self/mountinfo")` errors are common, for example) (also need to make sure idempotent RebuildMountMap if adding retry)

Or is failing on any error more correct to surface MountMap problems that require a node pod restart?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -290,52 +359,178 @@ func (dm *DaemonsetMounter) mountS3AtTarget(ctx context.Context, target string,
...
 }
 
-// waitForMount waits until Mountpoint is serving at target or an error occurs.
-func (dm *DaemonsetMounter) waitForMount(parentCtx context.Context, target, mountId, errFilePath string) error {
-	ctx, cancel := context.WithTimeout(parentCtx, mountReadyTimeout)
-	defer cancel()
+// Unmount unmounts the target path with pod-sharing awareness.
+//
+// Flow:
+//   - Unmounts the bind mount at target
```
jet-tong: style np: can use 1. 2. 3. instead of dashes like above
```
// Flow:
//  1. Acquire per-volume lock via MountMap (serializes concurrent NodePublishVolume for same PV)
```

pkg/driver/node/mounter/daemonset_mounter.go
jet-tong: That gave me idea on potential bug in reconciler.go, which doesn't do this normalization I think? In [`crdv2.FieldMountOptions:         strings.Join(pv.Spec.MountOptions, ","),`](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/v2.7.0/cmd/aws-s3-csi-controller/csicontroller/reconciler.go#L370) - could be related to duplicate S3PA (when mount options are editted for example).

pkg/mountpoint/mounter/mount.go
rajdchak: we are calling this  IsHealthyMountpoint from here 
```
if !dm.IsSourceHealthy(entry.SourcePath) {
			// Dead source — tear down and allow fresh mount with any params.
			klog.V(2).Infof("DaemonsetMounter: source %s is dead for volume %s, resetting entry for fresh mount", entry.SourcePath, volumeID)
			dm.mount.Unmount(entry.SourcePath)
			os.Remove(entry.SourcePath)
			entry.sourceMounted = false
		} else {
```

if it is corrupted we would unmount the source mount and create a fresh mount

pkg/driver/driver.go
rajdchak: Failing fast I think is correct here. /proc/self/mountinfo reads are in-memory - if they fail, it's a fundamentally broken environment where retry won't help. Meta file reads on hostPath are equally unlikely to be transiently broken.

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: Keeping them separate is better I think. Pod mounter will be in maintenance mode & gradually be deprecated — we won't be actively changing it. Daemonset mounter is evolving fast. Sharing code between them means V3 changes could accidentally break V2. We can remove the TODO.

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: Each error is already logged individually via klog.Errorf right where it happens.

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: yes will change to TODO

pkg/driver/node/mounter/daemonset_mounter.go
jet-tong: That's true I guess that's good enough, since we only log the errors at callsites anyways I realized (I was hoping we could bubble/surface error up, then we print the errors in one go).

pkg/mountpoint/mounter/mount.go
jet-tong: Ah you're right, errors handled included this one anyways and it did unmount which was what the comment said. Thanks.

pkg/driver/driver.go
jet-tong: Gotcha, meta file read not being transcient makes sense.

pkg/driver/node/mounter/daemonset_mounter.go
vladem: I expect we'll also need to handle it in periodic cleanup (list source mounts in `/mnt`, check which have no bind mounts, cleanup). Handling daemonset mounter in a special way in `node.go` is a bit clumsy: ideally caller should be agnostic to implementation of the mounter interface, but I have no strong objection.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -95,193 +101,256 @@ type DaemonsetMounter struct {
...
-		return fmt.Errorf("failed to find s3-csi-daemonset-mounter pod: %w. %s", err, helpMessageForCheckingMounterPodStatus())
-	}
-	mountId, err := GetMountId(target)
-	if err != nil {
+	if err := dm.fuseMount(ctx, bucketName, entry.SourcePath, volumeID, commDir, args, userEnv, credsEnv); err != nil {
+		if cleanErr := dm.cleanupMount(entry, credentialCtx.ToCleanupCtx()); cleanErr != nil {
+			klog.Errorf("DaemonsetMounter: cleanup after fuseMount failure for volume %s: %v", volumeID, cleanErr)
+		}
+		dm.mountMap.Delete(volumeID)
+		RemoveMeta(dm.kubeletPath, volumeID)
```
vladem, line 276: Do we want to delete the mount from the map if cleanup failed? I am inclined towards keeping it so that we don't leak resources, but that might also mean slower mount recovery. Interested in your thoughts here. Currently it seems that we don't have a firm opinion on this: here we forget the mount even if cleanup failed, in `Unmount` we do the opposite: `// Don't remove meta or map entry — leave for retry/background cleanup.`.

pkg/driver/node/mounter/daemonset_mounter_test.go
```diff
@@ -47,10 +47,20 @@ type dmTestCtx struct {
...
 
-func setupDM(t *testing.T, credProvider ...credentialprovider.ProviderInterface) *dmTestCtx {
+// targetPath returns a valid kubelet-style target path that targetpath.Parse can parse.
+// Format: <kubeletPath>/pods/<podUID>/volumes/kubernetes.io~csi/<volumeID>/mount
+func (testCtx *dmTestCtx) targetPath(podUID string) string {
+	return filepath.Join(testCtx.kubeletPath, "pods", podUID, "volumes", "kubernetes.io~csi", testCtx.volumeID, "mount")
+}
+
+// targetPathWithVolume returns a valid kubelet-style target path for a specific volumeID.
+func (testCtx *dmTestCtx) targetPathWithVolume(podUID, volumeID string) string {
```
vladem: unused

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: yeah, I didn't want to make any changes for the pod mounter mode so chose to do this specifically for daemonset mode.

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: I was thinking having the split makes it clear what happens outside the lock vs inside.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -95,193 +101,256 @@ type DaemonsetMounter struct {
...
+		FSGroup:                  fsGroup,
+	}
+
+	// If source is mounted, check health first. Dead source = tear down and treat as fresh.
+	// Only enforce compatibility on a healthy (living) source.
+	if entry.sourceMounted {
+		if !dm.IsSourceHealthy(entry.SourcePath) {
+			// Dead source — tear down and allow fresh mount with any params.
+			klog.V(2).Infof("DaemonsetMounter: source %s is dead for volume %s, resetting entry for fresh mount", entry.SourcePath, volumeID)
+			dm.mount.Unmount(entry.SourcePath)
```
vladem: We're going to re-mount with potentially different credentials, should we ensure the full cleanup here via `dm.cleanupMount`?

Imagine the case where PV was using driver level creds and will switch to pod-level, after re-mount there will be 2 token files in `mountCredDir` for this mount.

My preference here would be to disallow switching `MountParams` when handling a dead source. I can't justify it yet, but I'd prefer allowing non-compatible `MountParams` only after previous mount was explicitly shut down after reference count reaching zero.

rajdchak: My thinking was a source is unhealthy only when the older mounter pod has crashed, which means the comm dir is gone.

rajdchak: My preference here would be to disallow switching MountParams when handling a dead source. I can't justify it yet, but I'd prefer allowing non-compatible MountParams only after previous mount was explicitly shut down after reference count reaching zero. 
I think that will be a bad customer experience.
That said, your concern about stale credential files is valid. If I add a full cleanupMount call (including cleanupCredentials) in the dead-source teardown path before resetting sourceMounted = false. Does that address your concern?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -95,193 +101,256 @@ type DaemonsetMounter struct {
...
+
+	fdClosed := false
+	unmount := true
 	defer func() {
-		if err := mpmounter.CloseFD(fd); err != nil {
-			klog.Errorf("DaemonsetMounter: failed to close /dev/fuse fd %d: %v", fd, err)
+		if !fdClosed {
+			dm.closeFUSEDevFD(fd)
+		}
+		if unmount {
```
vladem: redundant given that we call `dm.cleanupMount` in the caller on error from this method

pkg/driver/node/mounter/mount_map.go
```diff
@@ -0,0 +1,155 @@
...
+				existing.ServiceAccountName, incoming.ServiceAccountName)
+		}
+
+		if existing.ServiceAccountEKSRoleARN != incoming.ServiceAccountEKSRoleARN {
+			return fmt.Errorf("serviceAccountEKSRoleARN mismatch: existing=%q, incoming=%q",
+				existing.ServiceAccountEKSRoleARN, incoming.ServiceAccountEKSRoleARN)
+		}
+	}
+
+	if existing.PodNamespace != incoming.PodNamespace {
```
vladem: Here we're saying that 2 pods from different namespaces can never share a mount, even a mount with driver-level creds. I think its a breaking change from V2. Probably you'd want to move this check under  `if existing.AuthenticationSource == "pod"`.

pkg/driver/node/mounter/mount_map_persistence.go
```diff
@@ -0,0 +1,125 @@
...
+	"path/filepath"
+
+	"k8s.io/klog/v2"
+	mountutils "k8s.io/mount-utils"
+)
+
+// MetaFileName returns the path to the .meta.json file for a given volume.
+// The file lives alongside the FUSE source mount directory.
+func MetaFileName(kubeletPath, volumeID string) string {
+	return filepath.Join(kubeletPath, "plugins", "s3.csi.aws.com", "mnt", volumeID+".meta.json")
```
vladem: Consider using a separate directory (not `mnt/`) for meta files. The concerning case is when pv is named `foo.meta.json` (includes our reserved extension). We'll fail in `RebuildMountMap` with such mount.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -502,6 +665,98 @@ func (dm *DaemonsetMounter) GetCommDir() (string, error) {
...
+
+		// Derive SourcePath from VolumeID (not persisted, always computable)
+		sourcePath := SourceMountPath(dm.kubeletPath, meta.VolumeID)
+
+		// Find the source mount in mountinfo
+		sourceMI := findMountByPath(mountInfos, sourcePath)
+		if sourceMI == nil {
+			// Source mount is gone — clean up the meta file
+			klog.V(2).Infof("MountMap: source mount %s for volume %s not found in mount table, cleaning up", sourcePath, meta.VolumeID)
+			os.Remove(metaPath)
```
vladem, line 777: might leave credentials behind, consider calling `dm.cleanupMount`

pkg/driver/node/mounter/mount_map_persistence_test.go
rajdchak: Agreed the nested t.Run style is cleaner and matches the rest of the project. I'd prefer to do it as a follow-up PR to keep this diff focused on the functional changes. Will track it

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -502,6 +659,109 @@ func (dm *DaemonsetMounter) GetCommDir() (string, error) {
 	return *dir, nil
 }
 
+// KubeletPath returns the kubelet path used by this mounter.
+func (dm *DaemonsetMounter) KubeletPath() string {
```
jet-tong: This might be unused?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -502,6 +659,109 @@ func (dm *DaemonsetMounter) GetCommDir() (string, error) {
...
+	return parseMountInfoFromProc()
+}
+
+// RebuildMountMap reconstructs the MountMap from disk on driver startup.
+// It scans the meta directory for .meta.json files, verifies each source mount
+// is still alive via /proc/self/mountinfo, and counts bind mounts (targets) by
+// matching device IDs.
+//
+// Algorithm:
+//  1. List all .meta.json files in the plugins/s3.csi.aws.com/mnt/ directory
```
jet-tong: I think it's now in meta directory (much cleaner now!)

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -95,193 +101,259 @@ type DaemonsetMounter struct {
...
 
-	// Provision credentials before the alreadyMounted early return so republish refreshes them
-	credsEnv, err := dm.provideCredentials(ctx, commDir, mountId, &credentialCtx)
+	// Provision credentials under the lock. We use the fresh commDir (from GetCommDir at the
+	// top of Mount) for all publish operations. When source is mounted (source healthy),
+	// commDir == entry.CommDir because the mounter pod hasn't restarted. When source not mounted
+	// (fresh mount or dead-source recovery), entry.CommDir is empty/stale, so the fresh
+	// commDir is the only valid path. Cleanup (releaseTarget) uses entry.CommDir to clean
+	// resources in the same location where they were originally created.
+	credsEnv, err := dm.provideCredentials(ctx, commDir, volumeID, &credentialCtx)
```
vladem: `commDir` -> `entry.CommDir`

pkg/driver/node/mounter/daemonset_mounter.go
vladem: nit: for completeness, I'd store entry (with `meta.CommDir`) in the map in case of a cleanup failure. This way we'll ensure the cleanup in the `commDir` where credentials were stored (no next `NodePublish` / periodic cleanup). 

In practice this doesn't matter much (`commDir` only changes when a new mounter pod is created; new mounter pod is only created after the old one was deleted; deletion of old mounter pod wipes its emptyDir volume which we recorded in entry's `commDir`). 

Consider as a defence in depth measure: don't rely on external components wiping emptyDir, driver itself ensures cleanup of files which it had created.

pkg/driver/node/node.go
```diff
@@ -239,8 +239,13 @@ func (ns *S3NodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUn
 		return nil, status.Errorf(codes.Internal, "Could not unmount %q: %v", targetContainer, err)
 	}
 	if !mounted {
-		klog.V(4).Infof("NodeUnpublishVolume: target path %s not mounted, skipping unmount", targetContainer)
-		return &csi.NodeUnpublishVolumeResponse{}, nil
+		// For daemonset mounter, always call Unmount for bookkeeping (MountMap refcount).
+		// For other mounters, skip if not mounted (original behavior).
+		if _, isDaemonset := ns.Mounter.(*mounter.DaemonsetMounter); !isDaemonset {
```
vladem, line 244: Usually `unmount` is idempotent (`not mounted` error treated as success) and there is no harm in unmounting a path which is not a mount. With very old versions of util-linux `unmount` might not be idempotent (return an error when not mounted) (see [mount-utils code](https://github.com/kubernetes/mount-utils/blob/master/mount_linux.go#L918)). My suggestion would be to lookup the path in the mount table and skip unmount if path is not a mount (probably in `DaemonsetMounter` only, with a wrapper method similar to `mountSyscallWithDefault`).

vladem (commented):
few small comments, otherwise LGTM

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -95,193 +101,259 @@ type DaemonsetMounter struct {
...
+		ServiceAccountName:       credentialCtx.ServiceAccountName,
+		ServiceAccountEKSRoleARN: credentialCtx.ServiceAccountEKSRoleARN,
+		PodNamespace:             credentialCtx.PodNamespace,
+		FSGroup:                  fsGroup,
+	}
+
+	// If source is mounted, check health first. Dead source = mark not mounted so we go
+	// through the fresh-mount path below. Only enforce compatibility on a healthy (living) source.
+	if entry.sourceMounted {
+		if !dm.IsSourceHealthy(entry.SourcePath) {
```
vladem: A risk worth investigating -- `IsSourceHealthy` / `IsMountPoint` behaviour when mount happened but fuse daemon (MP) never started serving on it. My suspicion is that `stat` and `Readdirnames(1)` will block forever and goroutine processing the `NodePublishVolume` won't return. This may lock the mount forever. A fix could be to run the check in a goroutine, canceling it according to the `ctx context.Context` provided in the request. In my opinion this deserves a unit/e2e test.

pkg/mountpoint/mounter/mount.go
```diff
@@ -119,6 +121,30 @@ func (m *Mounter) FindReferencesToMountpoint(target Target) ([]string, error) {
...
+
+	// Mount entry exists in the table — verify the FUSE daemon is alive
+	// by attempting to read the directory. A dead mount returns ENOTCONN.
+	// We use Open+Readdirnames(1) to issue a single getdents syscall,
+	// avoiding enumeration of potentially millions of entries.
+	f, err := os.Open(target)
+	if err != nil {
+		return false
+	}
+	_, err = f.Readdirnames(1)
```
vladem: Consider removing `Readdirnames` call. This relies on S3 being available for `IsHealthyMountpoint` to succeed. We unmount the source, invalidating all bind mounts, when health check fails. In face of a network issue, `Readdirnames` will fail after a certain amount of retries on MP side. Corrupting all bind mounts probably is not desirable in this case.  `f, err := os.Open(target)` should be a valid indicator of fuse daemon health: opening the root doesn't require any metadata from S3, but still ensures a call to MP is made.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -290,52 +362,174 @@ func (dm *DaemonsetMounter) mountS3AtTarget(ctx context.Context, target string,
...
+		errFile := filepath.Join(entry.CommDir, GetErrorFileName(entry.VolumeID))
+		if err := os.Remove(errFile); err != nil && !os.IsNotExist(err) {
+			klog.Errorf("DaemonsetMounter: cleanup error file %s: %v", errFile, err)
+			errs = append(errs, err)
+		}
+	}
+
+	// Unmount FUSE source (causes mount-s3 to exit via kernel FUSE teardown)
+	if entry.SourcePath != "" {
+		if err := dm.mount.Unmount(entry.SourcePath); err != nil && !os.IsNotExist(err) {
```
jet-tong: Do we need CheckMountpoint here before calling Unmount?

Repro from deleting mounter pod when a workload is still active. 

```
DaemonsetMounter: unmount source /var/lib/kubelet/plugins/s3.csi.aws.com/mnt/s3-pv: unmount failed: exit status 32 Unmounting arguments: /var/lib/kubelet/plugins/s3.csi.aws.com/mnt/s3-pv Output: umount: /var/lib/kubelet/plugins/s3.csi.aws.com/mnt/s3-pv: no mount point specified.
...
GRPC error: rpc error: code = Internal desc = Could not mount "xxxx" at "/var/lib/kubelet/pods/xxxxxx/volumes/kubernetes.io~csi/s3-pv/mount": failed to clean up stale resources for volume s3-pv, cannot proceed with fresh mount: incomplete cleanup for volume s3-pv (1 errors)
```

and without resource cleanup in the background (expected currently) when I try to delete / re-apply:
```
Internal desc = Could not mount "xxxx" at "/var/lib/kubelet/pods/xxxx/volumes/kubernetes.io~csi/s3-pv/mount": failed to clean up stale resources for volume s3-pv, cannot proceed with fresh mount: incomplete cleanup for volume s3-pv (1 errors)
```
(as the meta json is not cleaned up)

Btw, is it expected that currently without the cleanup job for now, deleting mounter pod will leave multiple dangling resources and will add noise to the s3-csi-node logs?

pkg/mountpoint/mounter/mount.go
jet-tong: On a side note, CheckMountpoint in mount.go says `Healthy Mountpoint mount, it returns "true, nil"` which is not quite true, which this function addresses, it's great that we have IsHealthyMountpoint I think

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: yeah added a check for is it a mountpoint here as well

rajdchak: Internal desc = Could not mount "xxxx" at "/var/lib/kubelet/pods/xxxx/volumes/kubernetes.io~csi/s3-pv/mount": failed to clean up stale resources for volume s3-pv, cannot proceed with fresh mount: incomplete cleanup for volume s3-pv (1 errors)
what was the reason for the incomplete clean up, did you check the other logs? was it incomplete creds, error file or source cleanup?

pkg/driver/node/mounter/daemonset_mounter_test.go
```diff
@@ -19,6 +19,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/fake"
 	"k8s.io/mount-utils"
+	mountutils "k8s.io/mount-utils"
```
jet-tong: np: repeated mount-utils import

pkg/driver/node/mounter/daemonset_mounter.go
jet-tong: Thanks for the fix, the large error block in this case is gone now. 

Now perhaps it's a different problem - same 'restart mounter pod when workload pod is active' case gave me:
`DaemonsetMounter: target /var/lib/kubelet/pods/xxxx/volumes/kubernetes.io~csi/s3-pv/mount is already mounted, credentials refreshed`, but Mountpoint isn't actually working `ls: unknown io error: '/data/', 'Os { code: 107, kind: NotConnected, message: "Transport endpoint is not connected" }'`, since bind mount target is dead (not found in `isMounted, err := dm.IsMountPoint(target)`). 

Maybe we can add a `dm.IsTargetHealthy` parallel to IsSourceHealthy? (And I assume background cleanup process will take care of the problem, while we surface errors correctly...?)

rajdchak: we cannot revive dead targets if mounter pod crashed (which have we accepted as of now), so there is no point checking for healthy target. If a mounter pod has crashed, customers will need to delete and recreate their workloads.

jet-tong: Am aware we can't revive dead targets, just wondering if we could make this failure more visible (than saying 'target is already mounted' and not asking kubelet to retry), e.g. an error log or etc highlighting dead target to prompt restarting the workload pod.

rajdchak: why would we want the kubelet to retry? what would happen on retry?

rajdchak:
Planning to have a follow up PR to document this feature in README.

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: As discussed with Jensen, let's have a task as part of Observability to make changes to inform customers which targets are dead

jet-tong: (As discussed offline - this will be part of follow-up PR later when we have a better way to make dead targets easily debuggable for customers e.g. with pod id)

pkg/driver/node/mounter/daemonset_mounter_test.go
rajdchak: will remove this in the PR for documentation

jet-tong: Please do keep a list of them somewhere (inc the potential test refactor from above)!

rajdchak: Created a task for it.

pkg/driver/node/mounter/daemonset_mounter.go
vladem: Good that we now return a response / log something in case of a time-outing `IsHealthyMountpoint`. Though we do still leak a goroutine blocked on a syscall. I'd invest a bit time into trying to reproduce this, maybe as a separate task. Questions are:

1. Do we release `IsHealthyMountpoint` goroutines eventually? probably when `NodePublish` / periodic cleanup does `unmount` of source?
2. A typical case for this "mount happened but fuse daemon (MP) never started serving on it." is when mounter never launched MP and never written an error file, while we were creating a source mount in `fuseMount`. Do we handle this cleanly? `fuseMount` returns an error and the defer-ed cleanup in `mountOrShareSource` successfully unmounts, releasing `CheckMountpoint` goroutine?

vladem (approved):

pkg/driver/node/mounter/daemonset_mounter_test.go
jet-tong: Saw it was fixed - thanks.

jet-tong (approved):

PR #865, opened 2026-07-20
Extend `provider_test` to run with daemonset mounter kind

vladem:
*Issue #, if available:* N/A

*Description of changes:*

1. For each of the supported cred source types run unit tests for all mounter kind-s (daemonset, pod, systemd)
2. Tests assert **the expected token name**, content of the token file, env vars for MP process (includes **IRSA role**, fetched from k8s mock API)
4. Moved cleanup tests to the end of the provisioning tests to avoid duplicate code

Main tests:

1. DriverLevel, long-term
2. DriverLevel, IRSA (web identity)
3. DriverLevel, EKS PI (container)
5. PodLevel, IRSA
6. PodLevel, EKS PI

Also DriverLevel creds may provide several types, e.g. IRSA + long-term. Also extended unhappy test scenarios with incomplete creds. 

---

### DIFF
--- a/pkg/driver/node/credentialprovider/provider_test.go
+++ b/pkg/driver/node/credentialprovider/provider_test.go
@@ -28,7 +28,6 @@ const testContainerCredentialsFullURI = "http://169.254.170.23/v1/credentials"
 const serviceAccountTokenAudienceEKS = "pods.eks.amazonaws.com"
 const testContainerAuthorizationToken = "test-container-authorization-token"
 const testEKSPodIdentityServiceAccountToken = "eks-pod-identity-token"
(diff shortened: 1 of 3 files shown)


pkg/driver/node/credentialprovider/provider_test.go
jet-tong: np: this function is replaced by `provideCtx := credentialprovider.ProvideContext{` now, if they're not used anymore perhaps we can delete this now?

jet-tong (commented):
Super clean refactors! 

LGTM overall, but I'd like a second look later since this is a big unit test change.

jet-tong (approved):

PR #879, opened 2026-07-31
Add `Client.expressAZ` to get the S3 Express AZ used in the test

muddyfish:
*Issue #, if available:* N/A

*Description of changes:*

- Do not fail if S3 Express is not present.
- Add new environment variable `S3_EXPRESS_AZ` which can be set to override the Express AZ

### DIFF
--- a/tests/e2e-kubernetes/s3client/client.go
+++ b/tests/e2e-kubernetes/s3client/client.go
@@ -4,6 +4,7 @@ package s3client
 import (
 	"context"
 	"fmt"
+	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


tests/e2e-kubernetes/s3client/client.go
```diff
@@ -78,12 +80,22 @@ func (c *Client) CreateStandardBucket(ctx context.Context) (string, DeleteBucket
 	return bucketName, c.create(ctx, input)
 }
 
+// expressAZ returns the S3 Express availability zone for the client's region.
+// It checks the S3_EXPRESS_AZ environment variable first, allowing callers to override
+// the built-in map for regions not yet in the hardcoded list.
+func (c *Client) expressAZ() string {
+	if az := os.Getenv("S3_EXPRESS_AZ"); az != "" {
```
jet-tong, line 87: np: Perhaps we can put the os.Getenv in a var at the top of this file, or just the string in a const like in provider.go.

jet-tong (approved):

PR #880, opened 2026-07-31
Use TEST_POD_IMAGE environment variable for defining the test image

muddyfish:
*Issue #, if available:* N/A

*Description of changes:*

- Instead of using `busybox`, use AL2023 as our global test image
- Allow the image to be overridden through an environment variable `TEST_POD_IMAGE`
- Add a test to verify that the new variable works
- Overrides the k8s `busybox` image because then it impacts the normal k8s tests. `KUBE_TEST_REPO_LIST` does not work as we need to override image tag as well as repo & repository.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,6 +155,9 @@ test:
 	# skipping controller test cases because we don't implement controller for static provisioning,
 	# this is a known limitation of sanity testing package: https://github.com/kubernetes-csi/csi-test/issues/214
 	go test -v ./tests/sanity/... -ginkgo.skip="ControllerGetCapabilities" -ginkgo.skip="ValidateVolumeCapabilities" -ginkgo.skip="should remove target path"
+	# Run unit tests in e2e-kubernetes (image override, parsing, etc.) without requiring a cluster.
+	# TestE2E is the Ginkgo e2e suite entry point which requires a live cluster, so we skip it.
+	cd tests/e2e-kubernetes && go test -v -skip TestE2E ./...
 
 .PHONY: cover
 cover:
--- a/tests/e2e-kubernetes/e2e_test.go
+++ b/tests/e2e-kubernetes/e2e_test.go
@@ -3,11 +3,14 @@ package e2e
 import (
 	"context"
 	"flag"
+	"fmt"
+	"os"
 	"testing"
 	"time"
 
 	"github.com/awslabs/mountpoint-s3-csi-driver/tests/e2e-kubernetes/s3client"
 	custom_testsuites "github.com/awslabs/mountpoint-s3-csi-driver/tests/e2e-kubernetes/testsuites"
+	"github.com/distribution/reference"
 
 	ginkgo "github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
@@ -16,6 +19,7 @@ import (
 	"k8s.io/kubernetes/test/e2e/storage/framework"
 	"k8s.io/kubernetes/test/e2e/storage/testsuites"
 	"k8s.io/kubernetes/test/e2e/storage/utils"
+	imageutils "k8s.io/kubernetes/test/utils/image"
 )
 
 func init() {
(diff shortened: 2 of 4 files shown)


tests/e2e-kubernetes/e2e_test.go
```diff
@@ -39,6 +42,62 @@ func init() {
...
+	configs[imageutils.BusyBox] = cfg
+}
+
+// parseImageRef splits a fully-qualified image reference into registry, name, and tag.
+// Examples:
+//
+//	"public.ecr.aws/amazonlinux/amazonlinux:2023" -> ("public.ecr.aws", "amazonlinux/amazonlinux", "2023")
+//	"registry.k8s.io/e2e-test-images/busybox:1.37.0-1" -> ("registry.k8s.io", "e2e-test-images/busybox", "1.37.0-1")
+//	"nginx:latest" -> ("", "nginx", "latest")
+func parseImageRef(image string) (registry, name, version string) {
```
jet-tong: I don't particularly want us to maintain a new function just for tests. Is there an existing package to make this PR much shorter? Saw `github.com/distribution/reference` [example here](https://github.com/docker/cli/blob/52b15cc571f187b521c090bae2be05a54f2090d1/cli/command/image/push.go#L95) that may be relevant?

tests/e2e-kubernetes/e2e_test.go
```diff
@@ -39,6 +42,62 @@ func init() {
...
+	// Override the upstream K8s e2e framework's default test image (BusyBox) with
+	// our TEST_POD_IMAGE so that tests like InitVolumesTestSuite use an image
+	// accessible from all regions instead of registry.k8s.io.
+	overrideUpstreamTestImage()
+}
+
+func overrideUpstreamTestImage() {
+	img := os.Getenv("TEST_POD_IMAGE")
+	if img == "" {
+		img = "public.ecr.aws/amazonlinux/amazonlinux:2023"
```
jet-tong, line 56: Would be interested to see if using AL2023 for all tests would lengthen the test time by even a bit (since its probably quite a bit bigger than busybox to download) or if it's negligible changes.

tests/e2e-kubernetes/e2e_test.go
muddyfish: Fixed

tests/e2e-kubernetes/e2e_test.go
muddyfish: I changed the default as the new EKS addon tests will use AL2023 as a base image, and I want our regular e2e test suite to be in sync so we don't get any errors about missing commands that might be in busybox but not in AL.

Makefile
```diff
@@ -155,6 +155,9 @@ test:
 	# skipping controller test cases because we don't implement controller for static provisioning,
 	# this is a known limitation of sanity testing package: https://github.com/kubernetes-csi/csi-test/issues/214
 	go test -v ./tests/sanity/... -ginkgo.skip="ControllerGetCapabilities" -ginkgo.skip="ValidateVolumeCapabilities" -ginkgo.skip="should remove target path"
+	# Run unit tests in e2e-kubernetes (image override, parsing, etc.) without requiring a cluster.
+	# TestE2E is the Ginkgo e2e suite entry point which requires a live cluster, so we skip it.
+	cd tests/e2e-kubernetes && go test -v -skip TestE2E ./...
```
jet-tong, line 160: QQ - why not keep it as part of e2e test suite as a test for test logic? I felt like putting image_override_test.go not in unit test folder nor in testsuites folder is a little bit hacky, and also this means unit tests would now require TEST_POD_IMAGE env var (not sure about the implications of that).

muddyfish: I made it a unit test deliberately as I didn't think it made much sense as a e2e test. The unit tests don't require `TEST_POD_IMAGE` as it has a default value, but supports it if present.

jet-tong: That's fair, unit test makes more sense than e2e test here, I just thought the placement of test for e2e test helper as part of make test unit test is a tiny bit off, but it should be fine :)

jet-tong (approved):

PR #891, opened 2026-08-04
ci: Restrict verify helm images workflow to PRs targeting main

Priyankakarumuru1:
*Description of changes:*
The verify helm images workflow currently runs on all pull requests, including release branch PRs where the image intentionally doesn't exist yet (it gets built after tagging). This restricts the verify helm images check to only run on PRs targeting main, preventing false failures on release branch PRs where the image has not yet been published.

### DIFF
--- a/.github/workflows/verify-helm-images.yaml
+++ b/.github/workflows/verify-helm-images.yaml
@@ -14,6 +14,7 @@ jobs:
   verify:
     name: Verify chart images exist in public ECR
     runs-on: ubuntu-latest
+    if: github.event_name != 'pull_request' || github.base_ref == 'main'
     steps:
       - uses: actions/checkout@v7
       - name: Install crane


.github/workflows/verify-helm-images.yaml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches: [ "main", "feature/*" ]
   pull_request:
+    branches: [ "main" ]
```
yerzhan7: I think we would need to remove also "Verify chart images exist in public ECR" status check as required for the release branches.

Alternatively we can do:

```yaml
      name: Verify chart images exist in public ECR
      runs-on: ubuntu-latest
+     if: github.event_name != 'pull_request' || github.base_ref == 'main'
      steps:
```

From https://docs.github.com/en/pull-requests/reference/status-checks

> A job that is skipped will report its status as "Success". It will not prevent a pull request from merging, even if it is a required check.

yerzhan7 (dismissed):

.github/workflows/verify-helm-images.yaml
Priyankakarumuru1: Thanks! Looked at this and I agree this is the better approach (skipping the job reports as "Success" so it won't block required checks on release PRs). I'll update the PR with the if condition instead.

yerzhan7 (approved):

PR #893, opened 2026-08-05
fix: grant leader election event permissions

### LINKED ISSUE #886: Controller ServiceAccount missing events permission
**What happened?**

With the introduction of leader election in v2.7.0 (PR #814), after upgrading to aws-mountpoint-s3-csi-driver v2.7.0, the controller repeatedly logs the following error whenever it emits a LeaderElection event:
  
 ```
event.go:359] "Server rejected event (will not retry!)"
    err="events is forbidden: User "system:serviceaccount:kube-system:s3-csi-driver-controller-sa"
         cannot create resource "events" in API group "" in the namespace "kube-system""
    ... Reason:LeaderElection, Message: ... became leader ...
  
```
  The v2.7.0 leader-election Role (`s3-csi-driver-controller-leader-election-role`) only grants permissions on `coordination.k8s.io/leases`, and there is no RBAC rule granting the controller ServiceAccount (`s3-csi-driver-controller-sa`) permission to create/patch core events:
  
```
  rules:
  - apiGroups: ["coordination.k8s.io"]
    resourceNames: ["aws-s3-csi-controller"]
    resources: ["leases"]
    verbs: ["get", "update"]
  - apiGroups: ["coordination.k8s.io"]
    resources: ["leases"]
    verbs: ["create"]
```
  
  The same omission is present in the upstream Helm chart:
  https://github.com/awslabs/mountpoint-s3-csi-driver/blob/v2.7.0/charts/aws-mountpoint-s3-csi-driver/templates/serviceaccount-csi-controller.yaml
  

**Request**
Add the missing core events permission (create, patch) for the controller ServiceAccount in the chart RBAC (leader-election Role) so that LeaderElection events can be recorded and the recurring "events is forbidden" errors are eliminated in a future release / add-on build.

**Anything else we need to know?**:
Leader election itself works correctly - the Lease is acquired as expected. The only impact is that the informational LeaderElection Event object fails to be recorded, resulting in repeated error-level log noise with no functional degradation. For EKS managed add-on users, the Role cannot be patched manually without conflicting with add-on reconciliation.

 
  

**Environment**
- Driver version: v2.7.0 (EKS managed add-on)

maxi-maxima:
## Summary

- grant the controller leader-election Role permission to create and patch core Events
- keep the existing Lease permissions unchanged

## Root cause

Leader election was added in v2.7.0, but its namespaced Role only permits access to `coordination.k8s.io/leases`. The client-go event recorder also creates and patches core Events, so the controller logs `events is forbidden` whenever it records leadership changes.

## Validation

- `helm lint charts/aws-mountpoint-s3-csi-driver`
- rendered the default chart and asserted the leader-election Role contains the Events rule
- rendered with `isEKSAddon=true` and asserted the Events rule is present
- `git diff --check`

Closes #886

### DIFF
--- a/charts/aws-mountpoint-s3-csi-driver/templates/serviceaccount-csi-controller.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/serviceaccount-csi-controller.yaml
@@ -92,6 +92,9 @@ rules:
   - apiGroups: ["coordination.k8s.io"]
     resources: ["leases"]
     verbs: ["create"]
+  - apiGroups: [""]
+    resources: ["events"]
+    verbs: ["create", "patch"]
 ---
 kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
--- a/deploy/kubernetes/base/serviceaccount-csi-controller.yaml
+++ b/deploy/kubernetes/base/serviceaccount-csi-controller.yaml
@@ -91,6 +91,9 @@ rules:
   - apiGroups: ["coordination.k8s.io"]
     resources: ["leases"]
     verbs: ["create"]
+  - apiGroups: [""]
+    resources: ["events"]
+    verbs: ["create", "patch"]
 
 ---
 kind: RoleBinding


yerzhan7:
The change LGTM.

Could you please use our PR template for PR description? in particular, could you please add as per the template:

> By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

maxi-maxima:
Thanks — I added the template confirmation to the PR description. The remaining ROSA failure is infrastructure-side: Terraform reported a duplicate fixed cluster name (`s3-csi-rosa-untrusted`), while the full EKS matrix and unit/controller checks passed. I don't have permission to rerun the upstream workflow. Could you please rerun the failed E2E workflow?

yerzhan7:
We recently discovered that we forgot to include leader election permissions in our Kustomize distribution. https://github.com/awslabs/mountpoint-s3-csi-driver/pull/896

Could you please add these new event permissions to Kustomize as well? (Afterwards, I will re-run the CI)

Apologies for the churn.

maxi-maxima:
Thanks for the pointer. I added the same leader election event permissions to the Kustomize base in deploy/kubernetes/base/serviceaccount-csi-controller.yaml.

Validation run locally:
- kubectl kustomize deploy/kubernetes/base
- git diff --check

Pushed in 3449073.

yerzhan7 (approved):

PR #894, opened 2026-08-05
Return mount success, when target is mounted and mountpoint pod can not be found

vladem:
*Issue #, if available:* N/A

*Description of changes:*

When `NodePublishVolume` is called for a volume that is already published at the target (credential refresh via `requiresRepublish`), but the backing Mountpoint Pod no longer exists, the driver currently waits 15s then returns an error. This triggers kubelet's per-volume exponential backoff (max 2m2s). Since kubelet serializes all `NodePublishVolume` calls for the same volume behind a single backoff timer, orphaned workloads (whose MP pod was deleted) repeatedly consume the slot and delay new workloads with healthy MP pods from mounting. With ~20 orphaned workloads contending, a new workload has a ~1/21 chance of being picked each cycle, with a 2m+ delay between attempts.

This change returns success when the target is already a mountpoint and the MP pod is not found — the mount is unrecoverable regardless (FUSE session is gone), but returning an error only causes collateral damage to other workloads. This also aligns with the [CSI spec](https://github.com/container-storage-interface/spec/blob/master/spec.md#nodepublishvolume): "This operation MUST be idempotent. If the volume corresponding to the `volume_id` has already been published at the specified `target_path` [...] the Plugin MUST reply `0 OK`."

---

### DIFF
--- a/pkg/driver/node/mounter/pod_mounter.go
+++ b/pkg/driver/node/mounter/pod_mounter.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
@@ -135,6 +136,15 @@ func (pm *PodMounter) Mount(ctx context.Context, bucketName string, target strin
 	if err != nil {
 		klog.Errorf("Failed to find corresponding MountpointS3PodAttachment custom resource for %q: %v. %s", target, err, pm.helpMessageForGettingControllerLogs())
 		return fmt.Errorf("Failed to find corresponding MountpointS3PodAttachment custom resource: %w. %s", err, pm.helpMessageForGettingControllerLogs())
+	}
+
+	// If target is already mounted (republish) but the MP pod is gone, return success to avoid
+	// kubelet exponential backoff that would block NodePublishVolume for all other pods on this volume.
+	if isTargetMountPoint {
+		if _, err := pm.podWatcher.Get(mpPodName); apierrors.IsNotFound(err) {
+			klog.Errorf("Mountpoint Pod %q for %q not found, credentials not updated. %s", mpPodName, target, pm.helpMessageForGettingMountpointPodStatus(err, mpPodName))
+			return nil
+		}
 	}
 
 	pod, podPath, err := pm.waitForMountpointPod(ctx, mpPodName)
--- a/pkg/driver/node/mounter/pod_mounter.go
+++ b/pkg/driver/node/mounter/pod_mounter.go
@@ -104,10 +104,12 @@ func NewPodMounter(
 // If Mountpoint is already mounted at `target`, it will return early at step 3 to ensure credentials are up-to-date.
 // If Mountpoint is already mounted at `source`, it will skip steps 4-7 and only perform bind mount to `target`.
 func (pm *PodMounter) Mount(ctx context.Context, bucketName string, target string, credentialCtx credentialprovider.ProvideContext, args mountpoint.Args, fsGroup string, userEnv envprovider.Environment) error {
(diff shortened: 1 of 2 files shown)


pkg/driver/node/mounter/pod_mounter.go
```diff
@@ -137,6 +138,15 @@ func (pm *PodMounter) Mount(ctx context.Context, bucketName string, target strin
 		return fmt.Errorf("Failed to find corresponding MountpointS3PodAttachment custom resource: %w. %s", err, pm.helpMessageForGettingControllerLogs())
 	}
 
+	// If target is already mounted (republish) but the MP pod is gone, return success to avoid
+	// kubelet exponential backoff that would block NodePublishVolume for all other pods on this volume.
```
rajdchak, line 144: how long does the kubelet keep trying? is it an indefinite block?

vladem: Kubelet doesn't stop retrying, the issue is in the added delay between retries (exponential backoff, up to 2m) and length of the NPV request (15s to wait for deleted MP pod). The issue is that if 20 orphaned workloads (with deleted MP pod) are contending for the same NPV slot, a new workload (with running MP) might not have a chance to mount for a long time (1/21 chance to be picked, if not picked - a 2m delay).

pkg/driver/node/mounter/pod_mounter.go
```diff
@@ -137,6 +138,15 @@ func (pm *PodMounter) Mount(ctx context.Context, bucketName string, target strin
...
+	if isTargetMountPoint {
+		if _, err := pm.podWatcher.Get(mpPodName); apierrors.IsNotFound(err) {
+			klog.Errorf("Mountpoint Pod %q for %q not found, credentials not updated. %s", mpPodName, target, pm.helpMessageForGettingMountpointPodStatus(err, mpPodName))
```
jet-tong: This message is "Seems like Mountpoint Pod is in 'Pending' status because it is unschedulable" - do we perhaps want to write a new message for this specific case where MP pod is not found? 

Would it be to check s3pa and delete all workloads associated with that MP pod, wait for s3pa entry to be removed, after then restart the workloads?

pkg/driver/node/mounter/pod_mounter.go
jet-tong: np - would be good to include this in PR description. 

You also mentioned this change is also a fix to comply with the NPV contract according to [CSI docs](https://github.com/container-storage-interface/spec/blob/master/spec.md#nodepublishvolume)
> This operation MUST be idempotent. If the volume corresponding to the volume_id has already been published at the specified target_path, and is compatible with the specified volume_capability and readonly flag, the Plugin MUST reply 0 OK.

pkg/driver/node/mounter/pod_mounter.go
vladem: How about this?

```
klog.Errorf("Mountpoint Pod %q for %q not found, credentials not updated. Mount will not recover — restart the workload %q to get a fresh mount.", mpPodName, target, workloadPodID)
```

We already have UID of the workload which is orphaned, so just log it. Querying s3Pa probably won't give more information, right?

pkg/driver/node/mounter/pod_mounter.go
```diff
@@ -103,10 +104,12 @@ func NewPodMounter(
 // If Mountpoint is already mounted at `target`, it will return early at step 3 to ensure credentials are up-to-date.
 // If Mountpoint is already mounted at `source`, it will skip steps 4-7 and only perform bind mount to `target`.
 func (pm *PodMounter) Mount(ctx context.Context, bucketName string, target string, credentialCtx credentialprovider.ProvideContext, args mountpoint.Args, fsGroup string, userEnv envprovider.Environment) error {
-	volumeName, err := pm.volumeNameFromTargetPath(target)
+	tp, err := targetpath.Parse(target)
 	if err != nil {
 		return fmt.Errorf("Failed to extract volume name from %q: %w", target, err)
```
jet-tong, line 109: small np: Failed to extract target path perhaps?

jet-tong (dismissed):

PR #903, opened 2026-08-14
Backport docker e2e tests

muddyfish:
*Issue #, if available:* N/A

*Description of changes:*

Backport conversion of e2e test suite to being docker compatible to V1. 

We only needed to change how ginkgo is launched compared to v2 as we changed some options.


muddyfish:
Re-opening from inside the repo rather than a fork

PR #905, opened 2026-08-14
V1 to v3 upgrade support

rajdchak:
*Issue #, if available:*

*Description of changes:*

Add support for maintenance of V1 mounts post upgrade to V3, mainly:
1. credential refresh.
2. clean unmount.

Tested by running the upgrade_and_rollback E2E test

### DIFF
--- a/charts/aws-mountpoint-s3-csi-driver/values.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/values.yaml
@@ -135,11 +135,6 @@ eksPodIdentityAgent:
 # TODO: Remove this in v3 as systemd mount support will be discontinued
 supportLegacySystemDMounts: true
 
-# Time in seconds to wait for host mounts to propagate into the CSI node container's
-# mount namespace after a pod restart. This ensures RebuildMountMap can see existing
-# FUSE mounts before processing. Set to 0 to disable the delay.
-mountPropagationDelaySeconds: 5
-
 experimental:
   # Enables support for `s3.csi.aws.com/reserve-headroom-for-mppod` scheduling gate on the Workload Pods.
   # See https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/docs/HEADROOM_FOR_MPPOD.md for more details.
--- a/charts/aws-mountpoint-s3-csi-driver/values.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/values.yaml
@@ -135,6 +135,11 @@ eksPodIdentityAgent:
 # TODO: Remove this in v3 as systemd mount support will be discontinued
 supportLegacySystemDMounts: true
 
+# Time in seconds to wait for host mounts to propagate into the CSI node container's
+# mount namespace after a pod restart. This ensures RebuildMountMap can see existing
+# FUSE mounts before processing. Set to 0 to disable the delay.
+mountPropagationDelaySeconds: 5
+
 experimental:
   # Enables support for `s3.csi.aws.com/reserve-headroom-for-mppod` scheduling gate on the Workload Pods.
   # See https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/docs/HEADROOM_FOR_MPPOD.md for more details.
--- a/deploy/kubernetes/base/node.yaml
+++ b/deploy/kubernetes/base/node.yaml
@@ -74,6 +74,8 @@ spec:
               value: /var/lib/kubelet
             - name: SUPPORT_LEGACY_SYSTEMD_MOUNTS
               value: "true"
+            - name: MOUNT_PROPAGATION_DELAY_SECONDS
+              value: "5"
             - name: MOUNTPOINT_NAMESPACE
               value: mount-s3
             - name: EKS_POD_IDENTITY_AGENT_CONTAINER_CREDENTIALS_FULL_URI
--- a/deploy/kubernetes/base/node.yaml
+++ b/deploy/kubernetes/base/node.yaml
@@ -74,8 +74,6 @@ spec:
               value: /var/lib/kubelet
             - name: SUPPORT_LEGACY_SYSTEMD_MOUNTS
               value: "true"
-            - name: MOUNT_PROPAGATION_DELAY_SECONDS
-              value: "5"
             - name: MOUNTPOINT_NAMESPACE
               value: mount-s3
             - name: EKS_POD_IDENTITY_AGENT_CONTAINER_CREDENTIALS_FULL_URI
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -127,19 +127,6 @@ func NewDriver(endpoint string, mpVersion string, nodeID string) (*Driver, error
 		klog.Info("Using daemonset mounter mode")
 		klog.Info("Note: s3-csi-daemonset-mounter uses OnDelete update strategy - helm upgrade will not restart/upgrade mounter pods automatically")
 
-		// Wait for mount propagation to settle before rebuilding the mount map.
-		// After CSI node pod restart (e.g., during upgrade), existing FUSE mounts
-		// on the host need time to propagate into this container's mount namespace
-		// via Bidirectional mount propagation. Without this delay, RebuildMountMap
-		// may not see existing mounts and incorrectly clean them up, or during
-		// upgrades older version mounts may get a new mount.
-		// Configurable via MOUNT_PROPAGATION_DELAY_SECONDS (default: 5s, set to 0 to disable).
-		if propagationDelay := util.MountPropagationDelay(); propagationDelay > 0 {
-			klog.Infof("Waiting %v for mount propagation to settle before rebuilding mount map", propagationDelay)
-			time.Sleep(propagationDelay)
-			klog.Info("Mount propagation delay complete, proceeding with startup")
-		}
-
 		dm := mounter.NewDaemonsetMounter(clientset, nodeID, mpmounter.New(), credProvider, nil, nil, nil)
 
 		// Rebuild mount map from persisted .meta.json files + kernel mount table.
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -127,6 +127,19 @@ func NewDriver(endpoint string, mpVersion string, nodeID string) (*Driver, error
 		klog.Info("Using daemonset mounter mode")
 		klog.Info("Note: s3-csi-daemonset-mounter uses OnDelete update strategy - helm upgrade will not restart/upgrade mounter pods automatically")
 
+		// Wait for mount propagation to settle before rebuilding the mount map.
+		// After CSI node pod restart (e.g., during upgrade), existing FUSE mounts
+		// on the host need time to propagate into this container's mount namespace
+		// via Bidirectional mount propagation. Without this delay, RebuildMountMap
+		// may not see existing mounts and incorrectly clean them up, or during
+		// upgrades older version mounts may get a new mount.
+		// Configurable via MOUNT_PROPAGATION_DELAY_SECONDS (default: 5s, set to 0 to disable).
+		if propagationDelay := util.MountPropagationDelay(); propagationDelay > 0 {
+			klog.Infof("Waiting %v for mount propagation to settle before rebuilding mount map", propagationDelay)
+			time.Sleep(propagationDelay)
+			klog.Info("Mount propagation delay complete, proceeding with startup")
+		}
+
 		dm := mounter.NewDaemonsetMounter(clientset, nodeID, mpmounter.New(), credProvider, nil, nil, nil)
 
 		// Rebuild mount map from persisted .meta.json files + kernel mount table.
--- a/pkg/driver/node/credentialprovider/provider.go
+++ b/pkg/driver/node/credentialprovider/provider.go
@@ -170,7 +170,7 @@ type CleanupContext struct {
 	PodID     string
 	VolumeID  string
 
-	// MountKind indicates whether the mount is managed by systemd, pod mounter or daemonset mounter or daemonset mounter
+	// MountKind indicates whether the mount is managed by systemd, pod mounter or daemonset mounter
 	MountKind MountKind
 }
 
--- a/pkg/driver/node/credentialprovider/provider.go
+++ b/pkg/driver/node/credentialprovider/provider.go
@@ -170,11 +170,11 @@ type CleanupContext struct {
 	PodID     string
 	VolumeID  string
 
-	// MountKind indicates whether the mount is managed by systemd, pod mounter or daemonset mounter
+	// MountKind indicates whether the mount is managed by systemd, pod mounter or daemonset mounter or daemonset mounter
 	MountKind MountKind
 }
 
-// SetAsSystemDMountpoint marks this context as managed by systemd instead of pod mounter.
+// SetAsSystemDMountpoint marks this context as managed by systemd instead of pod mounter or daemonset mounter.
 func (ctx *CleanupContext) SetAsSystemDMountpoint() {
 	ctx.MountKind = MountKindSystemd
 }
--- a/pkg/driver/node/mounter/daemonset_mounter.go
+++ b/pkg/driver/node/mounter/daemonset_mounter.go
@@ -409,6 +409,9 @@ func (dm *DaemonsetMounter) Unmount(ctx context.Context, target string, credenti
 		}
 		credentialCtx.SetAsSystemDMountpoint()
 		credentialCtx.WritePath = hostPluginDirWithDefault()
+		// Best-effort credential cleanup: the unmount above already succeeded, so we
+		// don't fail the RPC over leftover cred files. A retry would skip this branch
+		// (target no longer mounted) anyway.
 		if err := dm.credProvider.Cleanup(credentialCtx); err != nil {
 			klog.Errorf("DaemonsetMounter: failed to clean up V1 (systemd) credentials for %s: %v", target, err)
 		}
--- a/pkg/driver/node/mounter/daemonset_mounter.go
+++ b/pkg/driver/node/mounter/daemonset_mounter.go
@@ -618,48 +618,6 @@ func (dm *DaemonsetMounter) isSystemDMountpoint(target string) bool {
 	return len(references) == 0
 }
 
-// isV1MountOnHost checks if a direct (non-bind) FUSE mount exists at the given target path on the
-// host by reading /proc/1/mountinfo (the init process's mount namespace).
-//
-// The CSI node pod uses Bidirectional mount propagation, so host mounts eventually appear in the
-// container — but there's a brief window after container start where propagation hasn't completed.
-// During this window, IsMountPoint (which reads /proc/self/mountinfo) returns false, but
-// /proc/1/mountinfo already has the entry.
-//
-// To distinguish a direct V1 FUSE mount from a V3 bind mount (both show as "fuse" at the target),
-// we check if any OTHER mountpoint shares the same Major:Minor device. A bind mount shares its
-// device with the source FUSE mount; a direct V1 FUSE mount has a unique device.
-func (dm *DaemonsetMounter) isV1MountOnHost(target string) bool {
-	infos, err := mountutils.ParseMountInfo("/proc/1/mountinfo")
-	if err != nil {
-		klog.V(4).Infof("DaemonsetMounter: cannot parse /proc/1/mountinfo for V1 host mount check: %v", err)
-		return false
-	}
-
-	// Find the FUSE mount entry at the target path.
-	var targetInfo *mountutils.MountInfo
-	for i := range infos {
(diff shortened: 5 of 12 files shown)


charts/aws-mountpoint-s3-csi-driver/values.yaml
```diff
@@ -135,6 +135,11 @@ eksPodIdentityAgent:
 # TODO: Remove this in v3 as systemd mount support will be discontinued
 supportLegacySystemDMounts: true
 
+# Time in seconds to wait for host mounts to propagate into the CSI node container's
```
muddyfish: Does not convey why you'd want to change this, or what it does clearly

pkg/driver/node/credentialprovider/provider.go
```diff
@@ -170,11 +170,11 @@ type CleanupContext struct {
 	PodID     string
 	VolumeID  string
 
-	// MountKind indicates whether the mount is managed by systemd, pod mounter or daemonset mounter
+	// MountKind indicates whether the mount is managed by systemd, pod mounter or daemonset mounter or daemonset mounter
```
muddyfish: This already included 'or daemonset mounter'.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -153,6 +153,25 @@ func NewDaemonsetMounter(clientset kubernetes.Interface, nodeID string, mount *m
 func (dm *DaemonsetMounter) Mount(ctx context.Context, bucketName string, target string,
 	credentialCtx credentialprovider.ProvideContext, args mountpoint.Args, fsGroup string, userEnv envprovider.Environment) error {
 
+	// Handle V1 (systemd) legacy mounts before anything else.
+	// V1 mounts have the FUSE mount directly at the target path (no source + bind-mount indirection).
+	// V3 never puts a FUSE mount at the target — it mounts FUSE at a source path and bind-mounts to target.
+	// Detection: target is already a mountpoint with zero bind-mount references.
+	// The startup delay in driver.go ensures mount propagation has settled before we reach here.
```
muddyfish: What does this mean? What is 'mount propagation'?

tests/e2e-kubernetes/testsuites/upgrade_and_rollback.go
jet-tong: Not sure why there was upgrade.go line 318 here, maybe remove (context: was checking all upgrade.go mentions)

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -563,6 +601,23 @@ func (dm *DaemonsetMounter) unmountIfMounted(target string) error {
...
+// (unlike V2/V3 which bind-mount from a source path, creating references).
+func (dm *DaemonsetMounter) isSystemDMountpoint(target string) bool {
+	if !util.SupportLegacySystemdMounts() {
+		return false
+	}
+
+	references, err := dm.mount.FindReferencesToMountpoint(target)
+	if err != nil {
+		klog.Warningf("DaemonsetMounter: failed to find references to %s for systemd detection: %v", target, err)
+		return false
```
muddyfish, line 616: Why is false safe here if we cannot verify?

pkg/driver/driver.go
```diff
@@ -127,6 +127,19 @@ func NewDriver(endpoint string, mpVersion string, nodeID string) (*Driver, error
 		klog.Info("Using daemonset mounter mode")
 		klog.Info("Note: s3-csi-daemonset-mounter uses OnDelete update strategy - helm upgrade will not restart/upgrade mounter pods automatically")
 
+		// Wait for mount propagation to settle before rebuilding the mount map.
+		// After CSI node pod restart (e.g., during upgrade), existing FUSE mounts
+		// on the host need time to propagate into this container's mount namespace
+		// via Bidirectional mount propagation. Without this delay, RebuildMountMap
```
muddyfish: Is this true?

charts/aws-mountpoint-s3-csi-driver/values.yaml
jet-tong: `RebuildMountMap` might not be clear for customers - adjusting comment for customer perspective would be useful too.

deploy/kubernetes/base/node.yaml
```diff
@@ -74,6 +74,8 @@ spec:
               value: /var/lib/kubelet
             - name: SUPPORT_LEGACY_SYSTEMD_MOUNTS
               value: "true"
+            - name: MOUNT_PROPAGATION_DELAY_SECONDS
+              value: "5"
```
jet-tong: ~~Hardcoded? Should values.yaml mountPropagationDelaySeconds be used? (Or I misread?)~~
(Nevermind I didn't see it is the kustomize overlay - hard code is correct here)

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -153,6 +153,25 @@ func NewDaemonsetMounter(clientset kubernetes.Interface, nodeID string, mount *m
...
 	credentialCtx credentialprovider.ProvideContext, args mountpoint.Args, fsGroup string, userEnv envprovider.Environment) error {
 
+	// Handle V1 (systemd) legacy mounts before anything else.
+	// V1 mounts have the FUSE mount directly at the target path (no source + bind-mount indirection).
+	// V3 never puts a FUSE mount at the target — it mounts FUSE at a source path and bind-mounts to target.
+	// Detection: target is already a mountpoint with zero bind-mount references.
+	// The startup delay in driver.go ensures mount propagation has settled before we reach here.
+	isMounted, err := dm.IsMountPoint(target)
+	if err == nil && isMounted && dm.isSystemDMountpoint(target) {
+		klog.Infof("DaemonsetMounter: target %s is a legacy V1 (systemd) mount for volume %s, refreshing credentials", target, credentialCtx.VolumeID)
```
jet-tong: np: maybe use "Will only refresh credentials." like in pod_mounter.go so user knows only credential refresh is supported for v1 mounts.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -380,11 +399,30 @@ func (dm *DaemonsetMounter) fuseMount(ctx context.Context, bucketName string, mo
 //  2. Decrements refcount in MountMap
 //  3. If refcount reaches 0, unmounts the FUSE source and removes the entry
 func (dm *DaemonsetMounter) Unmount(ctx context.Context, target string, credentialCtx credentialprovider.CleanupContext) error {
+	// Handle V1 (systemd) legacy mounts
+	// Check IsMountPoint first — if not mounted, isSystemDMountpoint would falsely match
```
jet-tong, line 402: qq: Is that also a problem in pod_mounter.go?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -380,11 +399,30 @@ func (dm *DaemonsetMounter) fuseMount(ctx context.Context, bucketName string, mo
...
+	// (zero refs on an unmounted path).
+	isMounted, mountErr := dm.IsMountPoint(target)
+	if mountErr == nil && isMounted && dm.isSystemDMountpoint(target) {
+		klog.Infof("DaemonsetMounter: unmounting legacy V1 (systemd) mount at %s for volume %s", target, credentialCtx.VolumeID)
+		if err := dm.unmountIfMounted(target); err != nil {
+			return fmt.Errorf("failed to unmount V1 (systemd) target %q: %w", target, err)
```
jet-tong, line 408: np: maybe keep `klog.Errorf("Failed to unmount %q: %v", target, err)` log line as in pod_mounter.go

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -563,6 +601,23 @@ func (dm *DaemonsetMounter) unmountIfMounted(target string) error {
 	return dm.mount.Unmount(target)
 }
 
+// isSystemDMountpoint determines whether the specified target path is a V1 systemd-managed mountpoint.
+// A systemd mountpoint has zero bind-mount references to it — systemd mounts directly to the target
+// (unlike V2/V3 which bind-mount from a source path, creating references).
+func (dm *DaemonsetMounter) isSystemDMountpoint(target string) bool {
```
jet-tong, line 608: qq: noted almost exact same function in pod_mounter.go, will we keep them separate for the time being?

pkg/util/env.go
```diff
@@ -4,12 +4,35 @@ import (
...
 func SupportLegacySystemdMounts() bool {
 	return os.Getenv("SUPPORT_LEGACY_SYSTEMD_MOUNTS") == "true"
 }
 
+// MountPropagationDelay returns the configured mount propagation delay.
+// It reads from the MOUNT_PROPAGATION_DELAY_SECONDS environment variable.
+// If unset or empty, returns DefaultMountPropagationDelay.
+// If set to "0", returns 0 (disabling the delay).
+func MountPropagationDelay() time.Duration {
+	val := os.Getenv("MOUNT_PROPAGATION_DELAY_SECONDS")
```
jet-tong: Perhaps reuse GetEnvAsInt below? And we can handle the seconds < 0 case in callsite (like for maxVolumesPerNode)

pkg/driver/node/credentialprovider/provider.go
rajdchak: yeah got merged somehow, missed this

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: https://github.com/rajdchak/rajdchak-fork-mountpoint-s3-csi-driver/blob/main/pkg/driver/node/mounter/pod_mounter.go#L333-L341 we return false in V2 as well. Tried to keep the same behaviour.

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: https://github.com/rajdchak/rajdchak-fork-mountpoint-s3-csi-driver/blob/main/pkg/driver/node/mounter/pod_mounter.go#L119 in pod mounter we do check whether the target is a mountpoint before verifying whether it is a V1 mount.

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: I think so, moving it to a common place might not make much sense as we will deprecate V2

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: To answer what is mount propagation - https://kubernetes.io/docs/concepts/storage/volumes/#mount-propagation.

We have a Bidirectional mount propagation configured on csi node https://github.com/rajdchak/rajdchak-fork-mountpoint-s3-csi-driver/blob/main/charts/aws-mountpoint-s3-csi-driver/templates/node.yaml#L127-L130 so that whatever mounts are created under the kubelet path are visible both from the host and the csi node pod.

Now to answer why did I introduce a delay -
While testing the upgrade path some time back, the upgrade test was failing because V1 credentials weren't being refreshed — our code wasn't recognizing the mount as a legacy V1 mount. Root cause: the newly-recreated CSI node pod couldn't find the V1 target in its mount table.
My hypothesis was a startup timing issue: right after the CSI node pod is recreated during an upgrade, its mount namespace hasn't yet synced with the host's mount table, so pre-existing mounts aren't visible yet. Adding a startup delay before we read the mount table made the test pass consistently, which supported that theory.
More recently, I removed the delay and the test kept passing. So right now I'm trying to determine whether that sync race is actually deterministic — i.e., whether it happens on every restart or only intermittently under certain timing conditions.

rajdchak: since I am unable to reproduce the issue again, will remove the addition of delay for now.

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: may be it will be easy while debugging if we have V1 in the log? Will keep it this way for now

pkg/driver/node/mounter/daemonset_mounter.go
jet-tong: Maybe it's worth tracking this issue as a TODO comment perhaps?

qq: if it's an issue of mount table read being too early, can the background cleanup help with this sync instead?

pkg/util/env.go
jet-tong: Delay removed so not a problem now.

tests/e2e-kubernetes/testsuites/util.go
```diff
@@ -395,3 +395,57 @@ func isDaemonsetMounterMode(ctx context.Context, f *framework.Framework) bool {
 	}
 	return len(pods.Items) > 0
 }
+
+// checkReadFromPathSucceedEventually retries reading from a path in a pod, tolerating
```
jet-tong, line 399: Noted the non-eventually ones are above in this testsuite - maybe place it near the original?

May I ask if we can just replace the checks with eventually checks (or not really) so we don't need to pick between 2 checks later on?

np: minor grammar on the function names (maybe it was pre-existing, e.g. checkListingPathWithEntriesEventually maybe need 'Succeeds' between them? does "checkReadFromPathSucceedEventually" need an s for succeed?

tests/e2e-kubernetes/testsuites/upgrade_and_rollback.go
```diff
@@ -485,8 +492,24 @@ func buildHelmValuesForUpgrade() map[string]any {
 		values["image"] = map[string]any{
 			"repository": helmChartContainerRepository,
 			"tag":        helmChartContainerTag,
+			"pullPolicy": "Always",
```
jet-tong, line 516: qq: What happens/happened without pullPolicy Always?

tests/e2e-kubernetes/testsuites/upgrade_and_rollback.go
```diff
@@ -325,7 +325,7 @@ func (t *s3CSIUpgradeTestSuite) DefineTests(driver storageframework.TestDriver,
 		// Set A	|Before upgrade	| Test pre-upgrade workloads on rollback	|After rollback monitoring
 		// Set B	|Before upgrade	| Test upgrade + termination after upgrade	|After upgrade monitoring
 		// Set C	|After upgrade	| Test upgrade + termination after upgrade	|After upgrade monitoring
-		// Set D	|After upgrade	| Test post-upgrade workloads on rollback	|After rollback monitoring
+		// Set D	|After upgrade	| Test new version works after upgrade		|Before rollback
```
jet-tong: qq: May I ask why the after rollback -> before rollback change was made? (Sorry I'm not very familiar with the upgrade tests)

jet-tong (commented):
Source Code LGTM, got some questions on the upgrade tests mainly.

tests/e2e-kubernetes/testsuites/util.go
rajdchak: there are a lot of tests which are using the non eventually ones and may be that can be the reason for the flakiness. When we try to fix the flaky tests may be we could see if it is due to this and change them to use eventually and get rid of the non eventually ones. However, I don't want to touch all the tests as part of this change.

rajdchak: shall we make the grammar fix later as not urgent, as I will have to make changes in all the places from where it is called

tests/e2e-kubernetes/testsuites/upgrade_and_rollback.go
jet-tong: (Resolved offline - Set D is for same major version rollback tests only)

tests/e2e-kubernetes/testsuites/util.go
jet-tong: Sure that's fine - let's just make sure this is tracked.

jet-tong (approved):

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: Also, I was checking what will happen in this case, it will enter the V3 mount path https://github.com/awslabs/mountpoint-s3-csi-driver/blob/feature/daemonset-architecture/pkg/driver/node/mounter/daemonset_mounter.go#L180-L254. It will create an entry in our mountmap, create a meta file and provide creds inside the comm directory and then it will find that the target is a mountpoint and return hence it will not create a source mount. We have a background cleaner job which is currently a work in progress which iterates over all the entries in our mountmap and triggers a cleanup for all entries for which either the source path is not a mountpoint or it is no longer healthy, so this entry and related resources will be cleaned up. Hence, I think this is safe.

tadiwa-aizen (dismissed):

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -380,6 +398,24 @@ func (dm *DaemonsetMounter) fuseMount(ctx context.Context, bucketName string, mo
...
+	isMounted, mountErr := dm.IsMountPoint(target)
+	if mountErr == nil && isMounted && dm.isSystemDMountpoint(target) {
+		klog.Infof("DaemonsetMounter: unmounting legacy V1 (systemd) mount at %s for volume %s", target, credentialCtx.VolumeID)
+		if err := dm.unmountIfMounted(target); err != nil {
+			return fmt.Errorf("failed to unmount V1 (systemd) target %q: %w", target, err)
+		}
+		credentialCtx.SetAsSystemDMountpoint()
+		credentialCtx.WritePath = hostPluginDirWithDefault()
+		if err := dm.credProvider.Cleanup(credentialCtx); err != nil {
+			klog.Errorf("DaemonsetMounter: failed to clean up V1 (systemd) credentials for %s: %v", target, err)
```
muddyfish, line 416: What are the consequences of not returning `nil` here when there was a failure? It looks like it could be a typo, so a comment might be good here explaining it

tests/e2e-kubernetes/testsuites/util.go
```diff
@@ -395,3 +395,57 @@ func isDaemonsetMounterMode(ctx context.Context, f *framework.Framework) bool {
...
+// transient errors
+func checkReadFromPathSucceedEventually(ctx context.Context, f *framework.Framework, pod *v1.Pod, path string, toWrite int, seed int64) {
+	sum := sha256.Sum256(genBinDataFromSeed(toWrite, seed))
+	cmd := fmt.Sprintf("dd if=%s bs=%d count=1 | sha256sum | grep -Fq %x", path, toWrite, sum)
+	framework.Gomega().Eventually(ctx, func(ctx context.Context) error {
+		return e2epod.VerifyExecInPodSucceed(ctx, f, pod, cmd)
+	}).WithTimeout(30 * time.Second).WithPolling(5 * time.Second).Should(gomega.Succeed())
```
jet-tong: [I think I'm wrong here - since Should Succeed should surface error which I missed]

~~I think we needed this here so this function is masking the error~~
```
err := framework.Gomega().Eventually
...
framework.ExpectNoError(err, ...
```

~~So the eventually functions are actually masking the error.~~

Upgrade tests are not working on main branch - the failure should be elsewhere, let's root cause that before merging this PR.

jet-tong (changes requested):

tests/e2e-kubernetes/testsuites/util.go
rajdchak: Eventually tests have Should(gomega.Succeed(). It is not masking the errors.

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: added a comment for explaining
// Best-effort credential cleanup: the unmount above already succeeded, so we don't fail the RPC over leftover cred files. A retry would skip this branch (target no longer mounted) anyway.

tests/e2e-kubernetes/testsuites/util.go
rajdchak: there is a diff in behaviour between gomega and framework.Gomega() which was preventing errors from being surfaced even though we had a Should(gomega.Succeed(). Changed to use gomega.

jet-tong: Noted changing to plain gomega unmasks the error (somehow).

jet-tong (approved):

muddyfish (approved):

jet-tong:
As a manual test record:
- Blank -> Blank (v2 -> v2) works after with the seed issue patch like in https://github.com/awslabs/mountpoint-s3-csi-driver/pull/913: https://github.com/jet-tong/mountpoint-s3-csi-driver/actions/runs/32240738248
- v1.15.0 -> 0.0.0-source (v3) with major version ticked: All non-AL2023 workflows passes, SELinux issue for AL2023 in v1 remaining: https://github.com/jet-tong/mountpoint-s3-csi-driver/actions/runs/32279458734

PR #911, opened 2026-08-18
Add periodic resource cleanup for daemonset mounter

tadiwa-aizen:
*Issue #, if available:*

*Description of changes:*

Adds a periodic cleanup job for the daemonset mounter — the safety net that reclaims mount resources left behind when inline unmount doesn't fire.

The cleanup job runs every 2 minutes: it walks the `MountMap`, checks each entry against the kernel mount table, and tears down any source that is missing, dead, or has no consumers (source mount, meta file, cred dir). For a source still in the mount table, it probes liveness with a bounded timeout — a dead source is reclaimed, while an undeterminable ("unknown") result is left intact so a live source is never torn down. `releaseTarget` is made concurrency-safe since the job now also deletes map entries.

The republish churn a dead mount would otherwise cause (kubelet re-issuing NodePublishVolume ~every 60s, recreating meta/creds for the job to delete again) is prevented by #918, which this PR is rebased on top of — so the cleanup job reclaims a dead source without fighting republish.

*Testing:*

**Unit tests** (`mount_map_persistence_test.go`, next to the `RebuildMountMap` tests) call `CleanupOrphans()` directly and cover each branch: teardown of a missing, dead, or unused source, refcount reconcile, adoption of an untracked live bind mount, teardown-failure retry, and the skip cases.

**e2e tests** (`resource_cleanup_daemonset.go`) — two Serial specs: a cleanup pass leaves a healthy in-use mount intact, and the same PV mounts again after its source is reclaimed. Both pass on a live EKS cluster.

**Manual testing.** Installed this change on an EKS cluster in daemonset mode. With a workload pod running, I killed only the mounter pod so the source went dead (no inline cleanup fires — only the periodic job can reclaim it):

- The cleanup job left the healthy mount intact on each 2-minute tick (`cleanup: volume s3-pv healthy with 1 live consumer(s), leaving intact`), then detected the dead source on the first tick after the crash and reclaimed it (`cleanup: source .../mnt/s3-pv for volume s3-pv is dead, cleaning up`) — source removed from the mount table, meta file deleted.
- With #918 in place, the ~60s republishes did not recreate the reclaimed resources — the meta file was written once, at the original mount, and never again. No churn.
- Recreating the workload pod mounted a fresh source and read back previously written data.

### DIFF
--- a/pkg/driver/node/mounter/daemonset_mounter.go
+++ b/pkg/driver/node/mounter/daemonset_mounter.go
@@ -64,6 +64,12 @@ const (
 	commDirStaleCheckInterval = 1 * time.Second
 	commDirDiscoveryTimeout   = 60 * time.Second
 	commDirRediscoveryTimeout = 15 * time.Second
+
+	// cleanupInterval is how often the periodic cleanup job runs.
+	cleanupInterval = 2 * time.Minute
+
+	// cleanupHealthCheckTimeout caps the source health probe so a frozen FUSE daemon can't stall cleanup.
+	cleanupHealthCheckTimeout = 10 * time.Second
 )
 
 var mounterNamespace = os.Getenv("MOUNTER_NAMESPACE")
@@ -450,19 +456,28 @@ func (dm *DaemonsetMounter) Unmount(ctx context.Context, target string, credenti
 
 // releaseTarget handles the Unmount flow with MountMap refcounting.
 func (dm *DaemonsetMounter) releaseTarget(target string, volumeID string, credentialCtx credentialprovider.CleanupContext) error {
-	entry := dm.mountMap.Get(volumeID)
-	if entry == nil {
-		// No entry means this volume is not tracked (e.g., after CSI node restart without
-		// successful RebuildMountMap recovery). Best-effort unmount the target in case a
-		// stale bind mount still exists in the kernel mount table.
-		klog.V(4).Infof("DaemonsetMounter: no mount map entry for volume %s, best-effort unmount of %s", volumeID, target)
-		if err := dm.unmountIfMounted(target); err != nil {
-			klog.Errorf("DaemonsetMounter: best-effort unmount of untracked target %s failed: %v", target, err)
+	// The periodic cleanup job also deletes entries from the map, so between our
+	// Get and our Lock the entry could be deleted/replaced. Re-check after locking
+	// that we still hold the map's canonical entry; retry if not.
+	var entry *MountEntry
+	for {
+		entry = dm.mountMap.Get(volumeID)
+		if entry == nil {
+			// No entry means this volume is not tracked (e.g., after CSI node restart without
+			// successful RebuildMountMap recovery). Best-effort unmount the target in case a
+			// stale bind mount still exists in the kernel mount table.
+			klog.V(4).Infof("DaemonsetMounter: no mount map entry for volume %s, best-effort unmount of %s", volumeID, target)
+			if err := dm.unmountIfMounted(target); err != nil {
+				klog.Errorf("DaemonsetMounter: best-effort unmount of untracked target %s failed: %v", target, err)
+			}
+			return nil
 		}
-		return nil
+		entry.mu.Lock()
+		if dm.mountMap.Get(volumeID) == entry {
+			break
+		}
+		entry.mu.Unlock() // orphaned entry (deleted by concurrent unmount/cleanup), retry
 	}
-
-	entry.mu.Lock()
 	defer entry.mu.Unlock()
 
 	// Unmount the bind mount at target.
@@ -495,6 +510,113 @@ func (dm *DaemonsetMounter) releaseTarget(target string, volumeID string, creden
 
 	klog.V(4).Infof("DaemonsetMounter: volume %s unmounted from %s", volumeID, target)
 	return nil
+}
+
+// StartPeriodicCleanup runs CleanupOrphans on a ticker until stopCh is closed.
+//
+// It's a safety net for when inline cleanup and startup cleanup miss resources orphaned while the driver keeps running with no operation touching a volume
+// e.g. a cleanup that failed at refcount 0, or a source mount left dead after a mounter-pod crash.
+func (dm *DaemonsetMounter) StartPeriodicCleanup(stopCh <-chan struct{}) {
+	ticker := time.NewTicker(cleanupInterval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-stopCh:
+			return
+		case <-ticker.C:
+			dm.CleanupOrphans()
+		}
+	}
+}
+
+// For each entry, under the entry's lock:
+//   - if the source mount is unhealthy, tear it down;
+//   - otherwise reconcile the refcount against the live bind mounts, and if it drops
+//     to zero, tear it down.
+func (dm *DaemonsetMounter) CleanupOrphans() {
+	dm.mountMap.Range(func(volumeID string, entry *MountEntry) bool {
+		dm.cleanupEntry(volumeID, entry)
+		return true
+	})
+}
+
+// cleanupEntry reconciles and, if needed, tears down a single mount entry.
+func (dm *DaemonsetMounter) cleanupEntry(volumeID string, entry *MountEntry) {
+	entry.mu.Lock()
+	defer entry.mu.Unlock()
+
+	// A concurrent unmount may have deleted this entry between the Range read and
+	// the lock. Confirm the map still points to this exact entry; if not, skip it.
+	if dm.mountMap.Get(volumeID) != entry {
+		return
+	}
+
+	// An empty SourcePath means GetOrCreate inserted a blank entry that has not been
+	// given a mount yet (mid-creation). Nothing to clean, so leave it.
+	if entry.SourcePath == "" {
+		return
+	}
+
+	mountInfos, err := dm.mountInfoProviderWithDefault()
+	if err != nil {
+		klog.Errorf("DaemonsetMounter: cleanup: failed to read mount table for volume %s: %v", volumeID, err)
+		return
+	}
+
+	sourceMI := findMountByPath(mountInfos, entry.SourcePath)
+	if sourceMI == nil {
+		// Source is gone from the mount table — no bind mounts can reference it, so
+		// it's a true orphan (e.g. a crash between writing meta and creating the mount).
+		klog.V(2).Infof("DaemonsetMounter: cleanup: source %s for volume %s not in mount table, cleaning up", entry.SourcePath, volumeID)
+		dm.teardownEntry(volumeID, entry)
+		return
+	}
+
+	// Being in the mount table doesn't mean the source still works — a crashed mounter leaves
+	// a dead FUSE mount that lingers but errors on every I/O. So we probe it (with a timeout so
+	// a frozen daemon can't stall the pass). If it's dead, tear it down and reclaim its resources;
+	// if we can't tell, leave it alone rather than risk destroying a source still serving a workload.
+	ctx, cancel := context.WithTimeout(context.Background(), cleanupHealthCheckTimeout)
+	healthy, healthErr := dm.IsSourceHealthy(ctx, entry.SourcePath)
+	cancel()
+	switch {
+	case healthErr != nil:
+		klog.V(4).Infof("DaemonsetMounter: cleanup: source health unknown for volume %s (%v), leaving intact", volumeID, healthErr)
+		return
+	case !healthy:
+		klog.V(2).Infof("DaemonsetMounter: cleanup: source %s for volume %s is dead, cleaning up", entry.SourcePath, volumeID)
+		dm.teardownEntry(volumeID, entry)
+		return
+	}
+
+	liveTargets := findBindMountTargets(mountInfos, deviceID(sourceMI), entry.SourcePath)
+
+	// Sync targets and refcount to the kernel mount table (the source of truth for who is actually bind-mounted).
+	entry.Targets = liveTargets
+	entry.RefCount = len(liveTargets)
+
+	// If the kernel shows zero bind mounts on this source, nobody's using it — tear it down.
+	if len(liveTargets) == 0 {
+		klog.V(2).Infof("DaemonsetMounter: cleanup: volume %s has no remaining consumers, cleaning up", volumeID)
+		dm.teardownEntry(volumeID, entry)
+		return
+	}
+
+	// Source still in use — leave it. Logged (V4) so a healthy-volume pass is observable.
+	klog.V(4).Infof("DaemonsetMounter: cleanup: volume %s healthy with %d live consumer(s), leaving intact", volumeID, len(liveTargets))
+}
+
+// teardownEntry runs cleanupMount and, only on success, removes the in-memory
+// entry and then its meta file (entry before meta, matching the other teardown
+// paths). On failure both are kept so a later pass retries. Caller must hold entry.mu.
+func (dm *DaemonsetMounter) teardownEntry(volumeID string, entry *MountEntry) {
+	cleanupCtx := credentialprovider.CleanupContext{VolumeID: volumeID}
+	if err := dm.cleanupMount(entry, cleanupCtx); err != nil {
+		klog.Errorf("DaemonsetMounter: cleanup: %v (will retry next tick)", err)
+		return
+	}
+	dm.mountMap.Delete(volumeID)
+	RemoveMeta(dm.kubeletPath, volumeID)
 }
 
 // GetErrorFileName returns the error file name for a given volume ID.
--- a/pkg/driver/node/mounter/daemonset_mounter.go
+++ b/pkg/driver/node/mounter/daemonset_mounter.go
@@ -581,7 +581,7 @@ func (dm *DaemonsetMounter) cleanupEntry(volumeID string, entry *MountEntry) {
 	cancel()
 	switch {
 	case healthErr != nil:
-		klog.V(4).Infof("DaemonsetMounter: cleanup: source health unknown for volume %s (%v), leaving intact", volumeID, healthErr)
+		klog.Errorf("DaemonsetMounter: cleanup: source health unknown for volume %s (%v), leaving intact", volumeID, healthErr)
 		return
 	case !healthy:
 		klog.V(2).Infof("DaemonsetMounter: cleanup: source %s for volume %s is dead, cleaning up", entry.SourcePath, volumeID)
--- a/pkg/driver/node/mounter/mount_map.go
+++ b/pkg/driver/node/mounter/mount_map.go
@@ -149,6 +149,16 @@ func (m *MountMap) Delete(volumeID string) {
 	m.entries.Delete(volumeID)
 }
 
+// Range iterates every entry in the map, letting the cleanup job loop over every
+// tracked mount. Return false from fn to stop early.
+// Note: the entry pointer may change under a concurrent mount/unmount, so callers
+// must still lock it and re-check before acting.
+func (m *MountMap) Range(fn func(volumeID string, entry *MountEntry) bool) {
+	m.entries.Range(func(k, v any) bool {
+		return fn(k.(string), v.(*MountEntry))
+	})
+}
+
 // SourceMountPath returns the source mount directory for a given volume.
 // This path is stable across pod lifecycles — it's keyed by volume, not pod.
 func SourceMountPath(kubeletPath, volumeID string) string {
--- a/pkg/driver/node/mounter/mount_map_persistence_test.go
+++ b/pkg/driver/node/mounter/mount_map_persistence_test.go
@@ -688,9 +688,10 @@ func TestCleanupOrphans(t *testing.T) {
 		// mountInfo is the fake kernel mount table for this case.
 		mountInfo func(sourcePath, targetA string) []mountutils.MountInfo
 
-		expectEntryGone bool // map entry removed
-		expectMetaGone  bool // meta file removed
-		expectRefCount  int  // asserted when the entry survives
+		expectEntryGone   bool // map entry removed
+		expectMetaGone    bool // meta file removed
+		expectMetaPresent bool // meta file must remain — a surviving entry relies on it to rebuild the map after a CSI node restart
+		expectRefCount    int  // asserted when the entry survives
 	}{
 		{
 			// Dead source (nothing in the mount table): source, creds, error file,
@@ -715,7 +716,8 @@ func TestCleanupOrphans(t *testing.T) {
 			setup: func(t *testing.T, dm *DaemonsetMounter, fakeMounter *mountutils.FakeMounter, kubeletPath, sourcePath, targetA string) {
 				targetB := filepath.Join(kubeletPath, "pods", "wl-b", "mount")
 				registerSourceMount(t, fakeMounter, sourcePath)
-				seedEntry(dm, volumeID, sourcePath, "", []string{targetA, targetB})
+				entry := seedEntry(dm, volumeID, sourcePath, "", []string{targetA, targetB})
+				assert.NoError(t, WriteMeta(kubeletPath, entry))
 			},
 			mountInfo: func(sourcePath, targetA string) []mountutils.MountInfo {
 				return []mountutils.MountInfo{ // targetB is gone from the kernel
@@ -723,7 +725,8 @@ func TestCleanupOrphans(t *testing.T) {
 					{MountPoint: targetA, Major: 0, Minor: 42},
 				}
 			},
-			expectRefCount: 1,
+			expectRefCount:    1,
+			expectMetaPresent: true,
 		},
 		{
 			// Teardown's cleanupMount fails (unmount errors): the entry and meta are
@@ -738,7 +741,8 @@ func TestCleanupOrphans(t *testing.T) {
 			mountInfo: func(sourcePath, _ string) []mountutils.MountInfo {
 				return []mountutils.MountInfo{{MountPoint: sourcePath, Major: 0, Minor: 42}}
 			},
-			expectRefCount: 0, // entry survives (not torn down)
+			expectRefCount:    0, // entry survives (not torn down)
+			expectMetaPresent: true,
 		},
 		{
 			// Healthy source, no bind mounts left in the kernel: torn down.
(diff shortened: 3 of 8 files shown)


pkg/driver/node/mounter/mount_map.go
```diff
@@ -149,6 +149,16 @@ func (m *MountMap) Delete(volumeID string) {
 	m.entries.Delete(volumeID)
 }
 
+// This adds the ability to iterate every entry in the map.
+// It allows the cleanup job to loop over every trackedmount.
```
jet-tong: np: trackedmount

pkg/driver/node/mounter/mount_map_persistence_test.go
```diff
@@ -642,3 +650,172 @@ func TestRebuildMountMap_DeadSourceCleansCredentials(t *testing.T) {
...
+				assert.NoError(t, WriteMeta(kubeletPath, entry))
+			},
+			mountInfo: func(sourcePath, _ string) []mountutils.MountInfo {
+				return []mountutils.MountInfo{{MountPoint: sourcePath, Major: 0, Minor: 42}}
+			},
+			expectEntryGone: true,
+			expectMetaGone:  true,
+		},
+		{
+			// The bug we fixed: the kernel shows a live bind mount we don't track.
```
jet-tong: np: AI wording, but actually it's good that edge cases are found/fixed/tested.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -437,6 +454,113 @@ func (dm *DaemonsetMounter) releaseTarget(target string, volumeID string, creden
...
+	}
+
+	mountInfos, err := dm.mountInfoProviderWithDefault()
+	if err != nil {
+		klog.Errorf("DaemonsetMounter: cleanup: failed to read mount table for volume %s: %v", volumeID, err)
+		return
+	}
+
+	// Give the source-health check up to 30 seconds; if it hasn't answered by then, give up and move on.
+	ctx, cancel := context.WithTimeout(context.Background(), cleanupSourceHealthTimeout)
```
rajdchak: why is this needed?

tests/e2e-kubernetes/testsuites/resource_cleanup_daemonset.go
```diff
@@ -0,0 +1,127 @@
...
+			"expected 1 FUSE source mount before the crash")
+
+		// Kill only the mounter pod — the source dies but the workload pod keeps
+		// running, so nothing triggers inline cleanup.
+		ginkgo.By("Crashing the mounter pod to leave the source dead")
+		killMounterPodOnNode(ctx, f, targetNode)
+		waitForMounterPodReady(ctx, f, targetNode)
+
+		ginkgo.By("Waiting for periodic cleanup to reclaim the dead source mount")
+		framework.Gomega().Eventually(ctx, func(ctx context.Context) (int, error) {
```
jet-tong: Should check potential error masking seen in v1->v3 PR where framework.Gomega()....Should(Succeed()) actually masks the error, where plain gomega() wouldn't.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -437,6 +454,113 @@ func (dm *DaemonsetMounter) releaseTarget(target string, volumeID string, creden
...
+		return true
+	})
+}
+
+// cleanupEntry reconciles and, if needed, tears down a single mount entry.
+func (dm *DaemonsetMounter) cleanupEntry(volumeID string, entry *MountEntry) {
+	entry.mu.Lock()
+	defer entry.mu.Unlock()
+
+	// Between getting the entry and locking it, a concurrent unmount could havedeleted it..
```
jet-tong: np: syntax near end

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -437,6 +454,113 @@ func (dm *DaemonsetMounter) releaseTarget(target string, volumeID string, creden
...
+	// So after locking, we confirm the map still points to this exact entry; if not, it's stale, skip it.
+	if dm.mountMap.Get(volumeID) != entry {
+		return
+	}
+
+	// An entry with an empty SourcePath is a half-born entry — created but not yet given a mount. This happens in GetOrCreate(volumeID), which inserts a blank entry into the map immediately — before the source path is set.
+	// Thus there's nothing to clean, so we  leave it alone
```
jet-tong: np: syntax near end (Sorry for multiple of these np)

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -437,6 +454,113 @@ func (dm *DaemonsetMounter) releaseTarget(target string, volumeID string, creden
...
+	// Give the source-health check up to 30 seconds; if it hasn't answered by then, give up and move on.
+	ctx, cancel := context.WithTimeout(context.Background(), cleanupSourceHealthTimeout)
+	defer cancel()
+
+	// Dead source (e.g. mounter-pod crash): tear it down.
+	if !dm.IsSourceHealthy(ctx, entry.SourcePath) {
+		klog.V(2).Infof("DaemonsetMounter: cleanup: source %s for volume %s is unhealthy, cleaning up", entry.SourcePath, volumeID)
+		dm.teardownEntry(volumeID, entry)
+		return
+	}
```
jet-tong: >Note: with the crashed pod still running, kubelet re-issued NodePublishVolume
(~60s) and re-wrote meta/creds, so the job re-reaped them each tick. After
deleting the pod, everything settled clean. That churn is existing NodePublish
behavior, not a cleanup bug.

Noted this PR description comment relates to this code. 

Now that we cleanup meta/creds for entries with dead source mounts, perhaps the meta/cred churn is new? Maybe we should re-visit our logic for what happens when mounter pod crash / source unhealthy (perhaps you can help us clarify that too, existing logic for mounter pod crashing is not very clear to me for now). 

Other related info:
- Related upcoming PR in v2: https://github.com/awslabs/mountpoint-s3-csi-driver/pull/894, where [CSI spec NPV](https://github.com/container-storage-interface/spec/blob/master/spec.md#nodepublishvolume) says "This operation MUST be idempotent. If the volume corresponding to the volume_id has already been published at the specified target_path [...] the Plugin MUST reply 0 OK.". I don't have a full view of what we want the dead source behaviour to be, but worth noting if target is already published we should return 0 OK even though mountpoint is not working, so it would mean we should not churn due to NPV failure (which retries with exponential backoff at higher frequency), just on NPV republish successes.
- [This thread](https://github.com/awslabs/mountpoint-s3-csi-driver/pull/862#discussion_r3689687339) where we plan to have dead source mount visibility. 

Idea - if we don't want this churn maybe gate teardown to when there is no targets pointing at source mount? This contradicts our design to always clean up dead source mounts though.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -437,6 +454,113 @@ func (dm *DaemonsetMounter) releaseTarget(target string, volumeID string, creden
...
+
+	// An entry with an empty SourcePath is a half-born entry — created but not yet given a mount. This happens in GetOrCreate(volumeID), which inserts a blank entry into the map immediately — before the source path is set.
+	// Thus there's nothing to clean, so we  leave it alone
+	if entry.SourcePath == "" {
+		return
+	}
+
+	mountInfos, err := dm.mountInfoProviderWithDefault()
```
rajdchak, line 560: move reading of the mount table closer to the place where it is being used.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -390,19 +398,28 @@ func (dm *DaemonsetMounter) Unmount(ctx context.Context, target string, credenti
...
+	var entry *MountEntry
+	for {
+		entry = dm.mountMap.Get(volumeID)
+		if entry == nil {
+			// No entry means this volume is not tracked (e.g., after CSI node restart without
+			// successful RebuildMountMap recovery). Best-effort unmount the target in case a
+			// stale bind mount still exists in the kernel mount table.
+			klog.V(4).Infof("DaemonsetMounter: no mount map entry for volume %s, best-effort unmount of %s", volumeID, target)
+			if err := dm.unmountIfMounted(target); err != nil {
+				klog.Errorf("DaemonsetMounter: best-effort unmount of untracked target %s failed: %v", target, err)
```
jet-tong, line 471: For more visibility later, would it help if we log the untracked targets?

tests/e2e-kubernetes/testsuites/resource_cleanup_daemonset.go
```diff
@@ -0,0 +1,127 @@
...
+			dumpMountTable(ctx, f, targetNode, "waiting for cleanup")
+			return countFuseMountsForVolume(ctx, f, targetNode, pvName), nil
+		}).WithTimeout(cleanupPollTimeout).WithPolling(cleanupPollInterval).Should(gomega.Equal(0))
+
+		// Prove the periodic job did it (not a mounter-restart/remount side effect):
+		// the CSI node pod is never killed here, so its logs hold the cleanup line.
+		// Scope the marker to this PV so a prior test's cleanup can't satisfy it.
+		marker := fmt.Sprintf("for volume %s is unhealthy", pvName)
+		gomega.Expect(verifyCSINodeLogs(ctx, f, targetNode, marker)).To(gomega.BeTrue(),
```
jet-tong: Noted a couple of failures in e2e due to this test.

Probably due to NPV / Mount "klog.V(2).Infof("DaemonsetMounter: source %s is dead for volume %s, will clean up and re-mount", entry.SourcePath, volumeID)" via republish, vs 2 min periodic cleanup, causing 'flakiness'.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -476,6 +493,113 @@ func (dm *DaemonsetMounter) releaseTarget(target string, volumeID string, creden
...
+
+	// Dead source (e.g. mounter-pod crash): tear it down.
+	if !dm.IsSourceHealthy(ctx, entry.SourcePath) {
+		klog.V(2).Infof("DaemonsetMounter: cleanup: source %s for volume %s is unhealthy, cleaning up", entry.SourcePath, volumeID)
+		dm.teardownEntry(volumeID, entry)
+		return
+	}
+
+	sourceMI := findMountByPath(mountInfos, entry.SourcePath)
+	if sourceMI == nil {
```
rajdchak, line 567: can this be true if a isSourceHealthy returns true? can you double check?

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: I think I talked to Tadiwa, he said the NPV is called every 60s because we have requiredRepublish true, but can you double check if the behaviour is similar for other healthy volumes as well?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -429,19 +437,28 @@ func (dm *DaemonsetMounter) Unmount(ctx context.Context, target string, credenti
...
-		// successful RebuildMountMap recovery). Best-effort unmount the target in case a
-		// stale bind mount still exists in the kernel mount table.
-		klog.V(4).Infof("DaemonsetMounter: no mount map entry for volume %s, best-effort unmount of %s", volumeID, target)
-		if err := dm.unmountIfMounted(target); err != nil {
-			klog.Errorf("DaemonsetMounter: best-effort unmount of untracked target %s failed: %v", target, err)
+	// The periodic cleanup job also deletes entries from the map, so between our
+	// Get and our Lock the entry could be deleted/replaced. Re-check after locking
+	// that we still hold the map's canonical entry; retry if not.
+	var entry *MountEntry
+	for {
```
rajdchak, line 463: Is a retry loop required here?  a single recheck whether entry is present after acquiring lock and if not present return, isn't that sufficient?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -64,6 +65,13 @@ const (
 	commDirStaleCheckInterval = 1 * time.Second
 	commDirDiscoveryTimeout   = 60 * time.Second
 	commDirRediscoveryTimeout = 15 * time.Second
+
+	// cleanupInterval is how often the periodic cleanup job runs.
+	cleanupInterval = 2 * time.Minute
```
rajdchak, line 69: in v2 is this constant or configurable?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -476,6 +493,113 @@ func (dm *DaemonsetMounter) releaseTarget(target string, volumeID string, creden
...
+
+	// For any target we think exists (it's in our MountMap's entry.Targets), check whether the kernel's mount table also shows it. If the kernel doesn't, drop it from our record
+	kept := entry.Targets[:0:0]
+	for _, t := range entry.Targets {
+		if slices.Contains(liveTargets, t) {
+			kept = append(kept, t)
+		}
+	}
+	entry.Targets = kept
+	entry.RefCount = len(kept)
```
rajdchak: instead of doing these, I am thinking should we just sync the targets and refcount with the liveTargets from the mount table? I think that will be more reliable. If the kernel shows live bind mounts that aren't tracked in entry.Targets (len(liveTargets) > len(kept)), the code correctly avoids teardown that pass — but it persists a RefCount that's lower than the real number of consumers. A later inline releaseTarget decrements from that low count and could reach 0 while an untracked consumer is still live

rajdchak: while rebuilding the mountmap during csi node restart we rely on the mount table, so I think here also we can do the same.

tests/e2e-kubernetes/testsuites/resource_cleanup_daemonset.go
```diff
@@ -0,0 +1,127 @@
...
+			errs = append(errs, resource.CleanupResource(ctx))
+		}
+		framework.ExpectNoError(errors.NewAggregate(errs), "while cleanup resource")
+	}
+	ginkgo.BeforeEach(func(ctx context.Context) {
+		l = local{}
+		l.config = driver.PrepareTest(ctx, f)
+		ginkgo.DeferCleanup(cleanup)
+	})
+
```
rajdchak, line 68: Some additional tests worth adding mostly suggested by AI, but I too agree


```	
1	Multi-consumer refcount reconcile (3 pods, delete 1, source stays, refcount→2)	❌ Missing (test 2 only covers single-consumer "stays healthy")
2	Per-volume isolation (two PVs, one dead + one healthy, only dead reclaimed)	❌ Missing
3	Full teardown asserts all resources deleted like cred dir meta file etc, not just FUSE count	❌ Missing (test 1 checks only mount count + log line)
4	Remount same PV after reclaim	❌ Missing
5	Concurrent unmount vs cleanup tick (retry-loop race)	❌ Missing
6	Cleanup still works after CSI node restart	❌ Missing
7	Frozen (non-ENOTCONN) source must NOT be reclaimed	❌ Missing
```

tests/e2e-kubernetes/testsuites/resource_cleanup_daemonset.go
tadiwa-aizen: Neither Eventually uses .Should(Succeed()) — they use .Should(BeTrue()) and .Should(Equal(0)), so that masking bug doesn't apply here, to my understanding.

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: On your question: the ~60s NPV is requiresRepublish: true, and it  fires for all mounted volumes of running pods — not just dead ones. Healthy volumes get the same cadence; they just take the "already mounted, refresh creds" path. So the 60s isn't specific to the  dead-source case.
 
And on the churn: cleanup now gates teardown on having no live bind-mount  targets instead of source health, so a dead source with a workload still attached is left alone (reclaimed when the pod is removed) — no more meta/cred churn.

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: Already logged at L407 (volume + target) at V(4), which matches the rest of the unmount flow (e.g. the "unmounted from" and "last  consumer" lines are also V(4)). The failure case is already Errorf on  the next line. _I Lean toward keeping V(4) for consistency_ ....

tests/e2e-kubernetes/testsuites/resource_cleanup_daemonset.go
tadiwa-aizen: Yes, fixed. Removed the mounter-crash approach that raced  NPV/republish. The rewritten e2e keeps the source healthy (nothing  oscillates), waits for a cleanup pass to log it left the volume intact, then checks the mount/data are untouched — passes reliably. Dead-source reclaim is covered in unit tests (TestCleanupOrphans) instead, since staging it in e2e was the race.

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: The loop is required because a concurrent mount can swap the entry out from under us; a single "recheck then give up" would abandon a real unmount. The loop just grabs the new correct entry and continues.

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: In V2 it's a hardcoded const too — danglingMountpointCleanupInterval = 2 * time.Minute in pod_unmounter.go ... So we're just matching that...

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: Agreed. Switched to syncing directly with the mount table  — entry.Targets = liveTargets and entry.RefCount = len(liveTargets).

tests/e2e-kubernetes/testsuites/resource_cleanup_daemonset.go
tadiwa-aizen: Went through all 7:

| # | Suggested test | Coverage |
|---|---|---|
| 1 | Multi-consumer refcount reconcile | Unit (`reconciles_refcount`) + existing pod-sharing e2e |
| 2 | Per-volume isolation, dead one reclaimed | Not e2e — see below |
| 3 | Full teardown (cred + meta + source) | Existing pod-sharing e2e + unit |
| 4 | Remount same PV after reclaim | New e2e added, passing |
| 5 | Concurrent unmount vs cleanup tick | Unit (concurrency tests) — too racy for e2e |
| 6 | Cleanup after CSI node restart | `RebuildMountMap` unit + existing restart e2e |
| 7 | Frozen source not reclaimed | N/A — job no longer checks source health (removed to fix churn); decides on live targets only |

**On # 2:** to e2e-test the job reclaiming a dead source, we'd first have to create a dead source with no consumers that still leaves the driver's tracking entry behind (so the periodic job reclaims it, not the normal inline unmount). Staging that reliably needs the test to force an unmount failure inside the node's mount namespace, which our e2e harness can't reach. That's fine because the reclaim logic is fully unit-tested (`TestCleanupOrphans`) — and V2's cleanup job has no e2e either, it's unit-tested only, so we're matching that and adding a safety-property e2e on top.

tests/e2e-kubernetes/testsuites/resource_cleanup_daemonset.go
rajdchak: https://www.kubernetes.dev/blog/2023/04/12/e2e-testing-best-practices-reloaded/ see this doc which says 
```[framework.Gomega()](https://github.com/kubernetes/kubernetes/blob/222f65506252354da012c2e9d5457a6944a4e681/test/e2e/framework/expect.go#L80-L101) offers similar functions as the gomega package, except that they return a normal error instead of failing the test.```

I would suggest use gomega

File: pkg/driver/node/mounter/daemonset_mounter.go (hunk 7 above)
Comment (rajdchak) [reply pkg/driver/node/mounter/daemonset_mounter.go]: I dont think we should remove the cleanup of dead source. What is the impact of churn?

File: tests/e2e-kubernetes/testsuites/resource_cleanup_daemonset.go [hunk 16]
```diff
@@ -0,0 +1,129 @@
...
+		toWrite := 1024
+		seed := time.Now().UTC().UnixNano()
+		checkWriteToPathSucceedEventually(ctx, f, pods[0], "/mnt/volume1/keep.txt", toWrite, seed)
+
+		// Wait for a cleanup pass to run and log that it left THIS volume intact.
+		// The PV name is a fresh UUID, so this line can only come from a cleanup pass
+		// over this test's volume during this run — proving the job actually ran.
+		leftIntactMarker := fmt.Sprintf("volume %s healthy", pvName)
+		ginkgo.By("Waiting for a cleanup pass to run and leave the healthy mount intact")
+		framework.Gomega().Eventually(ctx, func(ctx context.Context) (bool, error) {
```
jet-tong: Rmb to use plain gomega or check returned error as per Rajdeepa's comment on error masking above!

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -171,6 +177,17 @@ func (dm *DaemonsetMounter) Mount(ctx context.Context, bucketName string, target
...
 
+	// Target is mounted but its FUSE source is dead (mounter pod crashed while the workload kept running).
+	// This mount can't recover, so return success without rewriting meta/creds — otherwise every ~60s republish would recreate them just for the cleanup job to delete again (churn).
+	// The periodic cleanup job tears down the dead source, but the workload pod must be recreated before it can get a fresh, working mount.
+	// Only a definitely-dead result short-circuits; "unknown" (transient/timeout) falls through.
+	if err == nil && isMounted {
+		if healthy, healthErr := dm.IsSourceHealthy(ctx, target); healthErr == nil && !healthy {
+			klog.Errorf("DaemonsetMounter: target %s is mounted but dead for volume %s; mount cannot recover, recreate the workload pod to get a fresh mount", target, credentialCtx.VolumeID)
+			return nil
+		}
```
rajdchak: I think we should have a fn like isTargetHealthy similar to isSourceHealthy without the 
```
if errors.Is(err, fs.ErrNotExist) {
			return false, nil
		}
```
I might have said keep the check at the beginning but let's keep this check after the V1 mount check

tests/e2e-kubernetes/testsuites/resource_cleanup_daemonset.go
rajdchak: can you mention which ones have been addressed from the table

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: Removed ... rebased on top of #918, which adds IsTargetHealthy after the V1 mount check; this block is no longer in thePR.

tests/e2e-kubernetes/testsuites/resource_cleanup_daemonset.go
tadiwa-aizen: Added a comment with test coverage

tadiwa-aizen:
Test Coverage

| What we cover | Tests |
|---|---|
| Dead source reclaimed — missing from the mount table, or health probe says dead — tears down source, meta, creds, map entry | Unit: `TestCleanupOrphans` (`dead_source_torn_down`, `dead_source_with_live_target_torn_down`), `TestRebuildMountMap_DeadSourceCleansCredentials`; e2e: `clean up meta file, credentials, and source mount after last consumer unmounts`, and `reclaim a dead source after the mounter pod is killed` (mounter crash → periodic job reclaims) |
| Healthy in-use mount left intact; refcount reconciled to live bind mounts | Unit: `TestCleanupOrphans` (`reconciles_refcount`, `untracked_live_bind_mount_adopted…`); e2e: `cleanup pass without disturbing a healthy in-use mount` |
| Last consumer gone → source reclaimed; same PV remounts after | Unit: `TestCleanupOrphans/last_consumer_gone_torn_down`; e2e: `same PV to mount again after its source is reclaimed` |
| Frozen / undeterminable source left intact (fail-closed) | Unit: `TestIsSourceHealthy_TimeoutIsUnknown` (classification); the `cleanupEntry` leave-on-unknown branch itself isn't directly tested |
| Teardown failure keeps entry + meta for a later retry | Unit: `TestCleanupOrphans/failed_cleanup_keeps_entry_and_meta` |
| Mount-table read error, and mid-creation placeholder → left alone | Unit: `TestCleanupOrphans_MountTableReadError`, `TestCleanupOrphans/skips_placeholder` |
| Concurrency-safe (cleanup and unmount both delete map entries) | Unit: `TestMountMap_ConcurrentDeleteAndGetOrCreate`, `TestMountMap_RetryLoop_ConvergesAfterDelete`; no direct `releaseTarget`-vs-`CleanupOrphans` test |
| Recovery after CSI node restart — map rebuilt from disk | Unit: `RebuildMountMap` tests (dead-source cleanup, live-source recovery, multi-volume); `TestCleanupOrphans` asserts a surviving entry keeps its meta file (restart relies on it); e2e: CSI-node-restart specs |

**Unit-only (can't be reliably staged in e2e):** teardown-failure retry, concurrency (`releaseTarget` vs `CleanupOrphans`), and the frozen/undeterminable-source branch — each needs an injected failure or race our e2e harness can't force. V2's equivalent cleanup (`TestCleanupDanglingMounts`) is unit-only too.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -497,6 +512,113 @@ func (dm *DaemonsetMounter) releaseTarget(target string, volumeID string, creden
...
+	// Being in the mount table doesn't mean the source still works — a crashed mounter leaves
+	// a dead FUSE mount that lingers but errors on every I/O. So we probe it (with a timeout so
+	// a frozen daemon can't stall the pass). If it's dead, tear it down and reclaim its resources;
+	// if we can't tell, leave it alone rather than risk destroying a source still serving a workload.
+	ctx, cancel := context.WithTimeout(context.Background(), cleanupHealthCheckTimeout)
+	healthy, healthErr := dm.IsSourceHealthy(ctx, entry.SourcePath)
+	cancel()
+	switch {
+	case healthErr != nil:
+		klog.V(4).Infof("DaemonsetMounter: cleanup: source health unknown for volume %s (%v), leaving intact", volumeID, healthErr)
```
rajdchak: this can be an error log

pkg/driver/node/mounter/mount_map_persistence_test.go
```diff
@@ -642,3 +650,194 @@ func TestRebuildMountMap_DeadSourceCleansCredentials(t *testing.T) {
...
+			entry := dm.mountMap.Get(volumeID)
+			if tt.expectEntryGone {
+				assert.Equals(t, true, entry == nil)
+			} else {
+				assert.Equals(t, true, entry != nil)
+				assert.Equals(t, tt.expectRefCount, entry.RefCount)
+			}
+			if tt.expectMetaGone {
+				assert.Equals(t, true, os.IsNotExist(statErr(MetaFileName(kubeletPath, volumeID))))
+			}
```
rajdchak, line 831: we should assert on the meta present case as well. That is imp, because when we restart csi node we rely on the meta file to rebuild the map.

tests/e2e-kubernetes/testsuites/resource_cleanup_daemonset.go
```diff
@@ -0,0 +1,131 @@
...
+	"time"
+
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/kubernetes/test/e2e/framework"
+	storageframework "k8s.io/kubernetes/test/e2e/storage/framework"
+	admissionapi "k8s.io/pod-security-admission/api"
+)
+
```
rajdchak, line 15: can we also add a test, to delete mounter pod and then see if the dead source is cleaned?

rajdchak (approved):

PR #916, opened 2026-08-20
Improvements to isHealthyMountpoint check and remove the remaining fr…

rajdchak:
*Description of changes:*
IsHealthyMountpoint / IsSourceHealthy are now tri-state ((bool, error)): healthy, definitely dead (ENOTCONN/corrupted), or unknown (transient error or timeout). Previously any error collapsed to "dead".
mountOrShareSource fails closed on unknown health: it returns an error so kubelet retries, instead of tearing down a source mount that may still be serving live consumers.
e2e tests: replaced framework.Gomega().Eventually(...) with package-level gomega.Eventually(...) in pod_sharing_daemonset.go

### DIFF
--- a/pkg/mountpoint/mounter/mount.go
+++ b/pkg/mountpoint/mounter/mount.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"syscall"
 
 	"k8s.io/klog/v2"
 	mountutils "k8s.io/mount-utils"
@@ -126,12 +127,32 @@ func (m *Mounter) IsMountPoint(target Target) (bool, error) {
 	return m.mount.IsMountPoint(target)
 }
 
-// IsHealthyMountpoint checks whether `target` is both a valid Mountpoint mount AND the FUSE daemon is alive.
+// IsHealthyMountpoint reports whether `target` is a live Mountpoint mount.
+//
+// It returns a three-way result so callers can distinguish "definitely dead"
+// from "couldn't determine", and avoid tearing down a possibly-live source on a
+// transient error:
+//   - (true, nil):  healthy.
+//   - (false, nil): definitely dead — not a Mountpoint mount, corrupted mount,
+//     or the FUSE daemon returned ENOTCONN.
+//   - (false, err): UNKNOWN — a transient error (e.g. failure reading the mount
+//     table, EINTR, fd exhaustion) prevented a determination. Callers MUST NOT
+//     treat this as dead; they should retry later.
+//
 // A dead FUSE mount (after mounter pod crash) stays in the mount table but returns ENOTCONN on any I/O.
-func (m *Mounter) IsHealthyMountpoint(target Target) bool {
+func (m *Mounter) IsHealthyMountpoint(target Target) (bool, error) {
 	isMounted, err := m.CheckMountpoint(target)
-	if err != nil || !isMounted {
-		return false
+	if err != nil {
+		// A corrupted mount is a definite "dead" signal.
+		if m.IsMountpointCorrupted(err) {
+			return false, nil
+		}
+		// Anything else (e.g. couldn't stat or list the mount table) is UNKNOWN.
+		return false, fmt.Errorf("cannot determine mount health for %q: %w", target, err)
+	}
+	if !isMounted {
+		// Not our mount (or not a mount at all) — definitely not a live Mountpoint.
+		return false, nil
 	}
 
 	// Mount entry exists in the table — verify the FUSE daemon is alive.
@@ -142,10 +163,15 @@ func (m *Mounter) IsHealthyMountpoint(target Target) bool {
 	// causing false negatives during transient network issues.
 	f, err := os.Open(target)
 	if err != nil {
-		return false
+		// A dead FUSE daemon returns ENOTCONN — that's a definite "dead".
+		if errors.Is(err, syscall.ENOTCONN) {
+			return false, nil
+		}
+		// Transient errors (EINTR, EMFILE/ENFILE, ENOMEM, ...) are UNKNOWN, not dead.
+		return false, fmt.Errorf("cannot open %q to probe FUSE liveness: %w", target, err)
 	}
 	f.Close()
-	return true
+	return true, nil
 }
 
 // IsMountpointCorrupted returns whether an error returned from [Mounter.CheckMountpoint]
--- a/pkg/mountpoint/mounter/mount.go
+++ b/pkg/mountpoint/mounter/mount.go
@@ -4,8 +4,8 @@ package mounter
 import (
 	"errors"
 	"fmt"
+	"io/fs"
 	"os"
-	"syscall"
 
 	"k8s.io/klog/v2"
 	mountutils "k8s.io/mount-utils"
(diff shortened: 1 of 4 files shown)


tests/e2e-kubernetes/testsuites/pod_sharing_daemonset.go
```diff
@@ -665,7 +665,7 @@ func (t *s3CSIPodSharingDaemonsetTestSuite) DefineTests(driver storageframework.
 			// mount-s3 process should be running while pod is mounted
 			ginkgo.By("Verifying mount-s3 process exists while pod is mounted")
 			dumpMountpointProcesses(ctx, f, targetNode, "while pod mounted")
-			framework.Gomega().Eventually(ctx, func(ctx context.Context) (int, error) {
+			gomega.Eventually(ctx, func(ctx context.Context) (int, error) {
```
jet-tong, line 668: I just realized we have so many instances of `framework.Gomega()...Should(gomega.BeTrue())` in our project. 

Do you know which exact version (should succeed?) is the one that causes error masking and what we'd need to fix for the whole repo to remove the error maskings? Or do we need to fix all framework.Gomega() into plain gomegas?

pkg/mountpoint/mounter/mount.go
```diff
@@ -126,12 +127,32 @@ func (m *Mounter) IsMountPoint(target Target) (bool, error) {
...
+//
 // A dead FUSE mount (after mounter pod crash) stays in the mount table but returns ENOTCONN on any I/O.
-func (m *Mounter) IsHealthyMountpoint(target Target) bool {
+func (m *Mounter) IsHealthyMountpoint(target Target) (bool, error) {
 	isMounted, err := m.CheckMountpoint(target)
-	if err != nil || !isMounted {
-		return false
+	if err != nil {
+		// A corrupted mount is a definite "dead" signal.
+		if m.IsMountpointCorrupted(err) {
```
jet-tong, line 152: From CheckMountpoint last error to handle - "`errors.Is(err, fs.ErrNotExist)` - it means the `target` does not exists, and the caller should create the target folder" - shall we filter this case out too since it means unhealthy Mountpoint as well? (ErrNotExist should mean ENOENT but worth checking; below Open doesn't cover this either but covers ENOTCONN which is definitely dead)

QQ: I guess block is just to handle all CheckMountpoint cases as per CheckMountpoint comment?

pkg/mountpoint/mounter/mount.go
```diff
@@ -126,12 +127,32 @@ func (m *Mounter) IsMountPoint(target Target) (bool, error) {
...
+			return false, nil
+		}
+		// Anything else (e.g. couldn't stat or list the mount table) is UNKNOWN.
+		return false, fmt.Errorf("cannot determine mount health for %q: %w", target, err)
```
tadiwa-aizen, line 156: Hmm quick question ....  If a dead mount makes statx (inside CheckMountpoint) return ENOTCONN, we'd hit this branch and call it UNKNOWN — the ENOTCONN check is only on the os.Open path below. Could a dead source's ENOTCONN surface here too? Want to make sure we don't turn a real "dead" into "unknown" and end up failing NPV instead of remounting.

pkg/mountpoint/mounter/mount.go
```diff
@@ -142,10 +163,15 @@ func (m *Mounter) IsHealthyMountpoint(target Target) bool {
 	// causing false negatives during transient network issues.
 	f, err := os.Open(target)
 	if err != nil {
-		return false
+		// A dead FUSE daemon returns ENOTCONN — that's a definite "dead".
+		if errors.Is(err, syscall.ENOTCONN) {
```
jet-tong: Another observation here here - IsMountpointCorrupted wires to mountutils.IsCorruptedMnt, which already includes ENOTCONN in underlying error there along with other 'definite dead' signals.

I spent a bit of time checking behaviour and what I understand (with some AI help):
- the statx probe only asks kernel on directory attributes and kernel uses cached copy, so it can reply without contacting the daemon
- the open() probe that we added requests an actual fuse opendir action, and it can actually return e.g. ENOTCONN. 

Looking at IsMountpointCorrupted -> IsCorruptedMnt: https://github.com/kubernetes/mount-utils/blob/a2fce2fa31e5ab560a1538c96c5cafae83cece97/mount_helper_unix.go#L46: 
```
// IsCorruptedMnt return true if err is about corrupted mount point
func IsCorruptedMnt(err error) bool {
	if err == nil {
		return false
	}
	var underlyingError error
	switch pe := err.(type) {
	case nil:
		return false
	case *os.PathError:
		underlyingError = pe.Err
	case *os.LinkError:
		underlyingError = pe.Err
	case *os.SyscallError:
		underlyingError = pe.Err
	case syscall.Errno:
		underlyingError = err
	}

	return underlyingError == syscall.ENOTCONN || underlyingError == syscall.ESTALE || underlyingError == syscall.EIO || underlyingError == syscall.EACCES || underlyingError == syscall.EHOSTDOWN || underlyingError == syscall.EWOULDBLOCK
}
```

Conclusion: perhaps we can just re-call IsMountpointCorrupted here on the error, so the function is a clean 2-probe check via statx + open() with the same error handlings?

pkg/mountpoint/mounter/mount.go
jet-tong: `IsMountpointCorrupted` returns true for `syscall.ENOTCONN` as well, so we would `return false, nil` above so ENOTCONN should be handled for both statx probe and open probe.

pkg/mountpoint/mounter/mount.go
rajdchak: yes you are right we should be returning (false, nil) for this errors.Is(err, fs.ErrNotExist)

rajdchak: ```QQ: I guess block is just to handle all CheckMountpoint cases as per CheckMountpoint comment?```

yes

tests/e2e-kubernetes/testsuites/pod_sharing_daemonset.go
rajdchak: Right now I know that replacing framework.Gomega() into plain gomegas prevents the error masking. Some places in the codebase using framework.Gomega() handle the error explicitly while some places in the codebase do not. We have a task to fix this issue for the whole repo

rajdchak: see this doc https://www.kubernetes.dev/blog/2023/04/12/e2e-testing-best-practices-reloaded/  where they explicitly mention this
```[framework.Gomega()](https://github.com/kubernetes/kubernetes/blob/222f65506252354da012c2e9d5457a6944a4e681/test/e2e/framework/expect.go#L80-L101) offers similar functions as the gomega package, except that they return a normal error instead of failing the test.```

So, I don't think this is any version issue. This is how framework.Gomega() behaves and we would need to fix our codebase.

tadiwa-aizen (approved):
LGTM

Two things I checked: a dead ENOTCONN still classifies as dead (unwraps fine through IsCorruptedMnt), so remount isn't regressed; and failing closed on "unknown" instead of tearing down a live source is the right call.

You can follow up on Jensens comments in another PR

Issue #843, opened 2026-06-26
Controller silently stops creating MountpointS3PodAttachments "creation is pending, requeuing"


cawwot:
/kind bug
> **_NOTE:_**  If this is a filesystem related bug, please take a look at the [Mountpoint repo](https://github.com/awslabs/mountpoint-s3) to submit a [bug report](https://github.com/awslabs/mountpoint-s3/issues/new?assignees=&labels=bug&projects=&template=bug-report.yml)

**What happened?**

On a healthy node with ample free capacity, the controller stopped creating MountpointS3PodAttachment (S3PA) objects for that node's workload pods. The controller logged this for every affected workload pod, repeating indefinitely (hours):
`"msg":"MountpointS3PodAttachment creation is pending, requeuing", ... "spec.nodeName":"ip-10-0-136-18.ec2.internal", "pvc":"...-feeds-pvc", ...`

No S3PA and no mountpoint pod were ever created for the node, so all workload pods mounting S3 PVs on it stayed Pending. Other nodes were served normally by the same controller at the same time.

A separate symptom from the same controller appeared concurrently: an S3PA whose mountpoint pod had been deleted was never reconciled, the controller just logged, every ~2 min, and never recreated the pod:
`"msg":"Failed to find Mountpoint Pod - ignoring","s3pa":"s3pa-2cslz"`

Node capacity/scheduling (not #543/#778): node was schedulable, untainted, 16/234 pods used, s3.csi.aws.com registered in CSINode, s3-csi-node pod 3/3 Running.
Stuck/terminating resources: zero S3PAs or mountpoint pods in Terminating/with finalizers for the node.
Workload pod state: pod was legitimately Pending, PodScheduled=True, not terminating.
Controller health: single replica, 0 restarts, no error/warn/panic logs (i.e. not an OOM/crash, distinct from a crashing controller).

**What you expected to happen?**

The controller should create the S3PA + mountpoint pod for any schedulable node with a pending workload pod, or log an actionable error explaining why it cannot.

**How to reproduce it (as minimally and precisely as possible)?**

Does not happen on every node, but at least one node per scale-up event sees this issue. We scale using karpenter based on data ingest which sometimes results in +2000 pods scheduling at the same time. We rate-limit it down to 50 pods/sec, but at nominal pod resources sizes this can result in 150-200 new nodes and 6000 S3 mounts over ~5 mins.

**Anything else we need to know?**:
Workaround:
[...]

PR #917, opened 2026-08-20
fix: clear S3PA expectations in stale cleaner

saikpr:
*Issue #, if available:*

Related to #843.

*Description of changes:*

The controller records an in-memory expectation after it creates a `MountpointS3PodAttachment` (S3PA). It normally clears that expectation when a later Pod reconciliation observes the S3PA in the informer cache.

If the originating workload disappears before that reconciliation, the stale attachment cleaner can be the first component to observe the S3PA. The cleaner can then remove the stale workload and delete the empty S3PA without clearing its creation expectation. Later workloads with the same attachment fields requeue indefinitely because the controller sees no S3PA but still considers its creation pending.

This change reconstructs the expectation fields from every S3PA observed by the stale attachment cleaner. It clears a matching pending expectation before cleanup. Clearing on observation is safe because the object is already visible through the controller client, and it avoids clearing an expectation for a replacement created after deletion.

The tests cover:

- reconstruction of all Pod-authentication expectation fields;
- the create, stale-workload cleanup, S3PA deletion, and expectation-clear sequence.
- expectation clearance when the S3PA retains a valid workload, for driver and Pod authentication.

*Testing:*

- Confirmed the regression test fails for both driver and Pod authentication on unpatched commit `d6f1cce`.
- `go test -race ./cmd/aws-s3-csi-controller/csicontroller`
- `make check_style check_mod_tidy`
- `make test`
- `make bin`
- Confirmed both survival cases fail if expectation clearance moves to S3PA deletion; both deletion cases still pass.


cmd/aws-s3-csi-controller/csicontroller/stale_attachment_cleaner.go
```diff
@@ -95,6 +95,11 @@ func (cm *StaleAttachmentCleaner) RunCleanup(ctx context.Context) error {
 // If S3PodAttachment has no remaining Mountpoint Pods, the entire S3PodAttachment is deleted.
 func (cm *StaleAttachmentCleaner) cleanupStaleWorkloads(ctx context.Context, s3pa *crdv2.MountpointS3PodAttachment, existingPods map[string]*corev1.Pod) error {
 	log := logf.FromContext(ctx).WithValues("s3pa", s3pa.Name)
+	fieldFilters := fieldFiltersForS3PodAttachment(s3pa)
+	if cm.reconciler.s3paExpectations.isPending(fieldFilters) {
+		log.Info("MountpointS3PodAttachment creation is pending, removing from pending")
+		cm.reconciler.s3paExpectations.clear(fieldFilters)
```
tadiwa-aizen, line 102: Could we add a one-line comment here on why it's safe to clear the expectation the moment we see the S3PA, before cleaning it up? (That reasoning is in the PR description but not in the code)

cmd/aws-s3-csi-controller/csicontroller/stale_attachment_cleaner_internal_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+			client, reconciler := newReconcilerWithObjects(t, s3pa, mountpointPod)
+			fieldFilters := reconciler.buildFieldFilters(workloadPod, pv, testCase.workloadRole)
+			reconciler.s3paExpectations.setPending(fieldFilters)
+
+			cleaner := NewStaleAttachmentCleaner(reconciler)
+			err := cleaner.RunCleanup(context.Background())
+			assert.NoError(t, err)
+
+			assertS3PADeleted(t, client, s3pa.Name)
```
tadiwa-aizen: This covers the case where the stale workload is removed and the S3PA is then deleted. Since the S3PA ends up deleted, this test would still pass even if the clear were done at deletion time. Would it be worth adding a case where the S3PA is observed but survives (still has a valid, non-stale workload) and checking the expectation is cleared anyway?

tadiwa-aizen (commented):
Couldn't reproduce the failure mode on my own cluster but I read through the logic and the fix makes sense to me.

cmd/aws-s3-csi-controller/csicontroller/stale_attachment_cleaner.go
saikpr: Addressed in 29e5aa9.

Added the inline comment: the S3PA is already visible in the informer cache, so the creation expectation is satisfied before cleanup.

cmd/aws-s3-csi-controller/csicontroller/stale_attachment_cleaner_internal_test.go
saikpr: Addressed in 29e5aa9.

Added survival cases for driver and Pod authentication. Each case keeps an existing workload Pod, verifies its S3PA and workload attachment survive cleanup, and checks that the creation expectation clears.

Verified the regression with a temporary Go overlay that moves expectation clearance to deletion: both survival cases fail, while both deletion cases pass. The production code passes all four cases.

Validation: controller race tests, `make check_style check_mod_tidy`, and `make bin test` pass with Go 1.27.0.

saikpr:
Both review comments are addressed in 29e5aa9. Local checks pass. Please review and approve the E2E run.

tadiwa-aizen (approved):
LGTM

PR #918, opened 2026-08-24
Do not proceed with mount for dead targets

rajdchak:
*Issue #, if available:*
When the periodic cleanup job in PR https://github.com/awslabs/mountpoint-s3-csi-driver/pull/911 reaps a dead source, kubelet's requiresRepublish keeps calling NodePublishVolume for the now-corrupted target, recreating meta/creds every ~60s only to have cleanup delete them again.

This PR adds an IsTargetHealthy IO probe (mirroring IsSourceHealthy) at the Mount entry point. A corrupted target (ENOTCONN etc) returns nil immediately — no meta, no creds, no churn. An absent target (fresh mount) proceeds normally.

### DIFF
--- a/pkg/driver/node/mounter/daemonset_mounter.go
+++ b/pkg/driver/node/mounter/daemonset_mounter.go
@@ -171,6 +171,20 @@ func (dm *DaemonsetMounter) Mount(ctx context.Context, bucketName string, target
 		return nil
 	}
 
+	// Classify the workload target before doing any mount work.
+	// IsTargetHealthy does an IO probe (same as source health check) with ctx timeout.
+	//   - (true, nil):  target is healthy or absent (fresh) — proceed.
+	//   - (false, nil): dead/corrupted — nothing to do; return nil.
+	//   - (false, err): unknown — let kubelet retry.
+	targetHealthy, targetErr := dm.IsTargetHealthy(ctx, target)
+	if targetErr != nil {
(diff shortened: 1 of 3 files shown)


pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -560,19 +577,53 @@ func (dm *DaemonsetMounter) IsSourceHealthy(ctx context.Context, sourcePath stri
...
+// IsTargetHealthy checks if the workload's bind-mount target is a live Mountpoint mount.
+// Same pattern as [DaemonsetMounter.IsSourceHealthy] (goroutine + ctx timeout, same IO probe),
+// but returns (bool, error) with ErrMountAbsent passed through rather than folded into dead.
+//
+// Callers interpret:
+//   - (true, nil):             target is healthy (already mounted, republish/idempotent).
+//   - (false, ErrMountAbsent): target is absent/fresh — proceed with mount.
+//   - (false, nil):            target mount is dead/corrupted — do NOT proceed (return nil).
+//   - (false, err):            UNKNOWN — let kubelet retry.
+func (dm *DaemonsetMounter) IsTargetHealthy(ctx context.Context, target string) (bool, error) {
```
passaro, line 604: minor: to avoid duplication and clarify each function responsibility, consider refactoring the timeout wrapper into a single function that IsSourceHealthy and IsTargetHealthy can call (and then handle ErrMountAbsent appropriately).

passaro (approved):
The change seems sensible, only a code-tidy suggestion.

PR #927, opened 2026-08-26
Remove duplicate isMountpoint checks in the Mount path

rajdchak:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/pkg/driver/node/mounter/daemonset_mounter.go
+++ b/pkg/driver/node/mounter/daemonset_mounter.go
@@ -720,7 +720,6 @@ const (
 )
 
 // CheckTargetState checks if the workload's bind-mount target is a live Mountpoint mount.
-//
 // Callers interpret:
 //   - (TargetAbsent, nil):  target is absent/fresh — proceed with mount.
 //   - (TargetHealthy, nil): target is healthy and mounted (republish/legacy).
--- a/pkg/driver/node/mounter/daemonset_mounter.go
+++ b/pkg/driver/node/mounter/daemonset_mounter.go
@@ -159,8 +159,7 @@ func NewDaemonsetMounter(clientset kubernetes.Interface, nodeID string, mount *m
 func (dm *DaemonsetMounter) Mount(ctx context.Context, bucketName string, target string,
 	credentialCtx credentialprovider.ProvideContext, args mountpoint.Args, fsGroup string, userEnv envprovider.Environment) error {
 
-	// Check target health and mount state in one call (avoids duplicate IsMountPoint checks).
-	// Dead/corrupted targets are caught immediately — prevents misclassification as V1 systemd.
+	// Check target health
 	//   - (TargetAbsent, nil):  target is absent/fresh — proceed with mount.
 	//   - (TargetHealthy, nil): target is healthy and mounted (republish/legacy).
 	//   - (TargetDead, nil):    target mount is dead/corrupted — return nil.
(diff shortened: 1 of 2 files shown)


pkg/driver/node/mounter/daemonset_mounter.go
passaro: Is this a leftover?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -723,26 +716,56 @@ func (dm *DaemonsetMounter) IsSourceHealthy(ctx context.Context, sourcePath stri
...
 //   - (false, err):            UNKNOWN — let kubelet retry.
-func (dm *DaemonsetMounter) IsTargetHealthy(ctx context.Context, target string) (bool, error) {
+//
+// TargetState represents the health/mount state of a workload target path.
+type TargetState int
+
+const (
+	// TargetAbsent means the target is not mounted (fresh mount, proceed).
+	// iota is a Go constant generator that starts at 0 and increments by 1 for
+	// each subsequent const in the block. So TargetAbsent=0, TargetHealthy=1, TargetDead=2.
```
passaro: Do we need the explanation here?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -723,26 +716,56 @@ func (dm *DaemonsetMounter) IsSourceHealthy(ctx context.Context, sourcePath stri
...
+)
+
+// IsTargetHealthy checks if the workload's bind-mount target is a live Mountpoint mount.
+//
+// Callers interpret:
+//   - (TargetAbsent, nil):  target is absent/fresh — proceed with mount.
+//   - (TargetHealthy, nil): target is healthy and mounted (republish/legacy).
+//   - (TargetDead, nil):    target mount is dead/corrupted — do NOT proceed (return nil).
+//   - (_, err):             UNKNOWN — let kubelet retry.
+func (dm *DaemonsetMounter) IsTargetHealthy(ctx context.Context, target string) (TargetState, error) {
```
passaro: Consider renaming now that we are not returning a bool.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -723,26 +716,56 @@ func (dm *DaemonsetMounter) IsSourceHealthy(ctx context.Context, sourcePath stri
...
-		ch <- result{healthy, err}
+		if err != nil {
+			ch <- result{TargetDead, err} // transient/unknown error
+			return
+		}
+		if healthy {
+			ch <- result{TargetHealthy, nil}
+		} else {
+			ch <- result{TargetDead, nil}
+		}
```
passaro, line 748: Consider something like this:

```suggestion
state := TargetHealthy
if !healthy || err != nil {
    state = TargetDead // transient/unknown error is treated as dead
}
ch <- result{state, err}
```

rajdchak: Makes sense as a simplification. I have a slight preference for the explicit branches since the caller treats (TargetDead, err) vs (TargetDead, nil) differently, and the separate paths make that distinction easier to spot

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: removed it

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: done

passaro (dismissed):

passaro (approved):

PR #929, opened 2026-08-26
Add user agent to daemonset mounter approach

tadiwa-aizen:
*Issue #, if available:*

*Description of changes:*
Adds the Mountpoint user-agent to the V3 daemonset mounter. V2 (`PodMounter`) already
stamps `mount-s3` requests with a telemetry string (`s3-csi-driver/<ver> credential-source#<src>
k8s/<ver> md/install#<method>`); V3 didn't, so V3 mounts were untagged. This brings V3 to parity.

How: give `DaemonsetMounter` the `kubernetesVersion` and `variant` it was missing (passed in
from `driver.go`), stop discarding the resolved auth source from `provideCredentials`, and set
`--user-agent-prefix` inside `fuseMount`. Setting it there — after the pod-sharing options are
snapshotted — keeps the user-agent out of the sharing compatibility key and the persisted meta.

### Testing
Unit tests
Extended the daemonset "Correctly passes mount options" test to assert the args sent to
`mount-s3` include `--user-agent-prefix=<UserAgent(...)>`, built from the same inputs.

Situations covered for the V3 user-agent:

| Situation | Covered by |
|---|---|
| V3 mount actually sets `--user-agent-prefix` and sends it to `mount-s3` | `daemonset_mounter_test.go` — "Correctly passes mount options" (this PR) |
| String format: `credential-source#driver` vs `#pod` | `user_agent_test.go` — `TestUserAgent` (pre-existing) |
| String format: k8s version present vs empty (segment dropped) | `user_agent_test.go` — `TestUserAgent` (pre-existing) |
| String format: OpenShift variant + install method (helm / eks-addon / kustomize / unknown) | `user_agent_test.go` — `TestUserAgent` (pre-existing) |

Manual tests
Deployed to an EKS cluster in daemonset mode, mounted a volume, and confirmed the live
`mount-s3` process's cmdline contained `--user-agent-prefix=s3-csi-driver/2.7.0
credential-source#driver k8s/v1.36.2-eks-... md/install#helm`.

No e2e test was added — the user-agent is unit-tested only, matching V2's testign coverage

### DIFF
--- a/pkg/driver/node/mounter/daemonset_mounter.go
+++ b/pkg/driver/node/mounter/daemonset_mounter.go
@@ -39,6 +39,7 @@ import (
 	"k8s.io/klog/v2"
 	mountutils "k8s.io/mount-utils"
 
+	"github.com/awslabs/mountpoint-s3-csi-driver/pkg/cluster"
 	"github.com/awslabs/mountpoint-s3-csi-driver/pkg/driver/node/credentialprovider"
 	"github.com/awslabs/mountpoint-s3-csi-driver/pkg/driver/node/envprovider"
 	"github.com/awslabs/mountpoint-s3-csi-driver/pkg/driver/node/targetpath"
@@ -101,6 +102,11 @@ type DaemonsetMounter struct {
 	mount        *mpmounter.Mounter
 	credProvider credentialprovider.ProviderInterface
 
+	// kubernetesVersion and variant identify the cluster; used to build the Mountpoint
+	// user-agent so S3-side telemetry sees which cluster the request came from.
+	kubernetesVersion string
+	variant           cluster.Variant
+
 	// Comm dir discovery: commDir caches the path (nil = stale),
 	// rediscoverCh wakes the background watcher to re-discover immediately.
 	commDir      atomic.Pointer[string]
@@ -120,13 +126,15 @@ type DaemonsetMounter struct {
 // mountSyscall, bindMountSyscall, and mountInfoProvider may be nil,
 // in which case the default implementations are used.
 func NewDaemonsetMounter(clientset kubernetes.Interface, nodeID string, mount *mpmounter.Mounter,
-	credProvider credentialprovider.ProviderInterface, mountSyscall mountSyscallFunc, bindMountSyscall bindMountSyscallFunc, mountInfoProvider mountInfoProviderFunc) *DaemonsetMounter {
+	credProvider credentialprovider.ProviderInterface, mountSyscall mountSyscallFunc, bindMountSyscall bindMountSyscallFunc, mountInfoProvider mountInfoProviderFunc, kubernetesVersion string, variant cluster.Variant) *DaemonsetMounter {
 	return &DaemonsetMounter{
 		clientset:         clientset,
 		nodeID:            nodeID,
 		kubeletPath:       util.ContainerKubeletPath(),
 		mount:             mount,
 		credProvider:      credProvider,
+		kubernetesVersion: kubernetesVersion,
+		variant:           variant,
 		rediscoverCh:      make(chan struct{}, 1),
 		mountSyscall:      mountSyscall,
 		bindMountSyscall:  bindMountSyscall,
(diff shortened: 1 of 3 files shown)


pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -281,7 +289,7 @@ func (dm *DaemonsetMounter) mountOrShareSource(ctx context.Context, bucketName s
 	// Provision credentials under the lock. We always use entry.CommDir which is set above
 	// (either from an existing healthy entry, or freshly assigned from commDir on new mount).
 	// This ensures credentials are written to the same location that cleanup will look at.
-	credsEnv, err := dm.provideCredentials(ctx, entry.CommDir, volumeID, &credentialCtx)
+	credsEnv, authSource, err := dm.provideCredentials(ctx, entry.CommDir, volumeID, &credentialCtx)
```
jet-tong, line 289: personal preference np: we could follow pod_mounter.go and return error here (instead of in provideCredentials function). We could also add `helpMessageForGettingMounterLogs` there too.

```
	credEnv, authenticationSource, err := pm.provideCredentials(ctx, podPath, string(pod.UID), s3PodAttachment.Spec.WorkloadServiceAccountIAMRoleARN, credentialCtx)
	if err != nil {
		klog.Errorf("Failed to provide credentials for %q: %v. %s", source, err, pm.helpMessageForGettingMountpointLogs(pod))
		return fmt.Errorf("Failed to provide credentials for %q: %w. %s", source, err, pm.helpMessageForGettingMountpointLogs(pod))
	}
```

pkg/driver/node/mounter/daemonset_mounter_test.go
```diff
@@ -502,6 +505,7 @@ func TestDaemonsetMounter(t *testing.T) {
 				},
 				nil,
 				nil,
+				"", cluster.DefaultKubernetes,
```
jet-tong: np - make it 2 lines to match style.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -345,8 +353,9 @@ func (dm *DaemonsetMounter) mountOrShareSource(ctx context.Context, bucketName s
 // fuseMount performs the FUSE mount + FD send + wait cycle at the given path.
 // Credentials are already provisioned by the caller (Mount).
 // commDir is passed from the caller to ensure a single GetCommDir() per NodePublishVolume.
+// authSource is the resolved AuthenticationSource returned by the credential provider
```
jet-tong: np missing full stop, but maybe we don't need this line at all since it's a minor comment (commDir one is more important).

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -370,6 +379,9 @@ func (dm *DaemonsetMounter) fuseMount(ctx context.Context, bucketName string, mo
 
 	args.Remove(mountpoint.ArgReadOnly)
 
+	// We set the user-agent here so it doesn't get captured into the pod-sharing key or persisted meta.
```
jet-tong: qq: set user-agent here opposed to where? Before `incomingParams := MountParams{`? (If yes perhaps worth saying this order matters in comment)

Edit - AI suggested: "add `func (a *Args) Clone() Args { return Args{a.args.Clone()} }` and `args = args.Clone()` at the top of fuseMount" if we don't want to mutate Arg in fuseMount, if that's what we want here.

tadiwa-aizen: Updated the comment to say the order matters. Didn't add Clone(): SortedList() already copies, so the key is frozen before fuseMount mutates args, and V2 mutates in place too.

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: Addressed. Moved the wrap to the call site and appended helpMessageForGettingMounterLogs(); provideCredentials now returns the raw error.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -848,7 +845,7 @@ func (dm *DaemonsetMounter) provideCredentials(ctx context.Context, commDir, vol
 
 	env, authSource, err := dm.credProvider.Provide(ctx, *credentialCtx)
 	if err != nil {
-		return nil, "", fmt.Errorf("failed to provide credentials for mount %s: %w", volumeID, err)
+		return nil, "", err
 	}
 	return env, authSource, nil
```
jet-tong, line 850: small np: maybe you don't need that anymore and you can just return dm.credProvider.Provide?

jet-tong (approved):

Issue #787, opened 2026-05-04
`kubectl drain` requires `--force` to evict `mount-s3` pods (no controller owner)


sam-stenvall-iceye:
**What happened?**

Draining a node with active `mount-s3` pods on it isn't possible, you'll get `Error draining machine foo: cannot delete cannot delete Pods that declare no controller (use --force to override): mount-s3/mp-848kh, mount-s3/mp-sbpd4`.

We thought the issue was fixed in 2.5.0 (similar to https://github.com/awslabs/mountpoint-s3-csi-driver/issues/675), but it seems it was only fixed for the narrow use case where autoscaling is used.

**What you expected to happen?**

Pods to be evicted in normal order

**How to reproduce it (as minimally and precisely as possible)?**

Just drain any node with `mount-s3` pods running

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`): v1.32.11+rke2r1
- Driver version: 2.5.0

cnemo-cenic:
I think #790 might be a fix for this? Or perhaps just part of one.

sam-stenvall-iceye:
Assuming it works the way I think (a DaemonSet running one pod that handles all the mounting) then yes, the `--ignore-daemonsets` option would then ignore these and the drain could continue.

vladem:
Hey @sam-stenvall-iceye, thanks for reporting this. We acknowledge the issue, the `mount-s3/mp-*` pods are created without a controller owner, which causes `kubectl drain` to refuse evicting them without `--force`.

Could you help us understand if this is blocking for you, or an operational inconvenience? Specifically — is there a reason you can't use `--force` in your drain workflow?

From the S3 CSI Driver's perspective, using `--force` is safe here. Mountpoint pods [ignore SIGTERM and keep serving volumes](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/docs/TROUBLESHOOTING.md#my-workload-pods-are-getting-transport-endpoint-is-not-connected-errors-during-node-drain) until workload pods using them terminate, ensuring correct shutdown order within the 10-minute grace period.

We're working on a [DaemonSet-based mounter architecture](https://github.com/awslabs/mountpoint-s3-csi-driver/issues/543#issuecomment-4305014764) that will resolve this — Mountpoint-processes will run inside an actual DaemonSet pod, so `kubectl drain --ignore-daemonsets` should handle it without `--force`. We'll update this issue when it's released.

[...]

PR #930, opened 2026-08-27
Document kubectl drain failure on Mountpoint pods

tadiwa-aizen:
*Issue #, if available:* #787

*Description of changes:*

Adds a troubleshooting entry for the `kubectl drain` failure on controller-less Mountpoint pods (#787): the `mp-*` pods have no owning controller, so `kubectl drain` refuses them without `--force` (`cannot delete Pods that declare no controller`).

The entry explains the `--force` workaround (safe because the Mountpoint pods keep serving the volume during their grace period), and notes that `--force` does not bypass PodDisruptionBudgets — a restrictive PDB can still stall the drain, and force-deleting those pods can interrupt the application.

**_Testing_**:
I tested manually and confirmed the behaviour

Docs-only change; no behavior change.

### DIFF
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -145,6 +145,24 @@ This error occurs when workload pods are terminated before the Mountpoint pod th
 
 **Recommendation:** Keep workload pod termination grace periods sufficiently low to ensure graceful shutdown completes within the Mountpoint pod's 10-minute window and the grace period specified in the `drain` operation (e.g. via Karpenter NodePool [settings](https://karpenter.sh/docs/concepts/nodepools/#spectemplatespecterminationgraceperiod)).
 
+## My node drain fails with "cannot delete Pods that declare no controller"
+
+`kubectl drain` refuses to evict the Mountpoint pods, failing with:
+
+```
+cannot delete Pods that declare no controller (use --force to override): mount-s3/mp-xxxxx
+```
+
+The Mountpoint pods (`mp-*` in the `mount-s3` namespace) have no owning controller, so `kubectl drain` won't evict them without `--force`. Passing `--force` is safe — the Mountpoint pods keep serving the volume during their grace period so workloads shut down first (see ["Transport endpoint is not connected"](#my-workload-pods-are-getting-transport-endpoint-is-not-connected-errors-during-node-drain) above):
+
+```
+kubectl drain <node> --ignore-daemonsets --delete-emptydir-data --force
+```
+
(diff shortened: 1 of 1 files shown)


docs/TROUBLESHOOTING.md
```diff
@@ -145,6 +145,22 @@ This error occurs when workload pods are terminated before the Mountpoint pod th
...
+`kubectl drain` refuses to evict the Mountpoint pods, failing with:
+
+```
+cannot delete Pods that declare no controller (use --force to override): mount-s3/mp-xxxxx
+```
+
+The Mountpoint pods (`mp-*` in the `mount-s3` namespace) have no owning controller, so `kubectl drain` won't evict them without `--force`. Passing `--force` is safe — the Mountpoint pods keep serving the volume during their grace period so workloads shut down first (see ["Transport endpoint is not connected"](#my-workload-pods-are-getting-transport-endpoint-is-not-connected-errors-during-node-drain) above):
+
+```
+kubectl drain <node> --ignore-daemonsets --delete-emptydir-data --force
```
muddyfish, line 159: Why is `--delete-emptydir-data` needed? It's not mentioned in the text

docs/TROUBLESHOOTING.md
```diff
@@ -145,6 +145,22 @@ This error occurs when workload pods are terminated before the Mountpoint pod th
...
+```
+
+Note that `--force` does **not** bypass PodDisruptionBudgets. If a workload's PDB is too restrictive (for example `maxUnavailable: 1` with many pods per node), the drain can still stall with `cannot evict pod as it would violate the pod's disruption budget`, and force-deleting those pods to proceed can interrupt your application.
```
muddyfish: This text should be improved. The last line says and force-deleting those pods to proceed can interrupt your application.' which contradicts the rest of the text

docs/TROUBLESHOOTING.md
tadiwa-aizen: It's needed because [`kubectl drain` won't evict a pod using an `emptyDir`](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#drain) and Mountpoint pods use an [in-memory `emptyDir`](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/e9031e3542fb421ec5a7242a2359a24043b6c6c1/pkg/podmounter/mppod/creator.go#L191-L199) (`/comm`) — I've added this to the text.

docs/TROUBLESHOOTING.md
tadiwa-aizen: Addressed.

muddyfish (approved):

PR #939, opened 2026-09-04
V2 to V3 upgrade support

rajdchak:
*Issue #, if available:*

*Description of changes:*

```
## Controller: drain-only mode

- Adds `DrainStaleAttachmentCleaner` — **the existing `StaleAttachmentCleaner` with a few changes**, kept separate so we can delete the V2 `reconciler.go` + `stale_attachment_cleaner.go` later with no impact:
  - Self-contained (holds only an API client + mountpoint namespace, no dependency on the V2 `Reconciler`).
  - Adds a Succeeded-Mountpoint-Pod deletion sweep (previously done by the reconcile loop, which drain-only mode doesn't have).
  - Headroom-pod cleanup runs without the feature-flag gate.
- In daemonset mode (`MOUNTER_MODE=daemonset`), `main.go` registers **only** this cleaner — the controller never spawns Mountpoint Pods, creates S3PAs, or manages headroom.
- Each pass: prune stale workload UIDs from S3PAs → annotate emptied Mountpoint Pods `needs-unmount` → delete emptied S3PAs; delete Succeeded Mountpoint Pods; delete orphaned V2 Headroom Pods.


## Node: legacy V2 detection + credential refresh

- `DaemonsetMounter` detects legacy V2 mounts and refreshes their credentials (IRSA role session + SA token rotation) so long-lived V2 mounts survive the upgrade window. The committed IAM role ARN / Mountpoint Pod name is resolved via an S3PA informer cache.
- V3 source mounts now live under `.../s3.csi.aws.com/v3mnt/<volumeID>` (siblings of V2's `.../mnt/`), so V2 vs V3 mounts are distinguished by directory, not naming heuristic.
- Under `SUPPORT_LEGACY_POD_MOUNTS=true`, the node starts the existing `PodUnmounter` (acts on the `needs-unmount` annotation) and the S3PA cache. End-to-end: cleaner annotates → node unmounts → pod goes Succeeded → cleaner deletes.

## Helm chart

- `controller.yaml`: the controller Deployment is now created in **all** modes (was skipped in daemonset mode, so the drain-only controller was never deployed); sets `MOUNTER_MODE=daemonset`.
- `serviceaccount-csi-controller.yaml`: RBAC tightened in daemonset mode (drops pod `create`, S3PA `create`, and PVC/PV/ServiceAccount reads used only on the spawn path). 
```


cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
```diff
@@ -0,0 +1,268 @@
...
+		return err
+	}
+
+	if mpPod.Annotations == nil {
+		mpPod.Annotations = make(map[string]string)
+	}
+	mpPod.Annotations[mppod.AnnotationNeedsUnmount] = "true"
+
+	// Update the pod
+	err = cm.Update(ctx, mpPod) // TODO: This probably needs to be a patch as we might've get a stale Mountpoint Pod.
```
muddyfish, line 231: TODO that's still in the code.

tests/e2e-kubernetes/testsuites/upgrade_and_rollback.go
```diff
@@ -217,11 +217,11 @@ func (t *s3CSIUpgradeTestSuite) DefineTests(driver storageframework.TestDriver,
 		path := filepath.Join(e2epod.VolumeMountPath1, filename)
 		testWriteSize := 1024 // 1KB
 
-		checkWriteToPathSucceedEventually(ctx, f, pod, path, testWriteSize, seed)
-		checkReadFromPathSucceedEventually(ctx, f, pod, path, testWriteSize, seed)
-		checkListingPathWithEntriesEventually(ctx, f, pod, e2epod.VolumeMountPath1, []string{filename, "test.txt"})
+		checkWriteToPathSucceed(ctx, f, pod, path, testWriteSize, seed)
```
muddyfish, line 220: Why did these change?

charts/aws-mountpoint-s3-csi-driver/templates/serviceaccount-csi-node.yaml
```diff
@@ -81,7 +81,7 @@ roleRef:
   kind: Role
   name: s3-csi-driver-node-daemonset-mounter-role
   apiGroup: rbac.authorization.k8s.io
-{{- else }}
+{{- end }}
```
muddyfish, line 84: Why was this changed?

charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
```diff
@@ -98,6 +98,8 @@ spec:
               value: {{ .Values.node.kubeletPath }}
             - name: SUPPORT_LEGACY_SYSTEMD_MOUNTS
               value: "{{ .Values.supportLegacySystemDMounts }}"
+            - name: SUPPORT_LEGACY_POD_MOUNTS
+              value: "{{ .Values.supportLegacyPodMounts }}"
```
muddyfish, line 102: We should have a check that if `.Values.experimental.mounterMode` is not `daemonset`, `supportLegacyPodMounts` is `true`

charts/aws-mountpoint-s3-csi-driver/values.yaml
```diff
@@ -135,6 +135,11 @@ eksPodIdentityAgent:
 # TODO: Remove this in v3 as systemd mount support will be discontinued
```
muddyfish, line 135: This should be removed as we're continuing to support it for now

muddyfish (commented):
Do we have any system for emitting metrics? It would be useful to emit a metric for 'v2 -> v3 migration complete' and similarly for v1 -> v3 so you could work out when to disable the legacy helpers.

tests/e2e-kubernetes/testsuites/upgrade_and_rollback.go
rajdchak: In our main branch today we have these verifications without Eventually so I thought of reverting back to original checks.

charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
rajdchak: Is not daemonset or is daemonset?

charts/aws-mountpoint-s3-csi-driver/templates/serviceaccount-csi-node.yaml
rajdchak: In the daemonset mode we need this s3-csi-driver-node-mountpoint-pod-namespace-role because to refresh credentials the driver needs to access the mp pods.

charts/aws-mountpoint-s3-csi-driver/values.yaml
rajdchak: will address this in next revision

charts/aws-mountpoint-s3-csi-driver/templates/node.yaml
muddyfish: If we're not using daemonset mode, we need to support pod mount mode

rajdchak: We never read this flag in the pod mounter mode code path? I dont think we need to have this check at all.

jet-tong: I think Simon assumed it means 'support pod mount' in general. Perhaps `maintainLegacyPodMounts` makes more sense?
(nevermind this matches the supportLegacySystemDMounts in v2 code)

charts/aws-mountpoint-s3-csi-driver/values.yaml
```diff
@@ -135,6 +135,11 @@ eksPodIdentityAgent:
...
 supportLegacySystemDMounts: true
 
+# Set to true to enable credential refresh for legacy V2 (pod-mounter) mounts
+# after upgrading to V3 (daemonset mounter). This ensures V2 Mountpoint Pods
+# receive refreshed credentials during the transition period.
+supportLegacyPodMounts: true
```
jet-tong, line 141: QQ: Why do we want supportLegacySystemDMounts / supportLegacyPodMounts flags at all as customer facing config? Why not have them be true for all cases? (I get the benefit of not having controller for instance + this matches what we had in supportLegacySystemDMounts, but not sure we want to complicate customer config even more if no one is gonna set that to false anyways).

charts/aws-mountpoint-s3-csi-driver/values.yaml
jet-tong: > CSI Driver v2 only:

This will also need an edit (maybe align with comment below)

cmd/aws-s3-csi-controller/main.go
```diff
@@ -110,3 +130,11 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+// isDrainOnlyMode returns true when the controller is running in daemonset (V3) mounter mode.
+// In this mode the controller runs only the DrainStaleAttachmentCleaner: it periodically cleans up
+// S3PodAttachments for pods that no longer exist and deletes completed Mountpoint Pods and leftover
+// Headroom Pods — it never spawns Pods.
+func isDrainOnlyMode() bool {
+	return os.Getenv("MOUNTER_MODE") == "daemonset"
```
jet-tong, line 139: np: add a var line above to store environment variables along with other GetEnv

```
var mountpointNamespace = flag.String("mountpoint-namespace", os.Getenv("MOUNTPOINT_NAMESPACE"), "Namespace to spawn Mountpoint Pods in.")
```

Maybe we don't need this function at all, just check mounter mode above and note we're using drain-only mode with NewDrainStaleAttachmentCleaner perhaps?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -194,16 +219,19 @@ func (dm *DaemonsetMounter) Mount(ctx context.Context, bucketName string, target
...
-	if err != nil {
-		return fmt.Errorf("failed to parse target path %q: %w", target, err)
+		// V2 (pod-mounter) check: the target's FUSE source lives under the V2 source dir ("mnt"),
+		// whereas V3 sources live under "v3mnt". If the target bind-mounts from a source under the
+		// V2 dir, it's a legacy V2 mount and we refresh its credentials via the V2 path.
+		isV2, err := dm.isLegacyV2Mount(target)
+		if err != nil {
+			return fmt.Errorf("failed to determine if target %q is a legacy V2 mount for volume %s, will retry: %w", target, credentialCtx.VolumeID, err)
+		}
+		if isV2 {
```
jet-tong, line 230: What happens if !isV2? (It would mean it's not v1 / v2 / v3?) Should we warn users about it?

Edit: Noted below if supportLegacyPodMount is disabled, isV2 would be false also, which means ignoring it and doing nothing.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -833,6 +861,172 @@ func (dm *DaemonsetMounter) isSystemDMountpoint(target string) bool {
...
+//
+// It returns true when the target bind-mounts (same device ID) from a source under the V2 dir.
+// Requires SUPPORT_LEGACY_POD_MOUNTS to be enabled.
+//
+// A non-nil error indicates the mount table could not be read — callers should propagate it so
+// kubelet retries, rather than treating it as "not a V2 mount" (which would risk a wrong V3 mount
+// attempt over an existing V2 mount).
+func (dm *DaemonsetMounter) isLegacyV2Mount(target string) (bool, error) {
+	if !util.SupportLegacyPodMounts() {
+		return false, nil
```
jet-tong, line 895: If users disable supportLegacyPodMounts, do we really want to ignore it and do nothing, or could we warn customers perhaps?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -833,6 +861,172 @@ func (dm *DaemonsetMounter) isSystemDMountpoint(target string) bool {
...
+	targetDevID := deviceID(targetMI)
+	// V2 sources live under the legacy "mnt" dir; V3 sources live under the sibling "v3mnt" dir.
+	v2SourceMountDir := SourceMountDir(dm.kubeletPath)
+
+	for i := range mountInfos {
+		mi := &mountInfos[i]
+		if deviceID(mi) != targetDevID || mi.MountPoint == target {
+			continue
+		}
+		if strings.HasPrefix(mi.MountPoint, v2SourceMountDir) {
```
jet-tong, line 917: Noted mnt-v3 would probs match here so v3mnt is okay (I was thinking why not mnt-v3 or mntv3 as name)

tests/e2e-kubernetes/testsuites/drain_only_daemonset.go
```diff
@@ -0,0 +1,257 @@
...
+					return false, err
+				}
+				return false, nil
+			}).WithTimeout(drainAssertionTimeout).WithPolling(drainAssertionPolling).Should(gomega.BeTrue(),
+				"expected drain-only cleaner to delete the emptied MountpointS3PodAttachment")
+
+			framework.Logf("Drain-only cleaner successfully drained stale S3PA %q", s3paName)
+		})
+
+		ginkgo.It("should delete a leftover Headroom Pod whose workload no longer exists", func(ctx context.Context) {
```
jet-tong, line 188: May I ask how the headroom pod logic works with the cleaner?

I'm aware it goes headroom pod -> workload pod -> mp pod replacing headroom pod, but worried there might be edge cases where workload pods are pending scheduling in v2, and then we perform upgrade to v3. Edge case if we don't ungateHeadroomSchedulingGateForWorkloadPod (reconciler) maybe so we have to add that function in?

tests/e2e-kubernetes/testsuites/upgrade_and_rollback.go
jet-tong: A bunch of them changed [here](https://github.com/awslabs/mountpoint-s3-csi-driver/pull/905/changes#diff-b2d94067a81af228062bc87092366952a3c5e36a73f61b373cdae477c90644e0), good that we don't need eventually checks actually.

tests/e2e-kubernetes/testsuites/drain_only_daemonset.go
```diff
@@ -0,0 +1,257 @@
...
+	crdv2 "github.com/awslabs/mountpoint-s3-csi-driver/pkg/api/v2"
+	"github.com/awslabs/mountpoint-s3-csi-driver/pkg/podmounter/mppod"
+)
+
+// Drain-only cleaner timing (mirrors cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go).
+// The cleaner runs every drainCleanupInterval and only prunes attachments older than
+// drainStaleAttachmentThreshold. We stamp the synthetic attachment well in the past so it is
+// immediately eligible, then wait out (interval + threshold + slack).
+const (
+	drainCleanerInterval  = 2 * time.Minute
```
jet-tong, line 31: drainCleanupInterval perhaps? (unused constant)

tests/e2e-kubernetes/testsuites/util.go
```diff
@@ -362,6 +362,44 @@ func waitForKubernetesObjectToDisappear[T any](ctx context.Context, get framewor
...
+	}
+	return len(names), names, nil
+}
+
+// waitForMountpointPodsDeleted waits until the number of Mountpoint Pods in the Mountpoint namespace
+// drops to expectedRemaining, or fails after timeout. Use this after terminating workloads to assert
+// that their corresponding Mountpoint Pods actually drain out (annotated needs-unmount -> unmounted ->
+// Succeeded -> deleted). Drain in daemonset mode is periodic (cleanup interval + staleness threshold),
+// so the timeout must be generous.
+func waitForMountpointPodsDeleted(ctx context.Context, f *framework.Framework, expectedRemaining int, timeout time.Duration) {
```
jet-tong, line 387: Unused function? countMountpointPods function above as well.

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+// NewDrainStaleAttachmentCleaner creates a DrainStaleAttachmentCleaner. It needs the API client and
+// the Mountpoint namespace (used to identify Mountpoint/Headroom Pods); drain-only mode never spawns
+// pods so it needs no image/priority-class config.
+func NewDrainStaleAttachmentCleaner(c client.Client, mountpointNamespace string) *DrainStaleAttachmentCleaner {
+	return &DrainStaleAttachmentCleaner{Client: c, mountpointNamespace: mountpointNamespace}
+}
+
+// Start runs the periodic cleanup loop until the context is cancelled. Satisfies manager.Runnable.
```
jet-tong, line 55: >Satisfies manager.Runnable.

May I ask what this means? ('context' comment is new from stale_attachment_cleaner, perhaps means it satisfies a contract?

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
```diff
@@ -0,0 +1,268 @@
...
+// spawn/create controller (reconciler.go) and its cleaner (stale_attachment_cleaner.go) can both be
+// deleted without touching drain-only mode.
+type DrainStaleAttachmentCleaner struct {
+	client.Client
+	mountpointNamespace string
+}
+
+// NewDrainStaleAttachmentCleaner creates a DrainStaleAttachmentCleaner. It needs the API client and
+// the Mountpoint namespace (used to identify Mountpoint/Headroom Pods); drain-only mode never spawns
+// pods so it needs no image/priority-class config.
```
jet-tong, line 50: This means we need to keep
```
mountpointPod:
  namespace: mount-s3
```
in values.yaml - should we add a comment in values.yaml in that place to note that v3 still needs it, so we don't forget and delete it when removing v2 code?

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	// Create a map of existing pod UIDs for quick lookup
+	existingPods := make(map[string]*corev1.Pod)
+	for _, pod := range podList.Items {
+		existingPods[string(pod.UID)] = &pod
+	}
+
+	// Always attempt Headroom Pod cleanup — in drain-only mode there is no feature-flag gate; any
+	// leftover V2 Headroom Pods must be removed once their workload is gone or past scheduling.
+	if err := cm.cleanupStaleHeadroomPods(ctx, existingPods); err != nil {
```
jet-tong, line 93: Noted in version before we gate on `isHeadroomForMountpointPodsFeatureEnabled` (which the comment also says) - does it mean in v2, if we update helm to disable headroom pod feature, there could be stale headroom pods that we didn't cleanup? (Any consequences of having this cleanup but no headroom.go running?)

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
```diff
@@ -0,0 +1,268 @@
...
+		log.Error(err, "Error cleaning up stale Headroom Pods")
+	}
+
+	// Delete completed (Succeeded) V2 Mountpoint Pods. These never get a new workload in drain-only
+	// mode, so once they've cleanly unmounted and exited there is nothing left to keep them around.
+	if err := cm.cleanupSucceededMountpointPods(ctx, existingPods); err != nil {
```
jet-tong, line 99: Noted this originally belonged to reconciler.go reconcileMountpointPod - do you know if it's safe to do it in a 2 min loop instead of the quick pod event based updates, and why?

Edit: Asked AI, they think _"The drain chain now spans two sweeps. Pass N prunes the attachment, annotates, and deletes the S3PA; the node unmounts in seconds; the pod goes Succeeded; pass N+1 deletes it. So the S3PA is deleted before the pod, inverting v2's ordering. Harmless — but it's the same window that makes the S3PA-missing branch in handleLegacyV2Refresh reachable"_ ; maybe worth looking into to make sure we don't bring wrong code over from the already delicate reconciler code. 

Relevant code in reconciler.go:
```
	case corev1.PodSucceeded:
		err := r.deleteMountpointPod(ctx, pod)
		if err != nil {
			log.Error(err, "Failed to delete succeeded Pod")
			return reconcile.Result{}, err
		}
		log.Info("Pod succeeded and successfully deleted")
```

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
```diff
@@ -0,0 +1,268 @@
...
+//     annotates emptied Mountpoint Pods with `needs-unmount`, and deletes emptied S3PAs.
+//   - Deletes completed (Succeeded) V2 Mountpoint Pods.
+//   - Deletes leftover V2 Headroom Pods once their workload is gone or past scheduling.
+//
+// It never spawns new Mountpoint Pods, creates/updates S3PAs for active workloads, or creates
+// Headroom Pods — those responsibilities belong to V3's DaemonsetMounter, not the controller.
+//
+// This type is intentionally SELF-CONTAINED: it depends only on the API client, so the V2
+// spawn/create controller (reconciler.go) and its cleaner (stale_attachment_cleaner.go) can both be
+// deleted without touching drain-only mode.
```
jet-tong, line 42: One big nitpick on review ease would be that since both code paths refactored into 1 file, then with some logic updates, made it harder to review. I find myself diffing in VSCode the drain / non-drain versions to find out what changed (not very effectively since I'm not familiar with this part of the code)

Perhaps a commit to just copy stale_attachment_cleaner.go and then next commit to update it (while minimizing changes) would make it more obvious what has changed and what was not consumed. Some comments in original version wasn't in the new version too so it'd make that more visible too.

Similar note for drain_stale_attachment_cleaner_test vs original tests.

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
jet-tong: This is pre-existing in reconciler.go `(r *Reconciler) addNeedsUnmountAnnotation` - wasn't obvious on first review for me too.

tests/e2e-kubernetes/testsuites/drain_only_daemonset.go
```diff
@@ -254,6 +244,31 @@ func (t *s3CSIDrainOnlyDaemonsetTestSuite) DefineTests(driver storageframework.T
...
+// The Mountpoint namespace enforces the "restricted" Pod Security Standard, so the synthetic pod
+// must set runAsNonRoot, a seccomp profile, and (in the container) drop ALL capabilities.
+//
+// OpenShift is special: its SecurityContextConstraints assign a non-root UID from a per-namespace
+// range and REJECT a hardcoded runAsUser outside that range — not just at create time, but on every
+// update. The drain-only cleaner updates the Mountpoint Pod (to add the needs-unmount annotation)
+// while draining, so a hardcoded runAsUser=1000 makes that update fail with an SCC error, the S3PA
+// is never emptied, and the test times out. Production V2 Mountpoint Pods avoid this by leaving
+// RunAsUser nil on OpenShift (see pkg/cluster/cluster.go MountpointPodUserID), letting SCC pick the
+// UID; we mirror that here.
```
jet-tong, line 259: Interesting. Noting relevant PR https://github.com/awslabs/mountpoint-s3-csi-driver/pull/402 and did see this function in cluster.go:

```
// MountpointPodUserID returns the appropriate RunAsUser for Mountpoint Pod based on the cluster variant.
func (c Variant) MountpointPodUserID() *int64 {
	if c == OpenShift {
		// OpenShift clusters automatically assign non-root uid from predefined namespace range
		// https://www.redhat.com/en/blog/a-guide-to-openshift-and-uids
		return nil
	}

	return defaultMountpointUID
}

```

Will this need a fix in mounter-daemonset.yaml? (I missed it before, currently doesn't have RunAsUser at all).

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
rajdchak: Both in the  V2stale attachment cleaner and V2 reconciler when a workload is terminated we annotate the mp pod needs unmount(it is not immediately deleted) and delete the s3pa. Later the mp pod is deleted after the unmount has happened. So s3pa is deleted before the mp pod. So we do not do anything different. We replicated the same behaviour.
https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/cmd/aws-s3-csi-controller/csicontroller/stale_attachment_cleaner.go#L123-L126

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/pkg/driver/node/mounter/pod_mounter.go#L327-L329 replicated the same behaviour as we have for v1 mounts in CSI Driver v2

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: It falls through the V3 path then, but it would just create an entry in the map, provision credentials and then since the target is already mounted it will just return and not create an source fuse mount. Our periodic resource cleaner then will cleanup the resources it created as the source is unhealthy(it does not exist). 

The behaviour is the same as we have for not v1 mounts in CSI Driver v2, it falls through the v2 path https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/pkg/driver/node/mounter/pod_mounter.go#L119-L134

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
rajdchak: ack

tests/e2e-kubernetes/testsuites/util.go
rajdchak: yeah will remove these, missed cleaning up

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
```diff
@@ -0,0 +1,268 @@
...
+)
+
+// DrainStaleAttachmentCleaner is the controller used in daemonset (V3) mounter mode.
+//
+// During a V2 -> V3 upgrade, existing V2 Mountpoint Pods must still be drained cleanly when their
+// workloads terminate — otherwise they would hang indefinitely. Rather than an event-driven
+// reconciler, drain-only mode runs a single periodic sweep that:
+//
+//   - Removes stale workload UIDs (whose pods no longer exist) from each MountpointS3PodAttachment,
+//     annotates emptied Mountpoint Pods with `needs-unmount`, and deletes emptied S3PAs.
```
muddyfish: What is an 'emptied' Mountpoint pod/S3PA?

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
```diff
@@ -0,0 +1,268 @@
...
+//   - Deletes completed (Succeeded) V2 Mountpoint Pods.
+//   - Deletes leftover V2 Headroom Pods once their workload is gone or past scheduling.
```
muddyfish, line 37: What happens to v2 headroom pods? Do we instead stick them into a daemonset?

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
```diff
@@ -0,0 +1,268 @@
...
+//
+// It never spawns new Mountpoint Pods, creates/updates S3PAs for active workloads, or creates
+// Headroom Pods — those responsibilities belong to V3's DaemonsetMounter, not the controller.
+//
+// This type is intentionally SELF-CONTAINED: it depends only on the API client, so the V2
```
muddyfish: This comment doesn't belong in the final code - comments should be describing the current state, not some past state.

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
```diff
@@ -0,0 +1,268 @@
...
+// spawn/create controller (reconciler.go) and its cleaner (stale_attachment_cleaner.go) can both be
+// deleted without touching drain-only mode.
+type DrainStaleAttachmentCleaner struct {
+	client.Client
+	mountpointNamespace string
+}
+
+// NewDrainStaleAttachmentCleaner creates a DrainStaleAttachmentCleaner. It needs the API client and
```
muddyfish: This is obvious by reading the parameters.

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+}
+
+// RunCleanup performs a single drain sweep: prune stale S3PA workload references, delete completed
+// Mountpoint Pods, and delete leftover Headroom Pods.
+func (cm *DrainStaleAttachmentCleaner) RunCleanup(ctx context.Context) error {
+	log := logf.FromContext(ctx)
+
+	podList := &corev1.PodList{}
+	if err := cm.List(ctx, podList); err != nil {
```
muddyfish, line 75: Should we be filtering for pods in the `mount-s3` namespace?

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
```diff
@@ -0,0 +1,268 @@
...
+		return err
+	}
+	// Create a map of existing pod UIDs for quick lookup
+	existingPods := make(map[string]*corev1.Pod)
+	for _, pod := range podList.Items {
+		existingPods[string(pod.UID)] = &pod
+	}
+
+	// Always attempt Headroom Pod cleanup — in drain-only mode there is no feature-flag gate; any
```
muddyfish: Comment is confusing. What feature flag gate? Seems like we're describing old state that is no longer useful

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
```diff
@@ -0,0 +1,268 @@
...
+		if err := cm.cleanupStaleWorkloads(ctx, &s3paList.Items[i], existingPods); err != nil {
+			log.Error(err, "Error cleaning up S3PodAttachment", "s3pa", s3paList.Items[i].Name)
+			continue
+		}
+	}
+
+	return nil
+}
+
+// cleanupStaleWorkloads removes workload references whose pod no longer exists (and whose attachment
```
muddyfish: This could be re-worded to make more sense. Took several reads to understand what was going on

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
```diff
@@ -0,0 +1,268 @@
...
+	modified := false
+	now := time.Now().UTC()
+
+	for mpPodName, attachments := range s3pa.Spec.MountpointS3PodAttachments {
+		var validAttachments []crdv2.WorkloadAttachment
+		for _, attachment := range attachments {
+			_, exists := existingPods[attachment.WorkloadPodUID]
+			isStale := now.Sub(attachment.AttachmentTime.Time) > drainStaleAttachmentThreshold
+			if exists || !isStale {
+				validAttachments = append(validAttachments, attachment)
```
muddyfish, line 131: This is a strange way of handling this. We're effectively re-building the list, and only saying it's modified if we skipped adding something to the new list.

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
```diff
@@ -0,0 +1,268 @@
...
+			}
+			log.Error(err, "Failed to delete succeeded Mountpoint Pod", "mountpointPod", pod.Name)
+			continue
+		}
+		log.Info("Deleted succeeded Mountpoint Pod", "mountpointPod", pod.Name)
+	}
+	return nil
+}
+
+// cleanupStaleHeadroomPods deletes leftover V2 Headroom Pods whose referenced workload no longer
```
muddyfish, line 182: Is there any time we wouldn't want to just clean up all headroom pods?

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner_test.go
```diff
@@ -0,0 +1,162 @@
...
+			c, cleaner := newDrainCleaner(t, mpPod)
+			assert.NoError(t, cleaner.RunCleanup(ctx))
+
+			err := c.Get(ctx, client.ObjectKey{Namespace: mountpointNamespace, Name: mpPod.Name}, &corev1.Pod{})
+			if !apierrors.IsNotFound(err) {
+				t.Fatalf("expected succeeded Mountpoint Pod to be deleted, get err=%v", err)
+			}
+		})
+
+		t.Run("does not delete Running Mountpoint Pod", func(t *testing.T) {
```
muddyfish: What about failed/pending MP pods?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -833,6 +861,172 @@ func (dm *DaemonsetMounter) isSystemDMountpoint(target string) bool {
...
+	return tracked
+}
+
+// isLegacyV2Mount checks whether the target is a legacy V2 (pod-mounter) mount by inspecting
+// the kernel mount table. Detection is directory-based and definitive:
+//   - V2 sources live under the V2 source dir ("mnt").
+//   - V3 sources live under a separate dir ("v3mnt").
+//
+// It returns true when the target bind-mounts (same device ID) from a source under the V2 dir.
+// Requires SUPPORT_LEGACY_POD_MOUNTS to be enabled.
```
muddyfish, line 888: What if it's not enabled?

pkg/driver/node/mounter/daemonset_mounter.go
muddyfish: +1 for `mnt-v3` or `mnt_v3`

jet-tong: @muddyfish I was saying mnt-v3 / mnt_v3 would not work because of this match, but I do like these 2 options more; prefix match needs change if we change its name (in comments + code)

pkg/driver/node/mounter/daemonset_mounter.go
jet-tong: That's fair, but do you think adding a warning would help here anyways? (I don't imagine any users disabling supportLegacyPodMounts though so not as important).

pkg/driver/node/mounter/mount_map.go
```diff
@@ -159,8 +159,20 @@ func (m *MountMap) Range(fn func(volumeID string, entry *MountEntry) bool) {
...
-// SourceMountPath returns the source mount directory for a given volume.
+// V3SourceMountDirName is the name of the V3 (daemonset-mounter) source mount directory.
+// V3 uses "v3mnt" to keep its source mounts separate from V2 (pod-mounter) mounts which
+// live under "mnt". This separation lets the V2->V3 upgrade path definitively distinguish
+// legacy V2 mounts (under "mnt", leaf = mounter pod name) from V3 mounts (under "v3mnt",
+// leaf = volumeID) purely by directory, with no naming heuristic.
+const V3SourceMountDirName = "v3mnt"
+
+// V3SourceMountDir returns the V3 source mount base directory (".../plugins/s3.csi.aws.com/v3mnt").
+func V3SourceMountDir(kubeletPath string) string {
```
jet-tong, line 170: np: I noticed some V2's SourceMountDir in mounter.go, could we put this function parallel to it (and perhaps maybe rename it as V2SourceMountDir) for tidiness?

charts/aws-mountpoint-s3-csi-driver/values.yaml
rajdchak: I think I was discussing with Simon that customers can have some visibility into when the migration from v1 to v3 or v2 to v3 is complete and then they can disable these flags.

for v2 to v3 may be customers can use no Mountpoint Pods in mount-s3 and no MountpointS3PodAttachment (s3pa) resources left as an indicator that migration is complete

for v1 to v3 I don't yet know what can be the indicator yet but we can think about it as part of a separate task.

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: I am just worried if there will be too many logs per-republish per surviving V2 volume until it drains. Would like to keep the existing behavior.

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: @muddyfish yes as Jensen pointed out, if we change to mnt-v3 or mnt_v3, our prefix match logic will be wrong and it would again require changes in the code logic. So, I would prefer to stick to v3mnt.

tests/e2e-kubernetes/testsuites/drain_only_daemonset.go
rajdchak: the drain-only controller shouldn't do scheduling - ungating is a scheduling action, not teardown. The narrow edge case (gated at the exact moment of cutover) is best handled as a migration precondition — a still-gated pod hasn't started, so recreating it post-upgrade is lossless. I'll note it in the upgrade docs rather than add scheduling logic here.

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
rajdchak: Here's the exact logic the drain-only cleaner uses for Headroom Pods (cleanupStaleHeadroomPods), and why it's safe to run it unconditionally:
```
How a Headroom Pod is identified and reaped, step by step:

1. It lists all pods once, then iterates. For each pod it skips anything that isn't in the Mountpoint namespace or isn't a Headroom Pod. Identity is purely name-prefix based — IsHeadroomPod just checks strings.HasPrefix(pod.Name, "hr-").
2. It reads the referenced workload from the s3.csi.aws.com/headroom-for-pod label (the workload pod's UID). If the label is empty, it logs and skips.
3. It deletes the Headroom Pod when either:
    the referenced workload UID no longer exists in the cluster (!exists), or
    shouldDeleteHeadroomPodForTheWorkloadPod(workloadPod) is true — i.e. the workload is terminating    (DeletionTimestamp != nil) or is scheduled and past Pending.
```
This is the same logic as V2's cleaner; the only difference is V2 gates the whole thing behind isHeadroomForMountpointPodsFeatureEnabled() (true only when both headroom priority-class names are set), and drain-only mode drops that gate.

Why dropping the gate is correct, not risky:

```
1. The operation is delete-only and identity-based. It never creates Headroom Pods and never touches scheduling. If no Headroom Pods exist (feature was never enabled), the loop matches nothing and no-ops.
2, In drain-only mode there is no headroom.go/reconciler creation path running, so cleanup can never race a creator — it can only remove leftover V2 Headroom Pods, which is exactly the migration goal.
3. It actually closes a V2 gap: in V2, if an operator disables the headroom feature (unsets the priority-class names) while Headroom Pods still exist, the V2 cleaner's feature-flag gate makes it stop reaping them, so they orphan. The unconditional drain-only cleanup reaps those regardless.
```

So the answer to your question: yes, in V2 disabling the headroom feature can leave stale Headroom Pods uncleaned — a pre-existing V2 behavior. The drain-only cleaner intentionally has no such gate, and there's no negative consequence to running cleanup without a creation path: it's a safe, delete-only sweep that only removes Headroom Pods whose workload is gone or past scheduling.

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
rajdchak: For now the replicated V2 shouldDeleteHeadroomPodForTheWorkloadPod returns false when the workload exists and is still Pending, and only deletes when the workload is gone, terminating, or scheduled-and-past-Pending. I am not sure if this makes sense in V3 because if a workload is mid scheduling when we migrated to v3, it will be stuck in that state and users will need to terminate the workloads and then our cleaner will get rid of the headroom pods but replicated the V2 behaviour here for safety.

Everywhere else, the predicate collapses to "delete": once the workload is Running (headroom served its purpose), terminated, or no longer exists, the headroom pod is reaped.

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
rajdchak: V2 Headroom Pods are drained (deleted), not migrated — there's no daemonset equivalent to move them into, because V3 has no headroom concept at all.

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
rajdchak: manager.Runnable is just a controller-runtime interface — anything that has a Start(ctx context.Context) error method counts as one.

We register the cleaner with mgr.Add(cm), and Add only accepts a Runnable. So once it's registered, the manager owns the cleaner's lifecycle.

So the Start(ctx) error shape isn't something we chose — it's what the interface requires, and the ctx is how the manager tells us to stop. The comment is just noting which contract the method fulfills. The V2 StaleAttachmentCleaner works the exact same way; this one only makes it explicit in a comment.

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
rajdchak: RunCleanup does an unscoped cm.List(ctx, podList) → builds existingPods keyed by pod UID.
That map is used by three consumers:
1. cleanupStaleWorkloads: checks existingPods[attachment.WorkloadPodUID] — the workload pods, which live in arbitrary user namespaces, NOT mount-s3.
2. cleanupStaleHeadroomPods: reads existingPods[workloadPodUID] (workload again, any namespace) and also filters headroom pods via isInMountpointNamespace.
3. cleanupSucceededMountpointPods: iterates and filters via isInMountpointNamespace.
So the answer is no.

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
rajdchak: Yeah, agreed it's a bit awkward — but this is copied almost exactly from V2's StaleAttachmentCleaner (only the threshold constant name changed). I kept it identical on purpose so the drain-only cleaner stays a faithful copy of V2, which makes deleting V2 cleanly after the migration easier.

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
rajdchak: "Emptied" was shorthand for "left with nothing referencing it" — reworded it now

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
rajdchak: Added comments for ease of review, removed now.

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner_test.go
rajdchak: we don't delete pending or failed mp pods same behaviour as V2, added tests for those cases as well.

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: It falls through the V3 path then, but it would just create an entry in the map, provision credentials and then since the target is already mounted it will just return and not create a source fuse mount. Our periodic resource cleaner then will cleanup the resources it created as the source is unhealthy(it does not exist).

The behaviour is the same as we have for not v1 mounts in CSI Driver v2, it falls through the v2 path https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/pkg/driver/node/mounter/pod_mounter.go#L119-L134

pkg/driver/node/mounter/mount_map.go
rajdchak: will not refactor anything related to V2 in this PR.

tests/e2e-kubernetes/testsuites/drain_only_daemonset.go
rajdchak: Mounter is planned to move to running as root (UID 0) with CAP_SETUID/SETGID/CHOWN/KILL (per the threat model) so it can setuid each mount-s3 child to a per-mount UID. At that point runAsUser becomes 0, so the current 1000 is interim.
Also, ROSA tests pass, which is the proof the mounter daemonset is already admitted fine there today. In the e2e test, the failure was the drain-only controller patching the Mountpoint Pod (to add needs-unmount), and OpenShift SCC re-validates runAsUser: 1000 on that update. The mounter daemonset pod is never patched at runtime, so that specific path doesn't hit it.

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
muddyfish: It seems like this is something we need to consider regarding guidance.

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
```diff
@@ -0,0 +1,270 @@
...
+		workloadPod, exists := existingPods[workloadPodUID]
+		if !exists || shouldDeleteHeadroomPodForTheWorkloadPod(workloadPod) {
+			log.Info("Deleting stale Headroom Pod", "headroomPod", pod.Name, "workloadPodUID", workloadPodUID)
+			if err := cm.Delete(ctx, pod); err != nil {
+				log.Error(err, "Failed to delete stale Headroom Pod", "headroomPod", pod.Name, "workloadPodUID", workloadPodUID)
+				continue
+			}
+		}
+	}
+	return nil
```
muddyfish: We never return an error here even if we ran into one.

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
```diff
@@ -0,0 +1,270 @@
...
+		if err := cm.Delete(ctx, pod); err != nil {
+			if apierrors.IsNotFound(err) {
+				continue
+			}
+			log.Error(err, "Failed to delete succeeded Mountpoint Pod", "mountpointPod", pod.Name)
+			continue
+		}
+		log.Info("Deleted succeeded Mountpoint Pod", "mountpointPod", pod.Name)
+	}
+	return nil
```
muddyfish: We never return an error even if we ran in to one

tests/e2e-kubernetes/testsuites/drain_only_daemonset.go
```diff
@@ -0,0 +1,339 @@
...
+					Name:      mpPodName,
+					Namespace: mountpointNamespace,
+					Labels:    map[string]string{"app": "mp-drain-e2e"},
+				},
+				Spec: v1.PodSpec{
+					RestartPolicy:   v1.RestartPolicyNever,
+					SecurityContext: mountpointPodSecurityContext(),
+					Containers: []v1.Container{{
+						Name:    "pause",
+						Image:   "public.ecr.aws/docker/library/busybox:stable-musl",
```
muddyfish: Do not use public ecr. Use the existing mechanism for `create_pod`

tests/e2e-kubernetes/testsuites/drain_only_daemonset.go
```diff
@@ -0,0 +1,339 @@
...
+// nil on OpenShift and let SCC assign it — mirroring how production Mountpoint Pods handle it (see
+// pkg/cluster/cluster.go MountpointPodUserID).
+func mountpointPodSecurityContext() *v1.PodSecurityContext {
+	sc := &v1.PodSecurityContext{
+		RunAsNonRoot:   ptr.To(true),
+		SeccompProfile: &v1.SeccompProfile{Type: v1.SeccompProfileTypeRuntimeDefault},
+	}
+	if ClusterType != "openshift" {
+		// On vanilla Kubernetes there is no SCC UID range, so pin a concrete non-root UID.
+		sc.RunAsUser = ptr.To(int64(1000))
```
muddyfish, line 249: Any reason not to just leave unassigned on other cluster types?

pkg/driver/node/mounter/daemonset_mounter.go
muddyfish: Why not change the v2 match to `== 'mnt'`?

muddyfish (changes requested):

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
rajdchak: Right, it swallows the error — but this is copied from V2's cleanupStaleHeadroomPods (same log.Error + continue → return nil). Kept identical on purpose so the drain-only cleaner stays a clean copy of V2 . It's logged and re-run each tick, so it retries

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
rajdchak: Right, it swallows the error — but this is copied from V2's cleanupStaleHeadroomPods behavior (same log.Error + continue → return nil). It's logged and re-run each tick, so it retries

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: as agreed changing the name to v3_mnt, keeping prefix matching approach

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
muddyfish: It doesn't make sense though - perhaps cleaning up both sides would be good.

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
rajdchak: yes, will note to provide guidance regarding this in the upgrade documentation task.

tests/e2e-kubernetes/testsuites/drain_only_daemonset.go
rajdchak: The framework's default Linux test image is registry.k8s.io/e2e-test-images/busybox (MakePod → GetDefaultTestImage() → imageutils.BusyBox), and its image config sets no User, so it runs as root (UID 0). The kubelet's verifyRunAsNonRoot fails such a container when runAsNonRoot: true and no runAsUser is set — "container has runAsNonRoot and image will run as root" — so we pin a UID. On OpenShift we leave it unset because SCC assigns one from the namespace range, and a hardcoded UID would be out of range and rejected.

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
```diff
@@ -0,0 +1,267 @@
...
+	cm.cleanupSucceededMountpointPods(ctx, existingPods)
+
+	s3paList := &crdv2.MountpointS3PodAttachmentList{}
+	if err := cm.List(ctx, s3paList); err != nil {
+		return err
+	}
+	for i := range s3paList.Items {
+		if err := cm.cleanupStaleWorkloads(ctx, &s3paList.Items[i], existingPods); err != nil {
+			log.Error(err, "Error cleaning up S3PodAttachment", "s3pa", s3paList.Items[i].Name)
+			continue
```
muddyfish, line 100: Not blocking: `continue` doesn't appear to do anything here.

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
rajdchak: fixed this

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
rajdchak: ack, fixed it

muddyfish (approved):

charts/aws-mountpoint-s3-csi-driver/templates/controller.yaml
```diff
@@ -74,6 +73,10 @@ spec:
               value: {{ .Values.mountpointPod.priorityClassName }}
             - name: MOUNTPOINT_POD_LABELS
               value: {{ toJson .Values.mountpointPod.podLabels | quote }}
+            {{- if eq .Values.experimental.mounterMode "daemonset" }}
```
jet-tong: jet-tong to check: can we disable controller if supportLegacyPodMounts is false?

charts/aws-mountpoint-s3-csi-driver/templates/NOTES.txt
```diff
@@ -8,6 +8,13 @@ The chart in this repository may reference container images that have not been p
 or contain in-progress changes that are incompatible with the images it references.
 Please see supported installation methods in https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/docs/INSTALL.md.
 {{- end }}
+{{- if ne .Values.experimental.mounterMode "daemonset" }}
```
jet-tong, line 11: qq: we will be deleting this warning after removing pod mode right? (Maybe there's no need to add this warning?)

charts/aws-mountpoint-s3-csi-driver/values.yaml
```diff
@@ -99,6 +99,8 @@ controller:
     name: s3-csi-driver-controller-sa
 
 mountpointPod:
+  # NOTE: `namespace` is still required in daemonset (V3) mode — the drain-only controller uses it to
+  # identify leftover V2 Mountpoint/Headroom Pods to clean up.
```
jet-tong, line 103: Thanks for adding it, just small np that since this is customer facing, maybe we can tell them this is only required if `supportLegacyPodMounts: true`? (and daemonset mode of course, but we'll remove mentions of that later anyways)

We could change it later when we remove pod vs daemonset mode mentions though.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -833,6 +861,172 @@ func (dm *DaemonsetMounter) isSystemDMountpoint(target string) bool {
...
+	if s3pa == nil {
+		klog.V(2).Infof("DaemonsetMounter: no matching S3PA found for V2 mount %s (workload %s), mount may be draining", credentialCtx.VolumeID, credentialCtx.WorkloadPodID)
+		return nil
+	}
+
+	mpPod, err := dm.clientset.CoreV1().Pods(mountpointPodNamespace).Get(ctx, mpPodName, metav1.GetOptions{})
+	if err != nil {
+		klog.V(2).Infof("DaemonsetMounter: V2 Mountpoint Pod %q not found, mount may be draining: %v", mpPodName, err)
+		return nil
+	}
```
jet-tong, line 946: (AI flagged, I noticed late): Noted this block replaces [pm.waitForMountpointPod](https://github.com/awslabs/mountpoint-s3-csi-driver/blob/68607a46f028fcc8be6740964b9a68f386e3c503/pkg/driver/node/mounter/pod_mounter.go#L140), so maybe we can improve it by only returning nil if `if apierrors.IsNotFound(err) {`? 

Edit: Looked into it more; to ensure no change in behaviour from v2, maybe we can add back the watcher to daemonset_mounter.go, adding back the 15s mountpointPodReadinessWaitDuration and ErrPodUnschedulable message? 

I did see podWatcher was added anyways in driver.go for CleanupDanglingMounts so can just pass it to DaemonsetMounter as well perhaps, so maybe worth adding? AI draft of that below

```
ctx, cancel := context.WithTimeout(ctx, mountpointPodReadinessWaitDuration) // 15s, pod_mounter.go:37
defer cancel()

mpPod, err := dm.podWatcher.Wait(ctx, mpPodName)
if err != nil {
    if errors.Is(err, watcher.ErrPodNotFound) {
        // V3-specific: in drain-only mode the pod legitimately may already be gone.
        klog.V(2).Infof("DaemonsetMounter: V2 Mountpoint Pod %q gone, mount is draining", mpPodName)
        return nil
    }
    // Not ready / unschedulable / transient — V2 treated all of these as retryable.
    return fmt.Errorf("failed to wait for V2 Mountpoint Pod %q, will retry: %w. %s",
        mpPodName, err, dm.helpMessageForV2MountpointPodStatus(err, mpPodName))
}
```

charts/aws-mountpoint-s3-csi-driver/values.yaml
jet-tong: Can we track this idea / task somewhere?

tests/e2e-kubernetes/testsuites/drain_only_daemonset.go
jet-tong: Can we track it in upgrade docs task? 
(Would it be easier to just add something to handle this edge case in code instead of in docs which increases user burden, or is it hard to do in code?)

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
jet-tong: Fairs, so this becomes 2 min instead of random uniform 0-2min; not a big discrepancy so it's okay I guess.

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
jet-tong: As before on another headroom pod comment - let's track upgrade docs headroom section somewhere.

charts/aws-mountpoint-s3-csi-driver/templates/controller.yaml
```diff
@@ -1,4 +1,3 @@
-{{- if ne .Values.experimental.mounterMode "daemonset" -}}
 kind: Deployment
```
jet-tong, line 1: qq: can we gate the controller if supportLegacyPodMounts is false?

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
```diff
@@ -0,0 +1,267 @@
...
+	log := logf.FromContext(ctx).WithValues("s3pa", s3pa.Name)
+	modified := false
+	now := time.Now().UTC()
+
+	for mpPodName, attachments := range s3pa.Spec.MountpointS3PodAttachments {
+		var validAttachments []crdv2.WorkloadAttachment
+		for _, attachment := range attachments {
+			_, exists := existingPods[attachment.WorkloadPodUID]
+			isStale := now.Sub(attachment.AttachmentTime.Time) > drainStaleAttachmentThreshold
+			if exists || !isStale {
```
jet-tong, line 130: (AI-flagged): AI suggested adding isPodActive check here since we don't do that check in reconciler anymore. 

I was reminded of @muddyfish's PR https://github.com/awslabs/mountpoint-s3-csi-driver/pull/842/ which updates the logic for isPodActive and added a unit test for all edge cases of when we want to mark as non-active. Maybe it's worth incorporating those cases into this PR? I think the only cases that flips would be the succeeded / failed / pending+terminating cases, not sure if it's significant enough that we need to add isPodActive here, so you judge whether we need this change (as I understand this entire part is basically identical to stale_attachement_cleaner.go already).

```go
	tests := []struct {
		name              string
		phase             corev1.PodPhase
		deletionTimestamp *metav1.Time
		expect            bool // isPodActive output
	}{
		{"Pending", corev1.PodPending, noDeleting, true},
		{"Pending + terminating", corev1.PodPending, isDeleting, false}, // adding isPodActive will flip this case
		{"Running", corev1.PodRunning, noDeleting, true},
		{"Running + terminating", corev1.PodRunning, isDeleting, true}, // PR 842 flipped this to true
		{"Succeeded", corev1.PodSucceeded, noDeleting, false}, // adding isPodActive will flip this case
		{"Succeeded + terminating", corev1.PodSucceeded, isDeleting, false}, // adding isPodActive will flip this case
		{"Failed", corev1.PodFailed, noDeleting, false}, // adding isPodActive will flip this case
		{"Failed + terminating", corev1.PodFailed, isDeleting, false}, // adding isPodActive will flip this case
	}

```

AI suggested these, might be useful as reference (after given context above): 
```go
for _, attachment := range attachments {
    pod, exists := existingPods[attachment.WorkloadPodUID]
    // Existence alone is too weak: a terminal pod that lingers in etcd — Jobs with
    // restartPolicy: Never, CronJob history, evicted pods — pins its Mountpoint Pod, and so its
    // mount-s3 process and FUSE source, until the pod object is garbage-collected. Drain-only mode
    // has no reconciler to notice, so nothing else ever drains it.
    //
    // Reuse isPodActive rather than a plain terminal-phase check: #842 requires a
    // Running-but-terminating pod to stay attached, since kubelet keeps republishing for it and
    // dropping the S3PA mid-grace-period leaves the pod hanging in Terminating. Its semantics are
    // pinned by TestIsPodActive.
    alive := exists && isPodActive(pod)
    isStale := now.Sub(attachment.AttachmentTime.Time) > drainStaleAttachmentThreshold
    if alive || !isStale {
        validAttachments = append(validAttachments, attachment)
    } else {
```

They also suggested splitting "keeps attachment whose workload pod still exists" test into running / Succeed cases for that.

charts/aws-mountpoint-s3-csi-driver/templates/NOTES.txt
rajdchak: yes, may be before launch itself we will remove the pod mode and while doing a cleanup we will remove this warning.

charts/aws-mountpoint-s3-csi-driver/values.yaml
rajdchak: can add that, as part of the documentation task for upgrade, keeping this noted in taskei.

tests/e2e-kubernetes/testsuites/drain_only_daemonset.go
rajdchak: I believe it is best to leave to customers to delete and recreate their workloads, we can document it, I have it noted.

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: waitForMountpointPod does this 
```
// waitForMountpointPod waits until Mountpoint Pod for given `podName` is in `Running` state.
// It returns found Mountpoint Pod and it's base directory.
```

As we had discussed in our design review, we are handling existing V2 mounts and not new V2 mounts. So we will not be waiting for mountpoint pod to be in running state, it should already be in that state so it is a one shot check.

cmd/aws-s3-csi-controller/csicontroller/drain_stale_attachment_cleaner.go
rajdchak: I think this bit makes sense, let me look into it and do it in a follow up PR

charts/aws-mountpoint-s3-csi-driver/templates/controller.yaml
jet-tong: (Resolved - we need it for DP later)

jet-tong (approved):
We can track the last 2 (isPodActive + MP-pod pending potential edge case) in follow-up PRs.

PR #940, opened 2026-09-04
Propagate seLinuxOptions to the mounter daemonset pod

tadiwa-aizen:
*Issue #, if available:*

*Description of changes:*
Propagates `seLinuxOptions` to the secondary mounter daemonset (`s3-csi-daemonset-mounter`) pod, matching how the primary `s3-csi-node` pod already sets it (`node.yaml`). On SELinux-enforcing clusters (e.g. OpenShift) the mounter runs `mount-s3` and serves the mounted files, so it needs the operator-chosen SELinux label or mounts can be denied.

Reuses the existing `.Values.node.seLinuxOptions` knob (one value for both pods, set once) with the same `{{- with }}` guard, so the block is omitted when the value is unset. No new values.

This is one item of the V2 operational-parity task; the remaining items and the full helm chart tests are separate CRs.

## Testing

**1) Helm chart validation.** I rendered the chart and checked the mounter pod. When the SELinux value is set, the block shows up with the right values. When it is not set, the block is left out. This shows the chart builds a correct pod spec.

**2) Deploying.** I deployed to a dev cluster in daemonset mode and looked at the running mounter pod. The SELinux settings are there on the live pod. This shows the setting really reaches the mounter, not just on paper.

### DIFF
--- a/charts/aws-mountpoint-s3-csi-driver/templates/mounter-daemonset.yaml
+++ b/charts/aws-mountpoint-s3-csi-driver/templates/mounter-daemonset.yaml
@@ -77,6 +77,15 @@ spec:
             drop: ["ALL"]
           seccompProfile:
             type: RuntimeDefault
+          {{- if eq (include "aws-mountpoint-s3-csi-driver.isOpenShift" .) "false" }}
+          {{- with .Values.node.seLinuxOptions }}
+          seLinuxOptions:
+            user: {{ .user }}
+            type: {{ .type }}
+            role: {{ .role }}
+            level: {{ .level }}
+          {{- end }}
+          {{- end }}
         volumeMounts:
           - name: comm
             mountPath: /comm


charts/aws-mountpoint-s3-csi-driver/templates/mounter-daemonset.yaml
```diff
@@ -77,6 +77,13 @@ spec:
             drop: ["ALL"]
           seccompProfile:
             type: RuntimeDefault
+          {{- with .Values.node.seLinuxOptions }}
+          seLinuxOptions:
+            user: {{ .user }}
+            type: {{ .type }}
+            role: {{ .role }}
+            level: {{ .level }}
+          {{- end }}
```
jet-tong, line 86: Notes
1. Noted it matches https://github.com/awslabs/mountpoint-s3-csi-driver/blob/cca09f72c7d73885f4c45d9a82bf6839cce3b23a/charts/aws-mountpoint-s3-csi-driver/templates/node.yaml#L85-L91
2. Noted relevant PR https://github.com/awslabs/mountpoint-s3-csi-driver/pull/155/changes

 
Questions
- You mentioned "On SELinux-enforcing clusters (e.g. OpenShift) the mounter runs mount-s3 and serves the mounted files, so it needs the operator-chosen SELinux label or mounts can be denied." - Rosa tests were passing before; does any e2e test reflect this change? 
- For the Deploying manual tests (if e2e can't test mounter pod selinux easily), maybe a failure -> success change after adding selinuxoptions would give more confidence than just SELinux settings existence.

tadiwa-aizen: **1.** Covered by the existing e2e. On OpenShift nothing changes — the label is skipped and the SCC sets it, so ROSA still passes. On non-OpenShift SELinux-enforcing clusters the mounter now runs *with* the label and the normal mount tests still pass. That lane is turned on here: [`.github/matrix.yaml#L16-L19`](https://github.com/tadiwa-aizen/mountpoint-s3-csi-driver/blob/cdc83edb6f77f0a344436cd9078b5a92eba228b7/.github/matrix.yaml#L16-L19) (AmazonLinux2023, `selinux-mode: enforcing`).

**2.** That enforcing lane is basically the failure→success test already: if the label were wrong, the mounts on those nodes would break and the store-data tests would fail. So a green run there gives the same confidence — no separate manual demo really needed.

tadiwa-aizen:
**How this is tested (e2e):**

The driver's e2e suite already runs on SELinux-enforcing nodes in two places, and the mounter runs with this change on both:

- **EKS, SELinux enforcing** — the AmazonLinux2023 lane runs with `selinux-mode: enforcing` ([matrix.yaml#L16-L19](https://github.com/tadiwa-aizen/mountpoint-s3-csi-driver/blob/cdc83edb6f77f0a344436cd9078b5a92eba228b7/.github/matrix.yaml#L16-L19)). Here the mounter gets the operator label, and the normal mount read/write tests run.
- **OpenShift (ROSA)** — always SELinux-enforcing. Here the label is skipped (the SCC assigns it), and the mounter comes up and mounts.

These tests run on enforcing nodes, so if the label were wrong the mounts would fail. If these tests pass, that proves the label is correct and the mounter works — so a green run is the proof this is tested and working.

PR #944, opened 2026-09-08
Version skew detection

rajdchak:
*Issue #, if available:*

*Description of changes:*
Reject a version-skewed CSI Driver Node before spawning anything: the mounter uses an OnDelete update strategy, so after a Helm upgrade a new driver may talk to this (still-old) mounter until the node is recycled. On mismatch we do NOT mount — we close the FUSE fd and surface a clear error via the .error file (which the driver's waitForMount polls), so NodePublishVolume fails fast instead of producing a silent/partial mount.

### DIFF
--- a/cmd/aws-s3-csi-daemonset-mounter/handler.go
+++ b/cmd/aws-s3-csi-daemonset-mounter/handler.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"net"
 	"regexp"
 	"syscall"
@@ -32,6 +33,22 @@ func handleConnection(conn *net.UnixConn, mountpointPath string, pm *ProcessMana
 	if mountId == "" || !validMountId.MatchString(mountId) {
 		syscall.Close(options.Fd)
 		klog.Errorf("Received mount request with invalid mountId: %q", mountId)
+		return
+	}
+
+	// Reject a version-skewed CSI Driver Node before spawning anything: the mounter uses an OnDelete
+	// update strategy, so after a Helm upgrade a new driver may talk to this (still-old) mounter until
+	// the node is recycled. On mismatch we do NOT mount — we close the FUSE fd and surface a clear
+	// error via the .error file (which the driver's waitForMount polls), so NodePublishVolume fails
+	// fast instead of producing a silent/partial mount.
+	if options.ProtocolVersion != mountoptions.ProtocolVersion {
+		syscall.Close(options.Fd)
+		msg := fmt.Sprintf("protocol version mismatch: mounter=%q driver=%q. "+
+			"The s3-csi-daemonset-mounter pod is running a version incompatible with the CSI Driver Node; "+
+			"recycle the node or delete the mounter pod so it is recreated at the matching version.",
+			mountoptions.ProtocolVersion, options.ProtocolVersion)
+		klog.Error(msg)
+		pm.WriteErrorFile(mountId, []byte(msg))
 		return
 	}
 
--- a/cmd/aws-s3-csi-daemonset-mounter/process_manager.go
+++ b/cmd/aws-s3-csi-daemonset-mounter/process_manager.go
@@ -97,11 +97,7 @@ func (pm *ProcessManager) Launch(mountId string, mountpointPath string, options 
 		pm.mu.Unlock()
 
 		if exitCode != 0 {
-			errPath := filepath.Join(pm.commDir, mountId+errorFileExt)
-			// TODO(vlaad): write error file atomically (open,write,rename)
-			if writeErr := os.WriteFile(errPath, stderr, errorFilePerm); writeErr != nil {
-				klog.Errorf("Failed to write error file for mount %s: %v", mountId, writeErr)
-			}
+			pm.WriteErrorFile(mountId, stderr)
 			klog.Errorf("Mountpoint for mount %s exited with code %d", mountId, exitCode)
 		} else {
 			klog.Infof("Mountpoint for mount %s exited cleanly", mountId)
@@ -109,6 +105,17 @@ func (pm *ProcessManager) Launch(mountId string, mountpointPath string, options 
 	}()
 
 	return nil
+}
+
+// WriteErrorFile writes `content` to <comm-dir>/<mountId>.error, the file the CSI Driver Node polls
+// (via waitForMount) to surface a mount failure. Used both when a Mountpoint process exits non-zero
+// and when a mount request is rejected before spawning (e.g. protocol version mismatch).
+func (pm *ProcessManager) WriteErrorFile(mountId string, content []byte) {
+	errPath := filepath.Join(pm.commDir, mountId+errorFileExt)
+	// TODO(vlaad): write error file atomically (open,write,rename)
+	if writeErr := os.WriteFile(errPath, content, errorFilePerm); writeErr != nil {
+		klog.Errorf("Failed to write error file for mount %s: %v", mountId, writeErr)
+	}
 }
 
 // Shutdown sends SIGTERM to all processes and waits for them to exit.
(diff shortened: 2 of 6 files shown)


cmd/aws-s3-csi-daemonset-mounter/handler.go
```diff
@@ -35,6 +36,22 @@ func handleConnection(conn *net.UnixConn, mountpointPath string, pm *ProcessMana
 		return
 	}
 
+	// Reject a version-skewed CSI Driver Node before spawning anything: the mounter uses an OnDelete
+	// update strategy, so after a Helm upgrade a new driver may talk to this (still-old) mounter until
+	// the node is recycled. On mismatch we do NOT mount — we close the FUSE fd and surface a clear
+	// error via the .error file (which the driver's waitForMount polls), so NodePublishVolume fails
+	// fast instead of producing a silent/partial mount.
+	if options.ProtocolVersion != mountoptions.ProtocolVersion {
```
muddyfish, line 44: We should support older protocols - it's only later protocols we should fail on.

cmd/aws-s3-csi-daemonset-mounter/process_manager.go
```diff
@@ -111,6 +107,17 @@ func (pm *ProcessManager) Launch(mountId string, mountpointPath string, options
 	return nil
 }
 
+// WriteErrorFile writes `content` to <comm-dir>/<mountId>.error, the file the CSI Driver Node polls
+// (via waitForMount) to surface a mount failure. Used both when a Mountpoint process exits non-zero
```
muddyfish, line 111: The comment does not need to include when it is used.

cmd/aws-s3-csi-daemonset-mounter/handler.go
rajdchak: The "old driver → new mounter" case only happens during rollback where the previous release version had an older protocol version and self-heals once the mounter pod is recreated (which OnDelete already requires). So backward compatible here means just smooth rollbacks, at the cost of a permanent additive-only protocol contract + keeping old-version parsing forever. Given it's a new architecture with nothing in prod, is that worth it, or is fail-fast + "recycle the mounter on rollback" in the docs enough?

cmd/aws-s3-csi-daemonset-mounter/handler.go
```diff
@@ -35,6 +36,22 @@ func handleConnection(conn *net.UnixConn, mountpointPath string, pm *ProcessMana
...
+		syscall.Close(options.Fd)
+		msg := fmt.Sprintf("protocol version mismatch: mounter=%q driver=%q. "+
+			"The s3-csi-daemonset-mounter pod is running a version incompatible with the CSI Driver Node; "+
+			"recycle the node or delete the mounter pod so it is recreated at the matching version.",
```
jet-tong, line 48: Do we want to tell them to 'cordon, drain, delete mounter pod, and uncordon' in here since just deleting mounter pod would be disruptive; and maybe point them towards docs relating to upgrading between v3 versions?

Also btw, have you tested it manually by bumping ProtocolVersion? Didn't see e2e tests but just want to make sure 
1/ the error is not too noisy (e.g. it doesn't cause crashloopbackoff for deployments in workload pods), 
2/ the upgrade path with the 'cordon, drain, delete mounter pod, and uncordon' process - downtime can be minimized; maybe we have to document how to gradually roll out the upgrade slowly to minimize downtime / downed nodes
3/ helm rollback with mismatch - requires cordon drain delete uncordon process as well?

pkg/mountpoint/mountoptions/mount_options.go
```diff
@@ -18,13 +18,25 @@ import (
...
 // An Options struct represents mount options to use while invoking Mountpoint.
 type Options struct {
 	// Fd will be passed over Unix socket using `SCM_RIGHTS`, not as part of the serialized JSON.
-	Fd         int      `json:"-"`
-	BucketName string   `json:"bucketName"`
-	Args       []string `json:"args"`
-	Env        []string `json:"env"`
+	Fd int `json:"-"`
+	// ProtocolVersion is the driver<->mounter protocol version; see [ProtocolVersion].
+	ProtocolVersion string   `json:"protocolVersion"`
```
jet-tong, line 36: qq - why not use integer like in fd?

cmd/aws-s3-csi-daemonset-mounter/process_manager_test.go
```diff
@@ -121,6 +122,56 @@ func TestHandleConnection_PropagatesOptionsToRunner(t *testing.T) {
...
+
+	// No Mountpoint process should have been spawned.
+	fr.mu.Lock()
+	assert.Equals(t, 0, len(fr.handles))
+	fr.mu.Unlock()
+
+	// A clear error must be written to the mount's .error file for the driver to surface.
+	content, err := os.ReadFile(filepath.Join(commDir, mountId+errorFileExt))
+	assert.NoError(t, err)
+	if len(content) == 0 {
```
jet-tong, line 170: Can we assert it contains 'protocol version mismatch'? (perhaps store it as a variable to reuse in source/test if we plan to change this string in the future)

cmd/aws-s3-csi-daemonset-mounter/handler.go
jet-tong, line 44: There's 2 ways for the version skews to mismatch right? 

| - | mounter Protocol 1 | mounter protocol 2 |
| -- | -- | -- |
| driver protocol 1 | match | rollback case mismatch |
| driver protocol 2 | upgrade case mismatch | match |

I'm thinking there is 3 different types of version skews.
- Upgrade mismatch only - Driver adds a requirement. e.g. Mountpoint adds optional feature flag, driver auto-injects it, e.g. memory-target. Only driver 2 mounter 1 fails. 
- Rollback mismatch only - Mounter adds a requirement (or removes a tolerance). e.g. Mountpoint removes deprecated flag, which the driver auto-injects e.g. max-cache-size. Only driver 1 mounter 2 fails. 
- Upgrade+Rollback mismatch - different requirements, e.g. renamed mountpoint feature flag. Fails for both. Of course we can mitigate by e.g. Mountpoint supporting both new and old flags for a number of versions, so there will be a version skew gap before that feature flag fully renames. 

I haven't figured out how we can support all 3 cases even with AI help, maybe worth thinking this problem more about. AI suggested something closer to version skew idea (a number of versions that new driver can support for old mounts, but not too old), maybe useful as reference:

```go
// Bump ProtocolVersion → type 1. Bump MinSupportedProtocol → type 2. Bump both → type 3.
// (jet-tong note: type 1/2/3 matches 3 bullet points above)

// mountoptions
// ProtocolVersion is the wire-format generation the driver stamps on every
// request. 0 means a driver predating versioning; its format equals generation 1.
const ProtocolVersion = 1

// mounter
const (
    maxProtocol = mountoptions.ProtocolVersion // newest driver I understand
    minProtocol = 1                            // oldest driver output I can still consume
                                               // bump ONLY when dropping tolerance
)

func checkProtocol(v int) error {
    if v == 0 {
        v = 1
    }
    if v > maxProtocol {
        return fmt.Errorf("CSI Driver Node protocol %d is newer than this mounter supports (%d); "+
            "the mounter pod is stale. Drain the node to complete the upgrade "+
            "(cordon, drain, delete the mounter pod, uncordon).", v, maxProtocol)
    }
    if v < minProtocol {
        return fmt.Errorf("CSI Driver Node protocol %d is older than this mounter supports (min %d); "+
            "the driver was rolled back but this mounter pod was not. Drain the node so the "+
            "mounter is recreated at the matching version, or roll the driver forward.", v, minProtocol)
    }
    return nil
}
```

jet-tong, line 44: Separately I was thinking whether in the future we might want to do different actions in driver/s3-csi-node depending on mounter version (e.g. new mountpoint flag; auto inject flag if protocol 2, otherwise don't inject flag?) Does this mean we want some checking logic in s3-csi-node instead? (Just food for thought)

PR #950, opened 2026-09-11
Add proxy support e2e tests

tadiwa-aizen:
*Issue #, if available:*

*Description of changes:*

Re-enable the proxy e2e test so it runs in daemonset mode.

Proxy support (`HTTPS_PROXY`/`NO_PROXY` set via `mountpointEnv.*` volume attributes, #663) needs to work in V3, and the daemonset path already passes these env vars through to `mount-s3`. This just turns the existing test back on to prove it works end-to-end.

The test was switched off on the feature branch alongside the cache/pod-sharing/credentials tests, but proxy doesn't depend on any of those — it only checks that `mount-s3` gets the proxy env — so it can be re-enabled as-is, no rewrite needed.

### Testing

### DIFF
--- a/tests/e2e-kubernetes/testsuites/proxy.go
+++ b/tests/e2e-kubernetes/testsuites/proxy.go
@@ -3,6 +3,7 @@ package custom_testsuites
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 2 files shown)


tests/e2e-kubernetes/testsuites/proxy.go
```diff
@@ -185,13 +186,44 @@ func (t *s3CSIProxyTestSuite) DefineTests(driver storageframework.TestDriver, pa
...
+			// each log line with the mount id (the PV name). Match a line that is both for this
+			// volume ("[<pv-name>]") and shows the proxy tunnel, so we don't match other volumes.
+			mounterPods, err := f.ClientSet.CoreV1().Pods(csiDriverDaemonSetNamespace).List(ctx, metav1.ListOptions{
+				LabelSelector: "app=s3-csi-daemonset-mounter",
+			})
+			framework.ExpectNoError(err)
+
+			volumePrefix := fmt.Sprintf("[%s]", resource.Pv.Name)
+			foundForVolume := false
+			for i := range mounterPods.Items {
```
jet-tong, line 202: np: As an alternative maybe we can just search the specific node where the workload pod is in, and search that one instead? (probs need to assert 1 mounter pod there too)

jet-tong (approved):

PR #952, opened 2026-09-14
Bound per-volume memory with --memory-target

yerzhan7:
*Issue #, if available:* N/A

*Description of changes:*
- Bound per-volume memory with --memory-target based on mounter memory resources
- Emit warning if --memory-target won't be set
- Add fail fast validation logic in Helm on misconfigs


charts/aws-mountpoint-s3-csi-driver/templates/mounter-daemonset.yaml
```diff
@@ -3,6 +3,29 @@
 {{- fail "daemonsetMounters must have exactly one entry" }}
 {{- end }}
 {{- $mounter := index .Values.daemonsetMounters 0 }}
+{{- $resources := $mounter.resources | default dict }}
+{{- $limits := $resources.limits | default dict }}
+{{- $requests := $resources.requests | default dict }}
+{{- /* The mounter divides its container's memory budget — limits.memory if set, else requests.memory —
+       by maxVolumesPerNode to give each Mountpoint a --memory-target. Fail early on a sizing that is
+       visibly undersized. Best-effort only: an unparseable quantity skips the check, and the mounter
```
rajdchak, line 11: can we elaborate on what value we consider to be visibly undersized?

charts/aws-mountpoint-s3-csi-driver/templates/_helpers.tpl
```diff
@@ -100,3 +100,24 @@ Determine if running on OpenShift (incl. ROSA)
...
+{{- $value := toString . -}}
+{{- if regexMatch `^[0-9]+(\.[0-9]+)?(Ki|Mi|Gi|Ti|Pi|Ei|k|M|G|T|P|E)?$` $value -}}
+{{- $suffix := regexFind `[A-Za-z]+$` $value -}}
+{{- $number := trimSuffix $suffix $value | float64 -}}
+{{- $multiplier := 1.0 -}}
+{{- if $suffix -}}
+{{- $multiplier = get (dict
+      "Ki" 1024.0 "Mi" 1048576.0 "Gi" 1073741824.0
+      "Ti" 1099511627776.0 "Pi" 1125899906842624.0 "Ei" 1152921504606846976.0
+      "k" 1000.0 "M" 1000000.0 "G" 1000000000.0
```
jet-tong, line 118: TIL lowercase k is correct and follows https://kubernetes.io/docs/reference/kubernetes-api/definitions/quantity-resource/ and https://en.wikipedia.org/wiki/Binary_prefix. Maybe worth adding a link as comment (maybe there's a better link than these ones)

cmd/aws-s3-csi-daemonset-mounter/memory_target.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+// containerMemoryBudgetBytes returns the memory budget to divide between hosted Mountpoints.
+func containerMemoryBudgetBytes() (budgetBytes int64, budgetField string) {
+	if bytes := memoryEnvBytes(memoryLimitEnvName); bytes > 0 {
+		return bytes, "limits.memory"
+	}
+	if bytes := memoryEnvBytes(memoryRequestEnvName); bytes > 0 {
+		return bytes, "requests.memory"
+	}
```
jet-tong, line 28: qq: Should limits be prioritized or requests be prioritized? I was thinking we should use 'requests' since it matches the final desired memory usage more than limits, and that it should be smaller than requeusts anyways; maybe I was missing an argument for the reverse. 

edit: or min(requests.memory, limits.memory)

cmd/aws-s3-csi-daemonset-mounter/memory_target.go
```diff
@@ -0,0 +1,112 @@
...
+		klog.Warningf("Not setting %s on hosted Mountpoints: %s. Each Mountpoint that does not set "+
+			"%s in its PV mountOptions will instead target 95%% of the memory it detects, so together "+
+			"they can exhaust the node. Set daemonsetMounters[].resources.limits.memory (or "+
+			"requests.memory) and a non-zero daemonsetMounters[].maxVolumesPerNode.",
+			mountpoint.ArgMemoryTarget, reason, mountpoint.ArgMemoryTarget)
+		return memoryTarget{}
+	}
+
+	// Floor the share, so the combined target of all volumes never exceeds this container's budget.
+	targetMiB := budgetBytes / maxVolumesPerNode / bytesPerMiB
```
jet-tong, line 95: For this formula - I think Mountpoint could still go over memory target, so just using the whole budget could mean exceeding limits.memory which would probably cause it to go OOM killed.

Do we want to consider e.g. using 95% of that budget? Or maybe adding a e.g. 256MiB reserved offset so we can use more than 95% of budget (so big budgets e.g. p5's 2TiB memory would have less wasted memory)?

cmd/aws-s3-csi-daemonset-mounter/memory_target_test.go
```diff
@@ -0,0 +1,211 @@
...
+		},
+		{
+			name:              "no memory budget means no target",
+			maxVolumesPerNode: 4,
+			want:              0,
+		},
+		{
+			name:              "no volume limit means no target",
+			memoryBudgetBytes: 2 * gib,
+			want:              0,
```
jet-tong, line 54: Would be good to have a test case passing in maxVolumesPerNode: 0, which we currently allow meaning volume limits are disabled.

(qq - do we think we should still allow setting 0 maxVolumesPerNode? I'm thinking that was a bad addition now that we're requiring it for memory limit integration, and later max-cache-size integration)

cmd/aws-s3-csi-daemonset-mounter/process_manager.go
```diff
@@ -52,6 +55,21 @@ func (pm *ProcessManager) Launch(mountId string, mountpointPath string, options
 	args := mountpoint.ParseArgs(options.Args)
 	args.Set(mountpoint.ArgForeground, mountpoint.ArgNoValue)
 
+	if !args.Has(mountpoint.ArgMemoryTarget) {
+		memoryTarget, err := pm.memoryTarget.arg()
+		switch {
+		case err != nil:
+			// Mountpoint's CLI would reject the undersized share with a usage error naming a flag the
+			// user never wrote, so refuse before spawning and report the sizing itself instead.
```
jet-tong, line 63: I was considering clamping + warn instead of refusing before; refusing would probably be safer for memory usage, but is it acceptable user experience to fail all mounts in a node? 

I understand there's already a helm gate so this is only for impossible edge case where helm charts passed but code not passed, maybe it's okay keeping the failure consistent instead of clamping and hiding failure?

cmd/aws-s3-csi-daemonset-mounter/process_manager.go
```diff
@@ -52,6 +55,21 @@ func (pm *ProcessManager) Launch(mountId string, mountpointPath string, options
 	args := mountpoint.ParseArgs(options.Args)
 	args.Set(mountpoint.ArgForeground, mountpoint.ArgNoValue)
 
+	if !args.Has(mountpoint.ArgMemoryTarget) {
```
jet-tong, line 58: What happens if user supplies memory target in their PVs currently? 

Do we want to at least warn based on the total memory-target we supplied to current mounts perhaps? 

We were discussing to enforce non-unique volumeHandles by tracking an in-memory cache + regenerating it from meta/ information + rejecting in ValidateCompatibility if it is a repeated volumeHandle. Maybe we can align this rejection/warning with that? i.e. Tracking total --memory-target across different PVs (with different volumeHandles for maxVolumesPerNode works) and reject or warn if exceeding that limit, regenerating via meta/ information reading the mountOption flag. @tadiwa-aizen

pkg/mountpoint/args.go
```diff
@@ -27,6 +28,9 @@ const (
 	ArgCABundle        = "--ca-bundle"
 )
 
+// MinMemoryTargetMiB is the smallest [ArgMemoryTarget] Mountpoint's CLI accepts.
+const MinMemoryTargetMiB = 512
```
jet-tong, line 32: qq: Is this likely to change or 99% chance will stay 512MiB? 

Just want to make sure before we hardcode a Mountpoint constant (not sure if there's a better way to extract this value from Mountpoint which would be nicer).

cmd/aws-s3-csi-daemonset-mounter/process_manager.go
```diff
@@ -52,6 +55,21 @@ func (pm *ProcessManager) Launch(mountId string, mountpointPath string, options
...
+		memoryTarget, err := pm.memoryTarget.arg()
+		switch {
+		case err != nil:
+			// Mountpoint's CLI would reject the undersized share with a usage error naming a flag the
+			// user never wrote, so refuse before spawning and report the sizing itself instead.
+			fuseDev.Close()
+			pm.writeErrorFile(mountId, []byte(err.Error()))
```
jet-tong, line 65: qq: Is there an advantage of injecting the flag in mounter pod instead of s3-csi-node? I think in v2 we did add some max-cache-size injection in node.go, so maybe it's easier to move the injection there, so we don't need to report errors via writeErrorFile? e.g. `args.Value(mountpoint.ArgMemoryTarget)` in node.go.

charts/aws-mountpoint-s3-csi-driver/templates/mounter-daemonset.yaml
```diff
@@ -3,6 +3,29 @@
...
+{{- $budgetField := "limits.memory" }}
+{{- if not $budget }}
+{{- $budget = $requests.memory }}
+{{- $budgetField = "requests.memory" }}
+{{- end }}
+{{- $maxVolumes := int $mounter.maxVolumesPerNode }}
+{{- if and $budget (gt $maxVolumes 0) }}
+{{- $budgetBytes := include "aws-mountpoint-s3-csi-driver.memoryQuantityBytes" $budget }}
+{{- if $budgetBytes }}
+{{- $shareMiB := divf (divf (float64 $budgetBytes) $maxVolumes) 1048576.0 }}
```
jet-tong, line 23: Would it still work if I put negative values for both sides? xD
Maybe negative check is useful here? (or maybe I missed it)

cmd/aws-s3-csi-daemonset-mounter/memory_target.go
```diff
@@ -0,0 +1,112 @@
...
+
+const (
+	memoryLimitEnvName   = "MOUNTER_MEMORY_LIMIT_BYTES"
+	memoryRequestEnvName = "MOUNTER_MEMORY_REQUEST_BYTES"
+)
+
+// containerMemoryBudgetBytes returns the memory budget to divide between hosted Mountpoints.
+func containerMemoryBudgetBytes() (budgetBytes int64, budgetField string) {
+	if bytes := memoryEnvBytes(memoryLimitEnvName); bytes > 0 {
+		return bytes, "limits.memory"
```
rajdchak, line 24: If each mountpoint has a target limit/N, the sum equals the limit with zero headroom for the mounter's own Go runtime/overhead, in that case the kernel OOM-kills the entire mounter pod. Do we need to consider that?

cmd/aws-s3-csi-daemonset-mounter/memory_target.go
rajdchak: +1 to min(requests.memory, limits.memory)

cmd/aws-s3-csi-daemonset-mounter/memory_target.go
```diff
@@ -0,0 +1,112 @@
...
+// it hosts, logging the inputs it used so `kubectl logs` can answer why a mount was sized the way it was.
+// budgetField names the resources field budgetBytes came from ("" when there is no budget), from
+// [containerMemoryBudgetBytes].
+func resolveMemoryTarget(budgetBytes int64, budgetField string, maxVolumesPerNode int64) memoryTarget {
+	if budgetBytes <= 0 || maxVolumesPerNode <= 0 {
+		// Report which input is missing rather than a byte count: a zero budget means "none declared",
+		// which is the opposite of a small one — Mountpoint then sizes itself against the whole node.
+		var reason string
+		switch {
+		case budgetBytes <= 0 && maxVolumesPerNode <= 0:
```
rajdchak, line 75: can these values be negative?

cmd/aws-s3-csi-daemonset-mounter/memory_target.go
```diff
@@ -0,0 +1,112 @@
...
+			reason = "this container declares no memory budget (neither resources.limits.memory nor " +
+				"resources.requests.memory is set), and maxVolumesPerNode is 0"
+		case budgetBytes <= 0:
+			reason = "this container declares no memory budget (neither resources.limits.memory nor " +
+				"resources.requests.memory is set)"
+		default:
+			reason = fmt.Sprintf("maxVolumesPerNode is 0, so this container's memory budget "+
```
rajdchak, line 82: I think the decision was 
```We have decided to integrate with mountpoint memory limiter and have proposed to have a default value of 1 GiB for memory target per mountpoint process based on Andy's experiments```

So may be we should do this for maxVolumesPerNode 0. Setting to nothing leads to the risk of memory pressure which will cause workload pod evictions as we had seen from my experiments.

cmd/aws-s3-csi-daemonset-mounter/process_manager.go
rajdchak: if we allow overriding the value for one mountpoint process, it could alone consume all of the memory budget, then will the split of memory limit/request make sense for the other mountpoint processes?

cmd/aws-s3-csi-daemonset-mounter/process_manager.go
rajdchak: @jet-tong I think mounter pod is the one that actually knows its own container memory budget

cmd/aws-s3-csi-daemonset-mounter/process_manager.go
jet-tong: The (memory limit/request - used_memory) / (maxVolumesPerNode - existing PV count) formula works but has pros and cons and still will be affected by the 1 mount consumning all memory budget problem.

cmd/aws-s3-csi-daemonset-mounter/process_manager.go
jet-tong: Good point I missed that previously... Will think about how max-cache-size should work too.

cmd/aws-s3-csi-daemonset-mounter/memory_target.go
yerzhan7: Ok, will switch to requests if both requests and limits are specified.

charts/aws-mountpoint-s3-csi-driver/templates/_helpers.tpl
yerzhan7: Added link, thanks for reviewing!

pkg/mountpoint/args.go
yerzhan7: For now it's likely to stay

charts/aws-mountpoint-s3-csi-driver/templates/mounter-daemonset.yaml
yerzhan7: Fixed by re-phrasing

cmd/aws-s3-csi-daemonset-mounter/memory_target.go
yerzhan7: As discussed offline, switched to only `requests`.

cmd/aws-s3-csi-daemonset-mounter/memory_target.go
yerzhan7: For the purposes of CSI Driver we assume Mountpoint adheres to the memory target. (Mountpoint already has overhead reserve for such cases)

cmd/aws-s3-csi-daemonset-mounter/memory_target_test.go
yerzhan7: Discussed offline:  Allow `maxVolumesPerNode = 0` only if strategy is `none`

cmd/aws-s3-csi-daemonset-mounter/process_manager.go
yerzhan7: We now fail at startup, I think it's better to fail fast.

cmd/aws-s3-csi-daemonset-mounter/process_manager.go
yerzhan7: Discussed offline.

cmd/aws-s3-csi-daemonset-mounter/memory_target.go
yerzhan7: Added 64MiB overhead for the mounter Go process.

cmd/aws-s3-csi-daemonset-mounter/memory_target.go
yerzhan7: This was just extra safety check. In theory, not possible because Helm render time validations.

cmd/aws-s3-csi-daemonset-mounter/memory_target.go
yerzhan7: Discussed offline

PR #954, opened 2026-09-14
ci: Cancel superseded PR runs via concurrency groups

passaro:
Pushing again to a PR left the previous run's jobs queued alongside the new one, holding slots on the ubuntu-latest runner pool. Cancellation is gated on the pull request event so runs on main and in the merge queue are never cancelled.

e2e-test-untrusted keys on the PR number instead of github.ref, since on pull_request_target the ref is the base branch and would group every open PR together.


jet-tong (approved):
We force pushed it twice after e2e trusted (triggered by pull request) started, and made sure the previous runs were cancelled correctly:
- E2E Trusted: https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/35237917781
- Controller tests: https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/35237917549
- Unit test: https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/35237916784

For Verify Helm Chart it ran very quickly (17s) so it passed before we force pushed it a second time: https://github.com/awslabs/mountpoint-s3-csi-driver/actions/runs/35237917134

We'll have to keep an eye out for E2E Untrusted which is not tested in this PR (and also because group handling is diffferent for pull_request_target).

PR #956, opened 2026-09-15
Warn when a different PV reuses a volumeHandle on a node

tadiwa-aizen:
*Issue #, if available:*

*Description of changes:*
Kubernetes identifies CSI volumes by driver + volumeHandle, so two PVs that share a volumeHandle collide: the scheduler undercounts them against maxVolumesPerNode, and two such volumes in one pod silently resolve to a single mount. Today nothing surfaces this beyond the troubleshooting docs.

Track, per node, which PV first used each volumeHandle at NodePublishVolume, and log a warning (once per offending PV) when a different PV reuses it. Release the claim on NodeUnpublishVolume.

Detection only: it does not block the mount. It covers the cross-pod case, where both NodePublishVolume calls reach the driver. It cannot catch two volumes in the same pod, since kubelet collapses same-handle volumes into one before calling NodePublishVolume.

### Testing
_Manual Testing_
Tested on an EKS cluster (v1.36, single node, daemonset mode) with the warn build deployed. For
each case we applied fresh PVs/pods with real S3 buckets and checked the pod status and the
`s3-csi-node` logs for the warning.

| Case | Setup | Expected | Result |
|---|---|---|---|
| Unique handle (control) | 1 pod, 1 PV | no warning |  Running, no warning |
| Legit same-PV sharing | 2 pods, same PV | no warning | both Running, no warning |
| Cross-pod duplicate | 2 pods, 2 PVs, same handle | warn, both mount | both Running, **warning logged** |
| Same-pod duplicate | 1 pod, 2 volumes, same handle | not caught (blind spot) | no warning (kubelet merges before NPV) |
| Reuse after release | owner → delete → reuse; then concurrent reuse | no warn after release; warn on concurrent |  no warn after release; warns on concurrent |

Notes: warn fires once per offending PV (no per-republish spam).


pkg/driver/node/node.go
```diff
@@ -100,6 +107,15 @@ func (ns *S3NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePubl
 		return nil, status.Error(codes.InvalidArgument, "Target path not provided")
 	}
 
+	// Warn if a different PV is reusing this volumeHandle on the node.
```
jet-tong, line 110: As discussed outside github, maybe we want to explore rejection more.

- On kubelet retrying - does kubelet still retry even with status.Errorf error, perhaps with codes.InvalidArgument or another code? (I think that way they won't retry)
  -  I checked EBS (via AI) seems to reject duplicate volumeHandles https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/8c52908bfe9198dbe077a26fde69a176c931ba85/pkg/driver/node.go#L194 with codes.Aborted - maybe check if this is what we want?
- On 'stuck-pod' - I don't think kubelet would be smart enough to place it in another pod after all (maybe still check online). But we probably still want to fail bad NPV calls so users are aware and can manually intervene and fix the problem (as duplicate volumeHandle will need manual fix anyways)

Edit: another point is on s3-csi-node restart - for warning or rejection, we'd still need to store volumeHandle to meta / mount map persistence and re-populate the map on node restart from with the mount map (and not rely on republish requests which is slow to repopulate, and before it gets populated a duplicate volumeHandle could land without failing)

PR #958, opened 2026-09-15
Reject duplicate volume handle

tadiwa-aizen:
*Issue #, if available:*

*Description of changes:*
Kubernetes identifies a CSI volume by its `volumeHandle`, so two PVs that share a handle on one node collide — two such volumes in one pod resolve to a single mount, and a node can end up running more Mountpoint instances than `maxVolumesPerNode` intends.

This enforces per-node `volumeHandle` uniqueness in the daemonset mounter. Each mounted volume's handle is tracked in the per-node `MountMap` and persisted to its `.meta.json` (rebuilt on csi-node restart). On `NodePublishVolume`, if a *different* PV already holds the handle on the node, the mount is rejected the same way `ValidateCompatibility` rejects an incompatible share. This caps the Mountpoint instances actually running on a node; it does not change the scheduler's volume count (that happens before the mount).

```mermaid
flowchart TD
    NPV[NodePublishVolume] --> New{"Need a new source mount?<br/>(fresh, or previous source died)"}
    New -- "No: already mounted (republish / extra pod, same PV)" --> Mount[Mount / share]
    New -- Yes --> Dup{"Handle already held by a different PV?"}
    Dup -- No --> Claim["Claim handle + persist to .meta.json"] --> Mount
    Dup -- Yes --> Reject["Reject: FailedMount; pod stuck, kubelet retries on same node"]
    Reject -. "re-checked on each retry" .-> Dup

    Unpub["NodeUnpublishVolume (last pod gone)"] --> Release["Release handle"]
    Release -. "handle now free → a stuck duplicate's next retry claims it" .-> Dup

    Restart["csi-node restart"] --> Rebuild["Re-claim handles from .meta.json<br/>(duplicates still rejected)"]
```

Known limit: two same-handle volumes in a *single* pod aren't caught (kubelet collapses them before `NodePublishVolume`).

*Testing:*
Unit tests for claim/reject/release and the restart re-claim (`mount_map_test.go`, `mount_map_persistence_test.go`), plus an e2e test (`pod_sharing_daemonset.go`) asserting a second PV reusing a mounted handle fails to mount with a clear `FailedMount`. Manually verified on EKS v1.36: reject, sharing intact, restart persistence, and handle release on unmount.


pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -1247,6 +1254,14 @@ func (dm *DaemonsetMounter) RebuildMountMap() error {
 
 		dm.populateEntryFromMeta(meta, sourcePath, true, targets)
 
+		// Re-claim the handle so the index is rebuilt. Legacy meta files predate this field
+		// and have no handle — skip those.
```
jet-tong: We haven't launched yet so we can just assume we have the handle, but then if there is no handle there might have been an error somewhere that we need to surface...?

(Need edit on "and volumes mounted by a pre-upgrade driver (meta without a handle) aren't protected until remounted." in PR description too)

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -272,6 +272,12 @@ func (dm *DaemonsetMounter) mountOrShareSource(ctx context.Context, bucketName s
 		ServiceAccountEKSRoleARN: credentialCtx.ServiceAccountEKSRoleARN,
 		PodNamespace:             credentialCtx.PodNamespace,
 		FSGroup:                  fsGroup,
+		VolumeHandle:             credentialCtx.VolumeID,
+	}
+
+	// credentialCtx.VolumeID is the CSI volumeHandle; volumeID is the PV name.
+	if err := dm.mountMap.ClaimHandle(credentialCtx.VolumeID, volumeID); err != nil {
```
jet-tong: Question (which I forgot to consider too): what happens when mounter pod restarted with active workloads? Do we need to background cleanup the volume handle map (or is it handled by the Delete function already)? And how do we make sure it's cleaned up quickly so new workloads with e.g. same volume handle but different PV don't get rejected? (Same PV/volume handle is okay I guess so deployment workloads or restarts are okay)

I rmb kubelet retries NPV up to 2m2s (or maybe it's the back off max wait time) - which is only very slightly longer than 2 min background cleanup...? Could there be edge cases if we miss the 2s gap and we need to decrease the cleanup interval?

jet-tong: Should this check be here, or next to entry.Params.ValidateCompatibility? I was thinking about the edge case where 2 EXISTING volumes with different PV name but same volumeHandle - would one of them break on republish if we keep it above the 'default:' case below? (Could this edge case exist in the first place though, via v2 -> v3 upgrade or etc?)

I also thought this check would be inside ValidateCompatibility, not sure if inside or outside is better though since different PV but same volumeHandle is arguably slightly different from request param checks...? @rajdchak

tadiwa-aizen: Pushed a commit which is moving this into the first-mount path — republishes now skip the check, which fixes the existing-volume-breaks-on-republish case; kept it separate from ValidateCompatibility since that's same-PV param checks and this is a different-PV handle collision.

tadiwa-aizen: So when the last pod unmounts, the handle is freed right away (releaseTarget→Delete) — so normal reuse doesn't wait on the 2-min job. It only falls back to the background cleanup if that unmount cleanup fails. And the map lives in the csi-node, so a mounter-pod restart doesn't touch it; a csi-node restart rebuilds it from meta.json

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -296,6 +297,13 @@ func (dm *DaemonsetMounter) mountOrShareSource(ctx context.Context, bucketName s
 	}
 
 	if !entry.sourceMounted {
+		// First mount for this PV on the node — enforce per-node volumeHandle uniqueness here so
+		// the check runs once per entry, not on every republish/share. credentialCtx.VolumeID is
+		// the CSI volumeHandle; volumeID is the PV name. Released in MountMap.Delete on teardown.
+		if err := dm.mountMap.ClaimHandle(credentialCtx.VolumeID, volumeID); err != nil {
+			return fmt.Errorf("cannot mount volume %s: %w", volumeID, err)
```
rajdchak, line 307: what happens to this entry then? how do we get rid of this entry?

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: https://kubernetes.io/docs/reference/kubernetes-api/core/persistent-volume-v1/#CSIPersistentVolumeSource volume handle is a mandatory field on PV. Is it possible that we have no volume handle?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -296,6 +297,13 @@ func (dm *DaemonsetMounter) mountOrShareSource(ctx context.Context, bucketName s
 	}
 
 	if !entry.sourceMounted {
+		// First mount for this PV on the node — enforce per-node volumeHandle uniqueness here so
+		// the check runs once per entry, not on every republish/share. credentialCtx.VolumeID is
+		// the CSI volumeHandle; volumeID is the PV name. Released in MountMap.Delete on teardown.
+		if err := dm.mountMap.ClaimHandle(credentialCtx.VolumeID, volumeID); err != nil {
```
rajdchak: there is one more case where source mounted becomes false that is when the source was already mounted but it was not healthy. So in that case we could have a stale volume handle which could lead to rejection of this new request. I think we should do a cleanup before claiming. So may be your cleaning up of the volume handle should move to cleanupMount fn and we do this claiming after the cleanup mount has succeeded.

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: cleanupEntry skips entries with SourcePath == "", so the blank entry lingers

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -296,6 +297,13 @@ func (dm *DaemonsetMounter) mountOrShareSource(ctx context.Context, bucketName s
 	}
 
 	if !entry.sourceMounted {
```
rajdchak, line 299: sourceMounted can be false also when source is dead. Dead-source re-mount does NOT falsely reject based on volume handle — confirm with a test.

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak, line 1264: should we also claim the volume handle in this case?

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: Done in latest commit...  since volumeHandle is a required PV field, I dropped the "skip if empty" assumption: a recovered volume with no handle is now logged as an error (it'd mean corrupt/incomplete meta) instead of silently skipped.

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: Fixed, so  on reject we now Delete the entry, so the blank placeholder doesn't linger.

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: Added `TestMountMap_ClaimHandle_SamePVReclaimsAfterDeadSource`  — confirms a same-PV re-claim after a dead source isn't rejected, while a different PV still is.

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: Yes — now re-claiming the handle here too, so the recovering volume keeps it.

pkg/driver/node/mounter/daemonset_mounter.go
rajdchak: I think we should have an e2e test as well?

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -1169,12 +1180,27 @@ func (dm *DaemonsetMounter) populateEntryFromMeta(meta *MountMeta, sourcePath st
...
 	entry.Targets = targets
 	entry.sourceMounted = sourceMounted
 }
 
+// reclaimHandle re-registers a recovered volume's handle in the uniqueness index on restart,
+// so duplicates stay rejected. volumeHandle is a required PV field, so an empty one means the
+// meta is corrupt/incomplete — surface it rather than skip silently; a conflict (two recovered
+// volumes claiming the same handle) is logged too.
+func (dm *DaemonsetMounter) reclaimHandle(meta *MountMeta) {
+	if meta.VolumeHandle == "" {
```
rajdchak, line 1195: can this ever be empty?

rajdchak: https://github.com/awslabs/mountpoint-s3-csi-driver/blob/feature/daemonset-architecture/pkg/driver/node/node.go#L86-L89 this check here I think prevents it from being empty, validate again.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -1169,12 +1180,27 @@ func (dm *DaemonsetMounter) populateEntryFromMeta(meta *MountMeta, sourcePath st
...
+		klog.Errorf("MountMap: recovered volume %s has no volumeHandle in meta; skipping uniqueness re-claim", meta.VolumeID)
```
rajdchak, line 1196: we should return an error here and not just log, because according to validation above this is not accepted.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -1169,12 +1180,27 @@ func (dm *DaemonsetMounter) populateEntryFromMeta(meta *MountMeta, sourcePath st
...
+		return
+	}
+	if err := dm.mountMap.ClaimHandle(meta.VolumeHandle, meta.VolumeID); err != nil {
+		klog.Warningf("MountMap: volumeHandle conflict recovering volume %s: %v", meta.VolumeID, err)
```
rajdchak, line 1200: this should be returned as an error, not a warning log and skip.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -1246,6 +1275,7 @@ func (dm *DaemonsetMounter) RebuildMountMap() error {
 		targets := findBindMountTargets(mountInfos, deviceID(sourceMI), sourcePath)
 
 		dm.populateEntryFromMeta(meta, sourcePath, true, targets)
+		dm.reclaimHandle(meta)
```
rajdchak, line 1278: I think the order should be reclaim handle and then populate entry from meta

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -1236,6 +1262,9 @@ func (dm *DaemonsetMounter) RebuildMountMap() error {
 				// Store entry in the map so future NodePublish or periodic cleanup can
 				// retry cleanup using the original commDir where credentials were written.
 				dm.populateEntryFromMeta(meta, sourcePath, false, nil)
+				// Reserve the handle even for this dead-but-kept entry, so a different PV can't
+				// take it and lock this volume out when it recovers (same rule as a live mount).
+				dm.reclaimHandle(meta)
```
rajdchak, line 1267: I think the order should be reclaim handle and then populate entry from meta

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -1169,12 +1180,27 @@ func (dm *DaemonsetMounter) populateEntryFromMeta(meta *MountMeta, sourcePath st
...
 	entry.RefCount = len(targets)
 	entry.Targets = targets
 	entry.sourceMounted = sourceMounted
 }
 
+// reclaimHandle re-registers a recovered volume's handle in the uniqueness index on restart,
+// so duplicates stay rejected. volumeHandle is a required PV field, so an empty one means the
+// meta is corrupt/incomplete — surface it rather than skip silently; a conflict (two recovered
+// volumes claiming the same handle) is logged too.
+func (dm *DaemonsetMounter) reclaimHandle(meta *MountMeta) {
```
rajdchak, line 1194: we should have e2e tests for this

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -296,6 +297,16 @@ func (dm *DaemonsetMounter) mountOrShareSource(ctx context.Context, bucketName s
 	}
 
 	if !entry.sourceMounted {
+		// First mount for this PV on the node — enforce per-node volumeHandle uniqueness here so
+		// the check runs once per entry, not on every republish/share. credentialCtx.VolumeID is
+		// the CSI volumeHandle; volumeID is the PV name. Released in MountMap.Delete on teardown.
+		if err := dm.mountMap.ClaimHandle(credentialCtx.VolumeID, volumeID); err != nil {
```
rajdchak, line 303: claiming of handle should be done after the cleanupMount, so this needs to be reordered. This is for the case of a stale mount, say the volume handle has changed and we found a conflict so we delete the entry from the mountmap. We will never be able to track resources which were never cleaned up.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -296,6 +297,16 @@ func (dm *DaemonsetMounter) mountOrShareSource(ctx context.Context, bucketName s
...
+			// Drop the blank entry GetOrCreate inserted for this PV; otherwise it lingers,
+			// since periodic cleanup skips entries with an empty SourcePath.
+			dm.mountMap.Delete(volumeID)
```
rajdchak, line 306: probably delete only if source path is empty, for the corrupted source cases we might not want to delete it here, for that the delete should be done only after cleanup is successful?

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: Yes, for a new mount node.go already stops it (it rejects an empty volume ID). The only way it's empty is on upgrade: a volume mounted by an older version has no handle in its .meta.json, so after restarting on the new version it reads as empty.

So instead of erroring on empty (I tried that — it crashlooped the node on upgrade), I now just log it and skip that one volume.

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: node.go only stops it for new mounts. A volume from an older version (before this field) has no handle in its meta, so on upgrade it reads as empty — erroring there crashlooped the node on my test. So I kept empty as log-and-skip, and made the conflict case the error instead.

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: Done — conflict now returns an error instead of log-and-skip.

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: Done — reclaim now runs before populate.

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: Done — same here, reclaim now runs before populate.

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: Done — moved the claim to after cleanupMount, so a rejected mount can't leave stale resources untracked.

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: Handled by the reorder — the claim (and this delete) now run after cleanupMount succeeds, so by the time we'd delete, the resources are already cleaned. Nothing gets left untracked, so no separate source-path check needed.

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: Added an e2e in pod_sharing_daemonset.go: mount a volume, restart the csi-node pod (so it rebuilds from meta), then a second PV reusing the handle is still rejected. The other two branches — an empty handle, and two recovered volumes with the same handle — aren't easily e2e'd (need an old binary / a hand-made meta), so those stay unit + the manual upgrade run.

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: No — node.go rejects an empty volume ID and we persist the handle on every write, so a real meta always has one. Made it return an error here.

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: Done, returns an error now.

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: Done, returns an error now — the rebuild fails instead of continuing.

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: Added two: a duplicate volumeHandle is still rejected after a csi-node restart, and a re-mount of the same PV after its source dies isn't falsely rejected.

pkg/driver/node/mounter/daemonset_mounter.go
tadiwa-aizen: cleanupMount runs first now, so we only reach this Delete after it succeeded. The entry dropped here is the fresh blank one, no source path set yet.

PR #963, opened 2026-09-21
ci: add DockSec container scan with SARIF upload

advaitpatel:
## What this adds

A GitHub Actions workflow that scans the Dockerfile with [DockSec](https://github.com/OWASP/DockSec)
and uploads the results to the Security tab as SARIF, so container findings are
annotated inline on pull requests.

DockSec is an OWASP Lab Project (MIT). It wraps Hadolint and Trivy and reports
Dockerfile misconfigurations and image CVEs through a single SARIF upload.

## Why

DockSec scores the current `Dockerfile` at **62.5/100** across 7 findings:

```
DL4006 warning  lines 27, 33, 38  Set SHELL -o pipefail before RUN with a pipe
DL3040 warning  line 22           `dnf clean all` missing after dnf command
DL3041 warning  line 22           Specify version with dnf install -y <pkg>-<ver>
SC2006 style    line 27           Use \$(...) instead of legacy backticks
```

DL4006 is the one with a real failure mode, and it appears three times: without
`pipefail`, a piped `RUN` succeeds even when the command on the left of the pipe
fails. In a build that downloads and pipes through a checksum or an extractor,
that turns a failed fetch into a silently broken layer.

`DL3040` also has a size effect: without `dnf clean all` the package cache ships
in the layer.

A note on scope: this workflow scans the Dockerfile only, so the image CVE counts
above are not what it will report on day one. Dockerfile scanning is the part that
needs no build and no registry access, which keeps the job fast and the diff small.
Adding `image:` to the same step turns on the image scan once you decide where the
built image should come from, and I am glad to extend the PR that way if you prefer.

## What it does not do

- No API key and no AI calls: the workflow runs with `scan_only: true`, which is
  fully local to the runner.
- It does not gate merges. There is no `fail_on`, so the job reports and never
  fails the build. Adding a gate later is a one-line change.
- It does not replace existing tooling; findings land in the same Security tab.

## Details

- Runs on pull requests and pushes that touch a Dockerfile, plus a weekly
  scheduled scan so newly disclosed CVEs in the base image surface without a
  code change.
- All three actions are pinned to a commit SHA.
- `security-events: write` is scoped to the single job that needs it; the
  workflow's default permission is `contents: read`.

Happy to adjust the trigger set, drop the schedule, or split it into an existing
workflow instead if you would prefer that.

---

Disclosure: I am the project lead for DockSec, the tool this workflow adds. Happy to close this if it is not a fit - no hard feelings either way.


jet-tong (commented):
Thanks for putting this together, but for now we don't particularly want to add a new workflow to the project, and will pass on this PR. 

For image CVE scans, we currently have plans to use AWS's inspector2, and for Dockerfile linting, it seems manually running e.g. `docker run --rm -i hadolint/hadolint:v2.15.1 < Dockerfile` (and Dockerfile.local) would cover what we need without the wrapper; we could consider adding it as a comment for manual run or to our Makefile for linting in the future. Some of the findings described seem to be useful to improve our Dockerfiles; if you'd like to raise a PR for them please go ahead and we can review it, otherwise we're happy to pick those up ourselves when we need to.

There's one thing that might be worth looking into on your side - I did run a workflow with DockSec on my fork, and the workflow reported `DS013 "RUN cd ... to change directory"` against Dockerfile:15, but we have no `RUN cd` there and line 15 is `ARG MOUNTPOINT_VERSION=1.24.0` on my run. Looking a bit further, the match is in a nested file in `tests/e2e-kubernetes/scripts/eks-addon/Dockerfile:15` where we do have `RUN cd tests/e2e-kubernetes && go test -c -o /e2e.test .`.  A quick agent scan found that [findings.py](https://github.com/OWASP/DockSec/blob/84c26968e6741c5c96ab6db1be106f0e71099c0f/docksec/findings.py#L278) in docksec is only a glob on the name (matching `tests/.../Dockerfile`), then [report_generator.py](https://github.com/OWASP/DockSec/blob/84c26968e6741c5c96ab6db1be106f0e71099c0f/docksec/report_generator.py#L269) seems to label all results with basename so the nested file's findings are printed on the root file. Please do have a look to see if it's a real bug.

advaitpatel:
Understood, thanks for the consideration and for the detailed look at the DS013 mismatch. Filing that as a bug on DockSec now, referencing this PR.

Appreciate the offer on the Dockerfile findings themselves too - I'll leave that to you and the team as suggested.

Closing this out.

PR #967, opened 2026-09-22
Return error on failure to read meta file

rajdchak:
*Issue #, if available:*

*Description of changes:*
Return error on failure to read meta file. A meta file is the only record of a mount's resources. Skipping it would silently leak those resources.


pkg/driver/node/mounter/mount_map_persistence_test.go
```diff
@@ -376,18 +376,27 @@ func TestRebuildMountMap_SkipsNonMetaFiles(t *testing.T) {
...
+	metaPath := filepath.Join(metaDir, "vol-bad.meta.json")
+	os.WriteFile(metaPath, []byte("not-json"), 0640)
 
 	dm := newTestDMWithMountInfo(kubeletPath, fakeMountInfoProvider(nil))
 	err = dm.RebuildMountMap()
-	assert.NoError(t, err) // should not error, just skip
+
+	// Fail closed: an unreadable/corrupt meta must abort rebuild (driver.go turns this into
+	// a fatal), not be silently skipped — skipping would leak the volume's commDir/credentials
+	// with no recovery path.
```
jet-tong, line 394: np: this should be in source code.

pkg/driver/node/mounter/daemonset_mounter.go
```diff
@@ -1213,8 +1213,7 @@ func (dm *DaemonsetMounter) RebuildMountMap() error {
 		metaPath := filepath.Join(metaDir, dirEntry.Name())
 		meta, err := readMeta(metaPath)
 		if err != nil {
-			klog.Warningf("MountMap: failed to read meta file %s, skipping: %v", metaPath, err)
-			continue
+			return fmt.Errorf("failed to read mount meta %s; remove or quarantine this file to allow the node to start: %w", metaPath, err)
```
jet-tong, line 1216: np: let's add a message on how to recover, e.g. drain/restart workloads and daemonset (the catch-all method) or otherwise.

jet-tong (commented):
As discussed offline today and testing via trying to delete a meta entry to see behaviour, we're deciding to temporarily accept this PR (please remember to add message on how to recover, e.g. drain/restart workloads and daemonset), since this isn't a normal use case scenario that only happens when users edit meta files. We should be clear in documentation not to touch them.

In the future, we should limit blast radius of this problem to PV-wise but not node-wise, but failing only in NPV, with a method to mitigate / reconcile any errors. This will be a nice-to-have in case of any user misedits or bug-caused errors relating to corrupted/missing meta files.

